Issue #15, opened 2025-11-06
Add security reusable workflows


WilliamBerryiii:
## Issue Description

The repository needs reusable workflow files for security tasks including dependency pinning validation, SHA staleness checking, secret scanning, and IaC security scanning.

## Additional Context

**Files to add:**
- `.github/workflows/reusable-dependency-pinning.yml` - Validate dependencies are pinned to SHAs
- `.github/workflows/reusable-sha-staleness.yml` - Check for stale SHA references
- `.github/workflows/reusable-gitleaks.yml` - Secret scanning with Gitleaks
- `.github/workflows/reusable-checkov.yml` - IaC security scanning with Checkov
- `.github/workflows/reusable-codeql.yml` - CodeQL security analysis

**Priority:** Wave 3 (Priority 3) - Depends on Wave 2 completion

**Dependencies:** Requires security scripts from Issue #9, configurations from Issue #3

**Testing:** Validate YAML syntax, test each reusable workflow with sample inputs, verify security scans detect known issues

PR #28, opened 2025-11-06
feat(workflows): add security reusable workflows

WilliamBerryiii:
Implemented comprehensive GitHub native security tooling for automated security analysis and vulnerability detection across the repository.

- **feat**(_workflows_): Added codeql-analysis.yml for JavaScript/TypeScript security analysis with security-extended queries
- **feat**(_workflows_): Added dependency-review.yml for PR-based dependency vulnerability checking
- **feat**(_workflows_): Updated weekly-security-maintenance.yml to use CodeQL and GitHub native security features
- **feat**(_workflows_): Updated security-scan.yml to orchestrate CodeQL analysis and dependency review
- **chore**(_workflows_): Removed obsolete third-party tool configurations

Security coverage includes CodeQL analysis (200+ vulnerability patterns), GitHub Secret Scanning (automatic), Dependabot Alerts (automatic), and custom SHA pinning validation. All workflows use SHA-pinned actions and follow security best practices.

Resolves: #15

🔒 - Generated by Copilot

### DIFF
--- a/.github/workflows/weekly-security-maintenance.yml
+++ b/.github/workflows/weekly-security-maintenance.yml
@@ -77,7 +77,7 @@ jobs:
           ## 📊 Dependency Health
           
           | Metric | Status | Details |
-          |--------|--------|---------||
+          |--------|--------|---------|
           | SHA Pinning Compliance | $(if ($unpinnedCount -eq '0') { '✅ Pass' } else { '⚠️ Warning' }) | Score: $pinningScore% |
           | Unpinned Dependencies | $(if ($unpinnedCount -eq '0') { '✅ None' } else { "⚠️ $unpinnedCount found" }) | $(if ($unpinnedCount -ne '0') { 'Review warnings above' } else { 'All pinned' }) |
           | Stale SHAs (>${{ inputs.max-age-days }} days) | $(if ($staleCount -eq '0') { '✅ None' } else { "⚠️ $staleCount found" }) | $(if ($staleCount -ne '0') { 'Review warnings above' } else { 'All current' }) |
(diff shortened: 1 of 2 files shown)


WilliamBerryiii:
The build will pass when the prior commits with the scripts this on relies on are there.

.github/workflows/weekly-security-maintenance.yml
```diff
@@ -0,0 +1,292 @@
...
+      - name: Upload validation report
+        if: always()
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v4.4.3
+        with:
+          name: dependency-pinning-results
+          path: logs/dependency-pinning-results.json
+          retention-days: 90
+
+  # Job 2: Check for stale SHA pins
+  check-staleness:
```
bderusha, line 32: Should this reference `sha-staleness-check.yml` instead of being reimplemented here?

.github/workflows/weekly-security-maintenance.yml
```diff
@@ -0,0 +1,292 @@
...
+        type: number
+        default: 30
+
+permissions:
+  contents: read
+  security-events: write
+
+jobs:
+  # Job 1: Validate all dependencies are SHA-pinned
+  validate-pinning:
```
bderusha, line 20: Should this be spun out into its own file like the others?

.github/workflows/sha-staleness-check.yml
```diff
@@ -0,0 +1,53 @@
+name: SHA Staleness Check
+
+on:
+  schedule:
+    - cron: '0 0 * * 0' # Weekly on Sunday at midnight UTC
```
bderusha: Can you provide more details on why this workflow is triggered here and is also run in the weekly security maintenance workflow. They seem the same, but am I missing something?

.github/workflows/checkov-scan.yml
```diff
@@ -0,0 +1,109 @@
+name: Checkov Scan
+
+on:
+  workflow_call:
```
bderusha: Consider if it would be valuable to organize workflows around a naming convention (while github doesn't yet work with subdirectories)

- Prefixing workflows only triggered by other workflows E.G. `_checkov-scan.yml`, `wc-checkov-scan.yml`, `subwf.checkov-scan.yml`, or the like
- Prefixing the triggered workflows to make them distinct from the rest, E.G `workflow.weekly-security-maintenance.yml`
- Or perhaps a convention which reveals the trigger type regardless. E.G:
  - `workflowcall.checkov-scan.yml`
  - `schedule.weekly-security-maintenance.yml`
  - `multi_push_pr.gitleaks.yml`

.github/workflows/weekly-security-maintenance.yml
WilliamBerryiii: OMG ... yes it should. I refactored the Edge-AI jobs to split them into tighter workflows so ..100%. Will make the change. Thanks for the catch.

.github/workflows/weekly-security-maintenance.yml
WilliamBerryiii: given that we have a dedicated PS script for this ... the answer is yes ... same as below.

WilliamBerryiii: fixed

.github/workflows/checkov-scan.yml
WilliamBerryiii: converted to issue to tackle later.

.github/workflows/sha-staleness-check.yml
WilliamBerryiii: I've ripped this out ... refactoring cruft from edge-ai. It's part of a more complex dependency scanning trigger in edge-ai that does not need to be in this project.

PR #284, opened 2026-01-25
fix: exclude test fixtures from markdown link checker

SrishJ23:
This change excludes test fixture markdown files from the markdown link checker,
preventing intentional broken links from appearing as false positives.


SrishJ23:
@microsoft-github-policy-service agree

SrishJ23:
Hi @WilliamBerryiii, thanks a lot for the detailed explanation and guidance 🙏  

You’re absolutely right — I misunderstood how `ignorePatterns` works and appreciate you pointing out that it applies to link targets, not source files.

I’ve now updated the implementation to exclude the test fixture markdown files at the file-selection stage inside `Get-MarkdownTarget`, by filtering them out after `git ls-files`, as you suggested. This ensures fixture files are no longer scanned while keeping the link-checker behavior intact.

Please let me know if this approach looks good to you or if you’d prefer a more configurable solution (e.g. an `-ExcludePath` parameter). Happy to make further adjustments if needed.  
Thanks again for your help!

scripts/linting/Markdown-Link-Check.ps1
```diff
@@ -98,37 +64,28 @@ function Get-MarkdownTarget {
...
         }
         elseif (Test-Path -Path $item -PathType Container) {
-            # Directory - get all tracked markdown files
             $absolutePath = (Resolve-Path $item).Path
             $relativePath = [System.IO.Path]::GetRelativePath($repoRoot, $absolutePath)
             $searchPath = if ($relativePath -eq '.') { '*.md' } else { "$relativePath/**/*.md" }
 
             Write-Verbose "Searching in: $searchPath"
-            $trackedFiles = git ls-files $searchPath 2>$null
+
```
WilliamBerryiii: Thanks for working on this! The fix itself is correct — filtering out test fixtures at the file-selection stage is the right approach.

However, the PR currently includes a lot of unrelated changes (removed functions, deleted documentation, etc.). Could you please reset to just the targeted fix? Here's how:

```
# 1. Fetch latest upstream and reset your branch to main
git fetch upstream
git checkout fix-markdown-link-check
git reset --hard upstream/main

# 2. Make the targeted change in scripts/linting/Markdown-Link-Check.ps1
#    Around line 125, change this:
#
#        $trackedFiles = git ls-files $searchPath 2>$null
#
#    To this:
#
#        $trackedFiles = git ls-files $searchPath 2>$null |
#            Where-Object { $_ -notlike 'scripts/tests/Fixtures/*' }

# 3. Commit and force push
git add scripts/linting/Markdown-Link-Check.ps1
git commit -m "fix: exclude test fixtures from markdown link checker"
git push --force-with-lease origin fix-markdown-link-check
```

That's the only change needed — no config file changes required since ignorePatterns applies to link targets (URLs), not source files.

Let me know if you have any questions!

PR #928, opened 2026-03-07
Fix compliance score calculation in Get-ComplianceReportData

04cb:
The \`Get-ComplianceReportData\` function was calculating compliance scores using a confusing formula that derived \`PinnedDependencies\` from the difference between total violations and non-Info severity violations. Since no violations are currently assigned 'Info' severity, this resulted in \`PinnedDependencies\` always being 0 when violations were found.

The fix simplifies the logic to explicitly set \`PinnedDependencies\` to 0 and \`ComplianceScore\` to 0% when violations are found (since all items in the Violations array are unpinned dependencies), and 100% when no violations are found. This preserves the existing behavior while making the code more maintainable and easier to understand.

**Changes:**
- Removed the confusing \`$unpinnedDeps\` calculation with severity filtering
- Removed the derived \`$pinnedDeps\` calculation
- Explicitly set \`PinnedDependencies = 0\` when violations exist
- Clarified that \`TotalDependencies\` represents the count of violations found
- Added clear comments explaining the calculation logic

### DIFF
--- a/scripts/security/Test-DependencyPinning.ps1
+++ b/scripts/security/Test-DependencyPinning.ps1
@@ -775,20 +775,20 @@ function Get-ComplianceReportData {
     $report.ScannedFiles = $ScannedFiles.Count
     $report.Violations = $Violations
 
-    # Calculate metrics
+    # Calculate metrics - TotalDependencies is set to the count of all violations found
+    # This represents dependencies that failed validation (unpinned or other violations)
     $totalDeps = @($Violations).Count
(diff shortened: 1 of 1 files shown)


WilliamBerryiii:
While there are some potential improvements that can be made in this section of the code, there's no issue filed that demonstrates incorrect behavior of the system or the calculations. Please file an issue first if you believe the calculations are incorrect and we will triage from there.

Issue #2086, opened 2026-06-20
Reach and enforce 80% PowerShell test coverage


WilliamBerryiii:
## Summary

Reach and enforce 80% PowerShell test coverage by cleaning the Pester coverage denominator, closing concentrated zero-coverage gaps, and turning the dormant 80% target into an enforced CI gate.

### 1. Denominator Hygiene

The measured coverage "drop" is a denominator artifact: generated `plugins/` output inflates the coverage denominator. Remove `'plugins'` from `$coverageDirs` in `scripts/tests/pester.config.ps1` so coverage reflects real source under test.

### 2. Targeted Coverage Tests

Add deterministic unit tests for concentrated zero-coverage areas to build a buffer above 80% (target ~84%+):

- PowerPoint skill helpers (`Invoke-EmbedAudio`, `Invoke-ExportSvg`, `Invoke-GenerateThemes`, `Invoke-PptxPipeline`)
- `scripts/lib/Modules/CIHelpers.psm1`
- Additional low-coverage helpers across `scripts/lib`, `scripts/security`, `scripts/linting`

### 3. Enforcement Ratchet

Replace the warn-only threshold (`$threshold = 18`) in `.github/workflows/pester-tests.yml` with a failing 80% gate aligned to `CoveragePercentTarget = 80` as the single source of truth. Keep Codecov `patch` status informational (contributor warn-only) while the project gate hard-fails.

## Acceptance Criteria

- Coverage denominator excludes generated `plugins/` output
- Local coverage stays >= 80% (with ~84%+ buffer)
- CI fails when coverage drops below 80%
- All existing and new Pester tests pass

jkim323:
Coverage percentage is a necessary signal but not a sufficient one in my opinion — a ≥80% gate is gameable by tests that execute lines without asserting on behavior (e.g., `Should -Not -Throw`-only tests, or "mock everything and assert the mock was called" patterns, which will be tempting for `Invoke-EmbedAudio` / `Invoke-ExportSvg` / `Invoke-PptxPipeline` in #2089 issue). Thoughts on adding a few quality bars to the acceptance criteria?

WilliamBerryiii:
[...]

PR #2102, opened 2026-06-22
test(tests): enforce 80% PowerShell coverage gate

WilliamBerryiii:
## Description

This PR turned the repository's dormant 80% PowerShell coverage target into an **enforced CI gate** driven by a single source of truth, and built a measured buffer above the threshold so the gate lands green.

Key changes:

* **Cleaned the coverage denominator.** Removed the generated *plugins* tree from the Pester coverage directory set in *pester.config.ps1*, so coverage now reflects authored scripts rather than generated output.
* **Built a coverage buffer.** Added Pester suites for the PowerPoint skill scripts (*Invoke-EmbedAudio.ps1*, *Invoke-ExportSvg.ps1*, *Invoke-GenerateThemes.ps1*) and the video-to-gif *convert* script, plus additional security and CIHelpers tests targeting previously uncovered branches.
* **Made the target a single source of truth.** *Invoke-PesterTests.ps1* now emits `CoveragePercent` and `CoverageTarget` into *logs/pester-summary.json*, both derived from `CoveragePercentTarget` in *pester.config.ps1*.
* **Enforced the gate in CI.** *pester-tests.yml* now reads the coverage target output and fails the build (`exit 1`) when measured coverage falls below it.
* **Refactored three PowerPoint scripts** behind a dispatch guard so their functions are unit-testable without executing the script body.

Measured at submission: **83.69%** coverage against the **80%** target — a 3.69-point buffer.

## Related Issue(s)

Related to #2086

## Type of Change

* [x] GitHub Actions workflow

**Other:**

## Testing

* Ran the full Pester suite with coverage via the direct runner: **2659 tests / 2587 passed / 0 failed / 5 skipped**.
* Verified coverage at **83.69%** against the **80%** target (3.69-point buffer).
* `npm run lint:ps` (PSScriptAnalyzer): **all PowerShell files passed**.

### Required Checks

### Required Automated Checks

* [x] PowerShell analysis: `npm run lint:ps`
* [x] PowerShell tests: `npm run test:ps` (with coverage)

## Security Considerations

## Additional Notes

A pre-merge review surfaced no Critical or High findings. A small number of Medium/Low observations (for example, a missing-venv error-message regression and an `$env:TEMP` restore in one test teardown) are tracked as follow-ups (WI-01/02/03) and do not block this gate.


.github/skills/experimental/video-to-gif/tests/convert.Tests.ps1
```diff
@@ -0,0 +1,222 @@
+#Requires -Modules Pester
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: MIT
+
+BeforeAll {
```
rezatnoMsirhC, line 5: `global:ffmpeg` and `global:ffprobe` are defined in `BeforeAll` but never torn down. Please add an `AfterAll` block to remove them so they do not leak into later test suites running in the same session:

```powershell
AfterAll {
    Remove-Item -Path Function:ffmpeg -ErrorAction SilentlyContinue
    Remove-Item -Path Function:ffprobe -ErrorAction SilentlyContinue
    Remove-Item -Path $script:StubRoot -Recurse -Force -ErrorAction SilentlyContinue
}
```

(The existing `AfterAll` on the stub directory can be folded in here too.)

rezatnoMsirhC (approved):

.github/skills/experimental/video-to-gif/tests/convert.Tests.ps1
WilliamBerryiii: Thank you for catching this, @rezatnoMsirhC! You're right that the `ffmpeg`/`ffprobe` stubs leaked into later suites. I've added an `AfterAll` that removes them with `Remove-Item -Path 'Function:\<name>' -Force -ErrorAction SilentlyContinue`, matching the teardown convention used elsewhere in the repo. While here I also fixed the same latent leak in two sibling security test files (`Invoke-PipAudit.Tests.ps1` and `Sign-PlannerArtifacts.Tests.ps1`). All suites validated green.

.github/workflows/pester-tests.yml
```diff
@@ -143,17 +146,27 @@ jobs:
             exit 0
           }
           $percentage = [double]$coverageValue
-          $threshold = 18  # Informational baseline
+          $targetValue = '${{ steps.pester.outputs.coverage_target }}'
+          if ([string]::IsNullOrWhiteSpace($targetValue)) {
+            # Fall back to the authoritative target in scripts/tests/pester.config.ps1
+            $targetValue = 80
```
jkim323: Hm if `Invoke-PesterTests.ps1 `ever stops emitting` CoverageTarget`, the gate silently keeps enforcing 80 even after `pester.config.ps1` is bumped. I suggest failing loudly instead.. Thoughts?

.github/workflows/pester-tests.yml
jkim323: For consistency with the `coverage_target` empty-output handling below, I think this empty-coverage branch should also fail loudly unless soft-fail is set. Otherwise the two empty-output paths behave inconsistently — one silently skips, the other (after the suggested fix) hard-errors.

jkim323 (changes requested):

.github/workflows/pester-tests.yml
WilliamBerryiii: Resolved in 4b8c48e3. The empty-coverage branch now fails loudly: the emit step uses an explicit `if ($null -ne $summary.CoveragePercent)` guard (so a 0%/falsy value is no longer silently skipped), and the `Coverage Threshold Check` step reads `logs/pester-summary.json` directly and hard-errors when the file is missing or `CoveragePercent` is null. Both empty-output paths now behave consistently, honoring `soft-fail` (warn + exit 0) only when explicitly set.

.github/workflows/pester-tests.yml
WilliamBerryiii: Resolved in 4b8c48e3. The hardcoded `\ = 80` fallback was removed entirely. The threshold now comes only from `pester.config.ps1` (`CoveragePercentTarget`), which `Invoke-PesterTests.ps1` writes into `logs/pester-summary.json` as `CoverageTarget`. The gate reads that value and fails loudly when `CoverageTarget` is null/missing instead of silently enforcing a stale 80, so bumping the target in `pester.config.ps1` is always honored.

jkim323 (approved):

PR #2134, opened 2026-06-22
 fix(ci): add composite action for PS module install with cache and retry

### LINKED ISSUE #2105: fix: CI reliability (PSGallery outages)
### Component

Instruction Files

### Bug Description

Model-Validation workflow failed in the [PR 2098](https://github.com/microsoft/hve-core/pull/2098). The error log says, `Install-Package: No match was found for the specified search criteria and module name 'PowerShell-Yaml'.` which is a known signature of a PSGallery outage. Seems like the CI workflows fail intermittently when PSGallery has transient outages during Install-Module. Since every workflow installs modules inline without caching or retry, a single gallery hiccup fails the entire job. 

### Expected Behavior

Introduce `action.yml` — a shared composite action that:
    - Caches all modules from ps-module-versions.json
    - Cache key: ps-modules-{os}-{hash(ps-module-versions.json)} — auto-invalidates on version bumps
    - On cache miss: installs each module with 3 retry attempts and exponential backoff (10s, 20s, 40s)
    - On cache hit: skips installation entirely (zero PSGallery calls)

### Steps to Reproduce

_No response_

### Additional Context

_No response_

jkim323:
# Pull Request

## Description
PSGallery transient outages caused CI failures across multiple workflows. Each workflow independently ran `Install-Module `without caching or retry, so a single PSGallery hiccup could fail an entire CI run. The composite action centralizes this logic with:
- Module caching keyed on runner.os + [hashFiles](https://github.com/microsoft/hve-core/blob/main/scripts/security/ps-module-versions.json)
- Exponential backoff (10s → 20s → 40s) with 3 retries per module
- Single source of truth for module versions (ps-module-versions.json)

Once the composite action was added under actions, it became clear that the security scanners (Test-SHAStaleness.ps1 and Test-DependencyPinning.ps1 only scanned workflows — missing the new actions/cache SHA pin. This PR closes that gap.

**Solution**
- Introduce `action.yml` a composite action that:
     - Caches modules keyed on runner.os + hashFiles
     - Retries installation with exponential backoff (3 attempts: 10s → 20s → 40s)
     - Centralizes module version pinning via `ps-module-versions.json`

- Replace inline Install-Module steps in all 11 affected workflows with `uses: ./.github/actions/setup-ps-modules`.
- Extend `Test-SHAStaleness.ps1` and `Test-DependencyPinning.ps1` to scan `.github/actions/**/*.yml` alongside workflows. This ensures the composite action's actions/cache SHA pin gets the same staleness and pinning audits as any workflow-level dependency — no blind spots introduced by this change.
- Update` copilot-instructions.md `and security `README` with new conventions.

**Testing**
- Full security Pester suite: 489 passed, 0 failed, 1 skipped (network-dependent)
- No conflict markers in staged files
- Composite action discovery verified via new test contexts

## Related Issue(s)
Fixes #2105 

## Type of Change

Select all that apply:

**Code & Documentation:**

* [ ] Documentation update

* [x] GitHub Actions workflow
* [ ] Security configuration
* [ ] Dependency update

**AI Artifacts:**

* [ ] Reviewed contribution with `prompt-builder` agent and addressed all feedback

>

**Other:**

* [ ] Script/automation (`.ps1`, `.sh`, `.py`)

**User Request:**

**Execution Flow:**

**Output Artifacts:**

**Success Indicators:**

### Required Checks

* [ ] Documentation is updated (if applicable)
* [ ] Files follow existing naming conventions
* [ ] Tests added for new functionality (if applicable)

### AI Artifact Contributions

* [ ] Used `/prompt-analyze` to review contribution
* [ ] Addressed all feedback from `prompt-builder` review

### Required Automated Checks

The following validation commands must pass before merging:

* [x] Markdown linting: `npm run lint:md`
* [ ] Frontmatter validation: `npm run lint:frontmatter`
* [ ] Skill structure validation: `npm run validate:skills`
* [x] PowerShell analysis: `npm run lint:ps`
* [ ] Plugin freshness: `npm run plugin:generate`
* [ ] Docusaurus tests: `npm run docs:test`

## Security Considerations


.github/actions/setup-ps-modules/action.yml
```diff
@@ -0,0 +1,40 @@
+name: Setup PowerShell Modules
```
WilliamBerryiii, line 1: I'm not sure we can build and run our own actions because of org policies. But if this does work, we should get a backlog of other common steps like this we can extract as well.

jkim323: Would you be able to help me dig into the org policies regarding this pls? I am not sure what channels I need to look into to get an answer for it.... I hope it is allowed because it would a great refactor to implement.

.github/workflows/pester-tests.yml
```diff
@@ -45,16 +45,8 @@ jobs:
...
-        run: |
-          Install-Module -Name Pester -RequiredVersion 5.7.1 -Force -Scope CurrentUser
-          Import-Module Pester -RequiredVersion 5.7.1
-
-      - name: Install PowerShell-Yaml
-        shell: pwsh
-        run: |
-          Install-Module -Name PowerShell-Yaml -RequiredVersion 0.4.7 -Force -Scope CurrentUser
+      - name: Setup PowerShell modules
+        uses: ./.github/actions/setup-ps-modules
```
rezatnoMsirhC, line 49: The old install step both installed and explicitly imported Pester at the pinned version:

```powershell
Install-Module -Name Pester -RequiredVersion 5.7.1 -Force -Scope CurrentUser
Import-Module Pester -RequiredVersion 5.7.1
```

The composite action only installs; it does not import. The explicit `Import-Module ... -RequiredVersion 5.7.1` ensured the runner's session loaded exactly the pinned version, even if the runner image already has a different Pester version earlier on `$env:PSModulePath`. Auto-loading will pick up the module on first use, but may silently load the wrong version.

Consider adding an import step to the composite action after installation:

```yaml
    - name: Import Pester at pinned version
      shell: pwsh
      run: |
        $config = Get-Content -Raw (Join-Path $env:GITHUB_WORKSPACE 'scripts/security/ps-module-versions.json') | ConvertFrom-Json
        if ($config.modules.PSObject.Properties.Name -contains 'Pester') {
          Import-Module Pester -RequiredVersion $config.modules.Pester.version -Force
        }
```

scripts/tests/security/Test-SHAStaleness.Tests.ps1
```diff
@@ -1062,6 +1062,36 @@ jobs:
...
+
+                Should -Invoke Get-BulkGitHubActionsStaleness -Times 1
+            }
+            finally {
+                Pop-Location
+            }
+        }
+    }
+
     Context 'No workflow directory' {
```
rezatnoMsirhC, line 1097: The test asserts `Get-BulkGitHubActionsStaleness -Times 1` but does not confirm the invocation was triggered by discovering the composite action file specifically. A parameter filter would make the assertion more targeted and ensure the function was called with repos sourced from the composite action:

```powershell
Should -Invoke Get-BulkGitHubActionsStaleness -Times 1 -ParameterFilter {
    $ActionRepos -contains 'actions/cache'
}
```

jkim323:
## Updates made to PR:
### Description
This PR generalizes the PowerShell module install pattern across the repository into a single, testable entry point with caching and exponential-backoff retry.

### Design Decisions
- No special-case Pester step — The generic import step handles all modules uniformly when callers set import: 'true'
- `copilot-setup-steps.yml` calls the script directly — The Copilot coding agent requires `-Scope AllUsers` (modules install to _/usr/local/share/powershell/Modules)_, but the composite action caches and restores to the CurrentUser path _(~/.local/share/powershell/Modules_). Using the composite action there would mean cache hits restore modules to a path the AllUsers scope never reads, so the workflow bypasses caching and calls `Install-PSModules.ps1` directly.
- Retry with backoff — Addresses the original PSGallery transient failures that motivated this PR

### Changes Made
- ` Install-PSModules.ps1` centralized install script with:
   - exponential backoff retry (default 3 attempts, 10s base delay)
   - reads pinned versions from `ps-module-versions.json`
   - new parameters and functions
- new action.yml composite action wrapping the script:
   - cache step keyed on `ps-module-versions.json` hash
   - install step (skipped on cache hit unless `force: true`)
   - import step (opt-in via `import:true`) imports all pinned modules generically
- new Install-PSModules.Tests.ps1 - 25 pester unit tests to cover all function
- migration (all callers moved to the composite action or script)
- `Test-SHAStaleness.Tests.ps1`: Added composite-action scanning context (detects SHA-pinned actions in `action.yml` + targeted` -ParameterFilter` assertion confirming actions/cache was sourced from the composite action
- `Test-DependencyPinning.Tests.ps1`: New tests for the composite action's SHA-pinned actions/cache reference

scripts/tests/security/Test-SHAStaleness.Tests.ps1
jkim323: updated!

.github/workflows/pester-tests.yml
jkim323: Ah my thoughts. Importing inside the composite action wouldn't help, though — composite run: steps are separate processes, so the import doesn't persist to the workflow's "Run Pester Tests" step. The version guarantee is already enforced in Invoke-PesterTests.ps1, which imports Pester at the pinned version in the same session right before Invoke-Pester. The action's gated import: 'true' step also imports all pinned modules for fail-fast validation.

.github/actions/setup-ps-modules/action.yml
```diff
@@ -0,0 +1,52 @@
...
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: Cache PowerShell modules
+      id: ps-cache
+      uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
+      with:
+        path: ~/.local/share/powershell/Modules
```
katriendg, line 25: **`scope` input is a cache/scope footgun.** The cache `path` is hardcoded to the CurrentUser location (`~/.local/share/powershell/Modules`), but the action exposes a `scope` input that accepts `AllUsers`. If a caller sets `scope: AllUsers`, `Install-Module` writes to `/usr/local/share/powershell/Modules`, so the cache would never restore to the path actually being read — silent caching breakage and zero speedup. This is the same mismatch the PR description already acknowledges for `copilot-setup-steps.yml` (which correctly bypasses the action). Consider either dropping the `scope` input from the composite action, or making the cached `path` scope-aware so the two can't drift.

.github/actions/setup-ps-modules/action.yml
```diff
@@ -0,0 +1,52 @@
...
+        key: ps-modules-${{ runner.os }}-${{ hashFiles('scripts/security/ps-module-versions.json') }}${{ inputs.cache-key-suffix && format('-{0}', inputs.cache-key-suffix) || '' }}
+
+    - name: Install PowerShell modules
+      if: steps.ps-cache.outputs.cache-hit != 'true' || inputs.force == 'true'
```
katriendg: **Cache-hit skips install with no integrity re-check.** On a cache hit the install step is skipped entirely, so a previously-saved partial or corrupt module cache (e.g. a job cancelled mid-install) would be restored and trusted on every subsequent run, persistently breaking the job until the cache key changes — counter to the reliability goal of this PR. Since `Install-PSModules.ps1` is already idempotent (`Test-ModulePresent` skips modules present at the pinned version), always running the install step and using the cache purely as a speedup would be more robust than gating install on the cache hit.

katriendg (commented):
Thanks for tackling this, @jkim323 — flagging for visibility that this is higher-priority than a routine CI tidy-up.

We're seeing the PSGallery transient outage (#2105) hit `main` CI **more frequently lately**, and because every workflow currently runs inline `Install-Module` with no retry or cache, a single gallery hiccup fails the **entire** run and blocks unrelated PRs. Centralizing install behind one tested entry point with exponential-backoff retry plus caching removes that whole class of flaky failures and gives us a single place to tune behavior going forward.

**Validation against #2105:** it solves both failure modes the issue called for — retry with backoff (`Install-SingleModule`, 10s→20s→40s, verified by unit tests) and caching keyed on `runner.os` + `hashFiles(ps-module-versions.json)` so the cache auto-invalidates on version bumps. The 11 workflow migrations, the devcontainer, and the linter entry point all route through the one tested path, and version literals are no longer duplicated so `lint:ps-module-pins` stays green.

**On the org-policy question (@WilliamBerryiii):** local composite actions referenced via a relative path (`./.github/actions/...`) are **not** governed by the org/repo "Allowed actions" allow-list — that policy only applies to actions consumed from other repos or the Marketplace. Actions defined in the same repository always run regardless of the allow-list, so this should work without a policy exception.

Two small robustness notes left inline so we don't trade one reliability gap for another. Core approach is solid — approving with comments.

.github/actions/setup-ps-modules/action.yml
jkim323: Ah good point - the script already skips modules present at the required version, so running unconditionally acts as a cache-integrity gate rather than blindly trusting the cache layer... Updated by dropping the `if:` gate on the install step and rely on the script's idempotency. I also added a test asserting the install step has no if condition.

.github/actions/setup-ps-modules/action.yml
jkim323: Yep, you are right. I thought about making the cache path scope but decided to drop the input (remove the scope input entirely and have action always install to CurrentUser. Reasons listed below and I would love to hear your thoughts:
- Only one caller (`copilot-setup-steps.yml`) needs AllUsers, and it already bypasses the action by design (it runs before the workspace is available for hashFiles)
- A scope-aware cache path couples the action to platform-specific module path conventions that could drift across runner images
- The simpler contract (always CurrentUser, script idempotency as integrity gate) is easier to reason about and test
- If a future caller genuinely needs AllUsers + caching, we can revisit I think in another PR.

Also added Pester tests asserting the composite action has no scope input, hardcodes Scope to CurrentUser, and caches the correct module path.

.github/workflows/pester-tests.yml
jkim323: I also did confirmed —` Invoke-PesterTests.ps1 `lines 93–105 read the pinned Pester version from `ps-module-versions.json` and call `Import-Module -Name Pester -RequiredVersion $pesterVersion -ErrorAction Stop` in the same session that runs `Invoke-Pester`.  Hope this addresses the concern..

jkim323:
Updated: Repointing model catalog updater to the renamed github/docs source.

context: https://docs.github.com/en/enterprise-cloud@latest/copilot/reference/copilot-billing/request-based-billing-legacy/what-changed-with-billing

WilliamBerryiii (approved):

rezatnoMsirhC (approved):

agreaves-ms (approved):

PR #1623, opened 2026-05-20
feat(skills): add demo-setup skill to design-thinking collection

chaosdinosaur:
# Pull Request

## Description

Adds the `demo-setup` skill for repeatable HVE Core demonstrations. The skill creates a customer persona, guides an accelerated Design Thinking session, scaffolds a browser-ready prototype, and produces presenter and recording guidance.

This update also aligns the contribution with the current repository distribution model by synchronizing plugin membership, generated reference documentation, and behavior-conformance coverage.

## Related Issue(s)

No linked issue.

## Type of Change

**Code & Documentation:**

* [ ] Bug fix (non-breaking change fixing an issue)
* [x] New feature (non-breaking change adding functionality)
* [ ] Breaking change (fix or feature causing existing functionality to change)
* [x] Documentation update

**Infrastructure & Configuration:**

* [ ] GitHub Actions workflow
* [ ] Linting configuration (markdown, PowerShell, etc.)
* [ ] Security configuration
* [ ] DevContainer configuration
* [ ] Dependency update

**AI Artifacts:**

* [x] Reviewed contribution with `hve-builder` and addressed all actionable findings
* [ ] Copilot instructions (`.github/instructions/*.instructions.md`)
* [ ] Copilot prompt (`.github/prompts/*.prompt.md`)
* [ ] Copilot agent (`.github/agents/*.agent.md`)
* [x] Copilot skill (`.github/skills/*/SKILL.md`)
* [ ] Copilot hook (`.github/hooks/*/*.json`)
* [x] Eval spec added/updated for changed AI artifacts (`evals/`)

**Other:**

* [ ] Script/automation (`.ps1`, `.sh`, `.py`)
* [ ] Other (please describe):

## Sample Prompts (for AI Artifact Contributions)

**User Request:**

```text
/demo-setup customer="Northwind Utilities" industry="energy" problem="Dispatchers cannot reconcile outage and crew status quickly"
```

**Execution Flow:**

1. Generate a customer persona brief and domain-specific conversation seeds.
2. Initialize a frozen demo coaching state and guide accelerated Design Thinking Methods 1 through 6.
3. Derive an experiment from the session evidence and scaffold a zero-install prototype at Method 7.
4. Produce presenter guidance and a narrated video script.

**Output Artifacts:**

* Customer persona brief and coaching state
* Experiment card and browser-ready HTML, CSS, JavaScript, and fixture data
* Basic page-view, click, and task-timing telemetry
* Presenter README and narrated demo video script

[...]

PR #2404, opened 2026-07-06
feat(skills): add demo-video skill for narrated video assembly

auyidi1:
## Summary

Adds an experimental **`demo-video`** skill that assembles ordered frames or clips with per-segment narration into a narrated MP4 via FFmpeg. It closes the video-assembly gap in the Design Thinking demo workflow by bridging existing skills: `tts-voiceover` (narration WAVs) and `vscode-playwright` (frame capture).

## What's included

- **`SKILL.md`** — overview, `segments.yml` manifest schema, quick start, parameters reference, and the reuse bridge to `tts-voiceover` / `vscode-playwright`, plus FFmpeg prerequisites.
- **`scripts/assemble_video.py`** — the assembler CLI (`--manifest`, `--output`, `--fps`, `--resolution`, `-v/--verbose`). Validates the manifest, resolves each segment's visual (frame image or clip) and narration WAV, derives per-segment duration from the narration length via `ffprobe` (overridable with `duration:`), renders normalized per-segment MP4s, and concatenates them. Uses `subprocess.run` with list args and no `shell=True`.
- **`scripts/assemble-video.sh`** + **`scripts/Invoke-AssembleVideo.ps1`** — dual bash/PowerShell wrappers mirroring the `video-to-gif` convention; the bash wrapper runs `uv sync` and forwards to the Python CLI.
- **`examples/`** — an example `segments.yml` manifest and a README describing the local fixture setup (fixtures are intentionally kept out of source control).
- **Tests** — `pytest` suite (parse/validation, duration default+override, frame/clip branches, error paths, subprocess safety), an Atheris fuzz harness with corpus fixtures, and a Pester helper test.
- **Registration** — added to the `experimental` and `hve-core-all` collections with regenerated plugin outputs.

## Manifest schema

```yaml
output: ./output/demo.mp4
resolution: 1280x720
fps: 24
segments:
  - type: frame
    visual: ./assets/intro.png
    narration: ./assets/intro.wav
    duration: 3.0
  - type: clip
    clip: ./assets/interaction.mp4
    narration: ./assets/interaction.wav
```

## Validation

- `npm run validate:skills` — pass
- `npm run test:py` — 13 passed
- `npm run lint:collections-metadata` / `plugin:validate` — pass
- `npm run lint:frontmatter` — pass (demo-video files clean)
- `npm run validate:copyright` — pass

## Notes / follow-ups

- **Runtime dependency:** FFmpeg + `ffprobe` must be on PATH. The end-to-end smoke run was not executed here because FFmpeg is not installed in the dev container.
- The `demo-setup` orchestration (PR #1623) that would auto-generate the demo script and capture prototype frames is pending; until it merges, the DT → demo-video wiring is manual.

## Checklist

- [ ] Reviewed and validated by a qualified human reviewer


.github/skills/experimental/demo-video/scripts/assemble_video.py
```diff
@@ -0,0 +1,491 @@
...
+    if not isinstance(data, dict):
+        raise ManifestError("Manifest root must be a YAML mapping")
+    return data
+
+
+def _validate_manifest(
+    data: dict[str, Any], manifest_path: Path
+) -> tuple[dict[str, Any], list[dict[str, Any]]]:
+    """Validate manifest structure and return normalized values."""
+    allowed_top_level_keys = {"output", "resolution", "fps", "segments"}
```
rezatnoMsirhC, line 102: `_validate_manifest` accepts `manifest_path: Path` as a second parameter but never references it in the function body. Every `ManifestError` message identifies the problem by segment index but omits the file path, which would aid diagnostics when running tests or processing multiple manifests.

**Option A:** Use the parameter in error messages:
```python
raise ManifestError(f"{manifest_path}: segment #{index} must be a YAML mapping")
```

**Option B:** Drop the unused parameter and update call sites:
```python
def _validate_manifest(
    data: dict[str, Any],
) -> tuple[dict[str, Any], list[dict[str, Any]]]:
```

.github/skills/experimental/demo-video/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+### Segment fields
+
+* `type` identifies whether the segment is a still image (`frame`) or a motion clip (`clip`)
+* `visual` points to an image file for a frame segment
+* `clip` points to a motion clip file for a clip segment
+* `narration` points to the WAV file generated from narration text (the script also accepts `narration_wav` as an alias)
+* `duration` is optional and overrides the inferred duration when you want a fixed segment length
+
+## Quick Start
+
```
rezatnoMsirhC, line 53: "planned" is a stale qualifier: both wrappers are included in this PR. Suggest removing it:

```markdown
Use the bash or PowerShell wrappers to invoke the assembler from the skill directory.
```

.github/skills/experimental/demo-video/tests/Invoke-AssembleVideo.Tests.ps1
```diff
@@ -0,0 +1,37 @@
...
+# Copyright (c) 2026 Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+<#!
+.SYNOPSIS
+    Pester tests for the demo-video PowerShell wrapper.
+.DESCRIPTION
+    Covers wrapper parameter handling and argument forwarding without invoking FFmpeg.
+#>
+
```
rezatnoMsirhC, line 11: `<#!` is not standard PowerShell comment-based help syntax. The `!` causes `Get-Help` to skip the block, so `.SYNOPSIS` and `.DESCRIPTION` are invisible to tooling. Change the opener to `<#`:

```powershell
<#
.SYNOPSIS
    Pester tests for the demo-video PowerShell wrapper.
.DESCRIPTION
    Covers wrapper parameter handling and argument forwarding without invoking FFmpeg.
#>
```

rezatnoMsirhC (approved):

bindsi (approved):
@auyidi1 if you don´t mind you can add the additional context from our chat yesterday to the readme as well. Thanks for this great skill

.github/skills/experimental/demo-video/examples/README.md
bindsi, line 1: @auyidi1 if you don´t mind you can add the additional context from our chat yesterday to the readme as well. Thanks for this great skill

jkim323 (approved):

chaosdinosaur (dismissed):
Thanks for the `demo-video` skill — the assembler/wrapper design is clean, the subprocess handling is injection-safe (list args, no `shell=True`, `yaml.safe_load`), and the repo Python-skill conventions (ruff/pytest/fuzz/`uv.lock`, dual wrappers, copyright headers) are all met. A few items I didn't see raised elsewhere on the PR:

- **Blocking:** the shipped `examples/segments.yml` uses `./examples/fixtures/...` paths, but the assembler resolves relative paths against the manifest's own directory (`examples/`), so they expand to `examples/examples/fixtures/...` and don't match the fixture layout `examples/README.md` documents — the flagship example fails on first run. Changing the manifest paths to `./fixtures/...` (and `output: ./output/demo.mp4`) fixes it; please verify end-to-end since the PR notes the smoke run was skipped.
- The `npm run test:py 13 passed` figure in the PR body isn't reproducible — the two discoverable pytest files contain 11 test functions; please re-run and update the count.
- `tests/DemoVideoHelpers.Tests.ps1` doesn't mirror its system under test (`Invoke-AssembleVideo.ps1`); renaming to `Invoke-AssembleVideo.Tests.ps1` matches the Pester convention.
- Minor: `SKILL.md`'s Manifest Schema omits the top-level `output`/`resolution`/`fps` keys the code and example use; an empty-string `narration` skips the `narration_wav` fallback; and `_validate_manifest` returns the segments list redundantly.

> [!NOTE]
> This review was AI-assisted and should be independently validated by a qualified human reviewer before acting on any finding.

bindsi (approved):

.github/skills/experimental/demo-video/examples/segments.yml
```diff
@@ -0,0 +1,11 @@
...
+resolution: 1280x720
+fps: 24
+segments:
+  - type: frame
+    visual: ./examples/fixtures/intro.png
+    narration: ./examples/fixtures/intro.wav
+    duration: 3.0
+  - type: clip
+    clip: ./examples/fixtures/interaction.mp4
+    narration: ./examples/fixtures/interaction.wav
```
rezatnoMsirhC: The assembler resolves manifest-relative paths against `manifest_path.parent` (the `examples/` directory). The paths in this file use the `./examples/...` prefix, which causes them to resolve to `<skill-root>/examples/examples/...` — a location that does not exist. The example will fail on first run.

The `examples/README.md` documents fixtures at `examples/fixtures/`, which means the manifest (located at `examples/segments.yml`) should reference them as `./fixtures/...`.

Suggested fix:

```yaml
output: ./output/demo.mp4
resolution: 1280x720
fps: 24
segments:
  - type: frame
    visual: ./fixtures/intro.png
    narration: ./fixtures/intro.wav
    duration: 3.0
  - type: clip
    clip: ./fixtures/interaction.mp4
    narration: ./fixtures/interaction.wav
```

.github/skills/experimental/demo-video/tests/Invoke-AssembleVideo.Tests.ps1
rezatnoMsirhC, line 1: Nit: the repo Pester convention names test files after the script they cover. Since this file tests `scripts/Invoke-AssembleVideo.ps1`, consider renaming it to `Invoke-AssembleVideo.Tests.ps1` to make the source-to-test mapping clear at a glance.

.github/skills/experimental/demo-video/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+Use the bash or PowerShell wrappers to invoke the assembler from the skill directory.
+
+```bash
+scripts/assemble-video.sh --manifest examples/segments.yml --output ./output/demo.mp4
+```
+
```
rezatnoMsirhC, line 59: The Manifest Schema section only shows the `segments` list. The three top-level keys that the assembler validates and uses (`output`, `resolution`, `fps`) are not documented, so a user reading SKILL.md alone would not know they exist or what their defaults are. The PR description's own schema example includes all three.

Consider expanding the schema example and field table to cover them:

```yaml
output: ./output/demo.mp4   # optional; path for the assembled MP4
resolution: 1280x720        # optional; default 1280x720
fps: 24                     # optional; default 24
segments:
  - type: frame
    ...
```

And add field descriptions alongside the existing segment-field bullet list:

- `output` sets the destination path for the assembled MP4 (resolved relative to the manifest; defaults to `./demo.mp4`)
- `resolution` controls output width and height in `WIDTHxHEIGHT` form (default `1280x720`)
- `fps` sets the frame rate applied when rendering each segment (default `24`)

jkim323 (approved):

chaosdinosaur (commented):
Re-reviewing at `2f2d912` — the `type`/source enforcement, `--fps` validation, non-string-key handling, and tighter fuzz scope from the follow-up commits all look good and are covered by new tests. I'm not repeating the items already open on the PR (the `examples/segments.yml` path-doubling thread and the Pester filename / SKILL.md top-level-keys notes from @rezatnoMsirhC still apply and remain unresolved — the example still fails on first run). A few net-new items I didn't see raised elsewhere:

- **Please verify (from the concat change):** the final concat now uses `-c copy` (per the earlier suggestion) instead of re-encoding. With fractional per-segment durations and fixed-size AAC frames, stream-copy concat can drift A/V sync or leave gaps at segment seams that the previous re-encode masked. Since FFmpeg isn't in the dev container and the smoke run is still skipped, this primary output path is untested — please confirm the final MP4 stays in sync end-to-end, or keep a re-encode fallback for the concat step.
- **Add a shipped-example regression test:** a test that loads `examples/segments.yml` and asserts the resolved fixture paths match the README layout would guard the open path-doubling issue from recurring.
- **PR body count:** `npm run test:py 13 passed` isn't reproducible — there are now 14 discoverable test functions for this skill (12 in `test_assemble_video.py` + 2 in `fuzz_harness.py`). Please re-run and update the figure (or note it''s a repo-wide aggregate).
- **Minor:** an empty-string `narration:` still skips the `narration_wav` fallback (only `None`/absent falls back); and `_validate_manifest` returns the segments list a second time even though the config dict it returns already contains `segments`.

Also note the branch is currently `BEHIND` `main` and the four approvals predate the still-open 07-15 findings, so a fresh look after the fixes would be worthwhile.

> [!NOTE]
> This review was AI-assisted and should be independently validated by a qualified human reviewer before acting on any finding.

bindsi (approved):

PR #2466, opened 2026-07-14
feat(skills): add label margin and font overrides to arrow_flow in powerpoint skill

auyidi1:
## Summary

Adds optional styling knobs to the `arrow_flow` element in the PowerPoint skill so chevron labels can fit on a single line without manual PowerPoint edits after a build.

### What changed

- `add_arrow_flow_element` now honors optional element-level keys, each overridable per item:
  - `label_margin` — chevron text-frame left/right margin (inches); reduce to keep labels on one line
  - `gap` — spacing between chevrons (inches, default `0.3`)
  - `font` / `font_size` (`size` per item) — label font family/size
  - `font_color` (`color_text` per item) — label color
- Defaults preserve the existing output (Segoe UI, 14pt, `#F8F8FC`, 0.3in gap, python-pptx default margins).
- Documented the new knobs in `content-yaml-template.md`.

### Why

Long labels wrapped onto two lines and overlapped the chevron outline. Previously the only fix was editing the built `.pptx` by hand, which is lost on rebuild. These knobs make it settable from content YAML.

### Testing

- Added tests covering margin/font overrides and per-item precedence.
- `ruff check` clean; full `test_build_deck.py` suite passes (127 passed).


WilliamBerryiii (approved):

.github/skills/experimental/powerpoint/scripts/build_deck.py
```diff
@@ -540,16 +546,22 @@ def add_arrow_flow_element(slide, elem, colors, typography):
...
         p = tf.paragraphs[0]
         p.text = item["label"]
         p.alignment = ALIGNMENT_MAP["center"]
         run = p.runs[0]
-        run.font.name = "Segoe UI"
-        run.font.size = Pt(14)
-        apply_color_to_font(run.font.color, resolve_color("#F8F8FC"))
+        run.font.name = item.get("font", default_font)
+        run.font.size = Pt(item.get("size", default_size))
+        apply_color_to_font(
```
rezatnoMsirhC, line 559: Nit: the per-item override keys for font size and color (`size`, `color_text`) differ from their element-level counterparts (`font_size`, `font_color`). `label_margin` and `font` use the same name at both levels, which makes these two exceptions stand out.

`color_text` is used by other element types in this file, so keeping it may be the right call for cross-element consistency. For `size` vs `font_size`, there is no such cross-element precedent. Aligning it to `font_size` at both levels would make the YAML contract easier to learn, and since this is a new feature the change carries no backward-compatibility risk.

No action required if the asymmetry is intentional.

rezatnoMsirhC (approved):

WilliamBerryiii (approved):

rezatnoMsirhC (approved):

PR #2540, opened 2026-07-28
fix(hooks): correct telemetry payload handling and cross-platform entry points

### LINKED ISSUE #2538: fix(hooks): environment compatibility bugs in telemetry hook
### Component

Scripts

### Bug Description

Telemetry hook is blocked on users with different environments (powershell and bash versions)

### Expected Behavior

Hook should succeed

### Steps to Reproduce

_No response_

### Additional Context

_No response_

vyta:
# Pull Request

This PR supersedes #2532 

## Description

Local telemetry recorded incomplete and sometimes wrong data depending on which surface fired the hook and which shell ran it. This PR fixes the cross-surface payload handling in the shared telemetry engine, repairs the PowerShell and Bash entry points, and makes the report and cleanup tools behave predictably when scoped to a single store.

Kudos to @chaosdinosaur and @auyidi1 for identifying powershell/bash incompatibilities and fixes!

## Related Issue(s)

Fixes #2538

## Type of Change

Select all that apply:

**Code & Documentation:**

* [x] Documentation update

* [ ] GitHub Actions workflow
* [ ] Security configuration
* [ ] Dependency update

**AI Artifacts:**

* [x] Copilot hook (`.github/hooks/*/*.json`)

>

**Other:**

## Testing

Automated:

* `uv run pytest -q` in `.github/hooks/shared/telemetry` — **65 passed**.
* `uv run ruff check .` in `.github/hooks/shared/telemetry` — **All checks passed**.

New regression coverage:

* Snake_case `tool_response` payload infers `PostToolUse` (previously `PreToolUse`).
* Snake_case `agent_name` + `stop_reason` infers `SubagentStop` (previously `SubagentStart`).
* `tool_response` with nested `text_result_for_llm` produces a correct `tool_response_len`.
* VS Code subagent payload carrying only `agent_type` yields a named agent on the stack.
* A session whose first event is not `SessionStart` still registers its telemetry directory.

Test hygiene: added an autouse fixture pinning `HVE_HOME` to `tmp_path`. Registry and launcher tests previously wrote into the real user home directory.

Diff-based assessment: verified that `telemetry-collector.sh` and `Invoke-TelemetryCollector.ps1` contain no payload key logic of their own, so consolidating aliases in `_telemetry_core.py` leaves no second source of truth.

Alias verification: every entry in `PAYLOAD_ALIASES` was checked against the published GitHub and VS Code hooks references. An undocumented `toolResponse` alias was removed.

Manual:
* Validated powershell scripts on windows (Powershell 5.1 for collector, 7+ for other ps1)
* Validated bash in WSL

### Required Checks

### AI Artifact Contributions

### Required Local Checks

## Security Considerations

* [ ] Any new dependencies have been reviewed for security issues (N/A — no dependency changes)
* [ ] Security-related scripts follow the principle of least privilege (N/A — no `scripts/security/` changes)

### DIFF
--- a/.github/hooks/shared/telemetry/Invoke-TelemetryClean.ps1
+++ b/.github/hooks/shared/telemetry/Invoke-TelemetryClean.ps1
@@ -39,12 +39,19 @@ $ErrorActionPreference = 'Stop'
 #region Resolve repo root
 # Match the collector and anchor on the caller's workspace: the script itself
 # may live outside the repo it cleans, so its location says nothing.
-$RepoRoot = $env:HVE_REPO_ROOT
-if (-not $RepoRoot -and (Get-Command git -ErrorAction SilentlyContinue)) {
-    try { $RepoRoot = & git rev-parse --show-toplevel 2>$null } catch { $RepoRoot = $null }
-}
-if (-not $RepoRoot) {
-    $RepoRoot = (Get-Location).Path
+# Only needed to build the default telemetry path; an explicit -Path makes the
+# whole lookup unnecessary, including the git subprocess.
+if (-not $Path) {
+    $RepoRoot = $env:HVE_REPO_ROOT
+    if (-not $RepoRoot -and (Get-Command git -ErrorAction SilentlyContinue)) {
+        try { $RepoRoot = & git rev-parse --show-toplevel 2>$null } catch { $RepoRoot = $null }
+    }
+    if (-not $RepoRoot) {
+        # ProviderPath, not .Path: the latter is PowerShell-qualified ('Work:\',
+        # 'HKLM:\SOFTWARE'). A non-filesystem provider survives this as a path with
+        # no drive qualifier, which the guard below rejects.
+        $RepoRoot = $PWD.ProviderPath
+    }
 }
 #endregion Resolve repo root
 
@@ -66,6 +73,21 @@ if (-not (Test-Path $CorePy)) {
 }
 
 $TelemetryDir = if ($Path) { $Path } else { Join-Path $RepoRoot '.copilot-tracking' 'telemetry' }
+
+# Hand Python a native absolute path. -Path may be relative or on a PSDrive, and
+# PowerShell does not sync [Environment]::CurrentDirectory with Set-Location, so
+# the child process would resolve a relative path against a different directory
+# than the one shown in the confirmation prompt below. Guarding the resolved path
+# covers both branches: a non-filesystem location leaves a drive-less path
+# ('HKEY_LOCAL_MACHINE\SOFTWARE\...') that Python would silently take as relative
+# to its own working directory and delete from there.
+$Provider = $null
+$TelemetryDir = $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath(
+    $TelemetryDir, [ref]$Provider, [ref]$null)
+if ($Provider.Name -ne 'FileSystem') {
+    Write-Error "Telemetry path resolved to '$TelemetryDir' on the $($Provider.Name) provider from location '$($PWD.Path)'; re-run from a filesystem directory, pass an absolute -Path, or set HVE_REPO_ROOT."
+    exit 1
+}
 
 # An explicit -Path narrows the scope. The generated launcher always passes
 # -AllDirs, so announce the override rather than silently widening or dropping
@@ -94,5 +116,14 @@ $CliArgs = @('clean')
 if ($UseAllDirs) { $CliArgs += '--all-dirs' }
 if ($DryRun) { $CliArgs += '--dry-run' }
 
+# Scope the variable to this call, as the bash wrapper's prefix assignment does:
+# leaking it would silently redirect the collector and report scripts for the
+# rest of the session.
+$PreviousTelemetryDir = $env:HVE_TELEMETRY_DIR
 $env:HVE_TELEMETRY_DIR = $TelemetryDir
-& $Python.Source $CorePy @CliArgs
+try {
+    & $Python.Source $CorePy @CliArgs
+}
+finally {
+    $env:HVE_TELEMETRY_DIR = $PreviousTelemetryDir
+}
--- a/.github/hooks/shared/telemetry/Invoke-TelemetryClean.ps1
+++ b/.github/hooks/shared/telemetry/Invoke-TelemetryClean.ps1
@@ -17,7 +17,8 @@
     registry, plus the generated launchers, report, and registry in the HVE home
     directory.
 .PARAMETER Path
-    Telemetry directory. Default: <repo>/.copilot-tracking/telemetry
+    Telemetry directory. Scopes the cleanup to this directory alone, overriding
+    -AllDirs. Default: <repo>/.copilot-tracking/telemetry
 .PARAMETER DryRun
     List what would be removed without deleting anything.
 .PARAMETER Force
@@ -36,12 +37,14 @@ param(
 $ErrorActionPreference = 'Stop'
 
 #region Resolve repo root
+# Match the collector and anchor on the caller's workspace: the script itself
+# may live outside the repo it cleans, so its location says nothing.
 $RepoRoot = $env:HVE_REPO_ROOT
 if (-not $RepoRoot -and (Get-Command git -ErrorAction SilentlyContinue)) {
-    try { $RepoRoot = & git -C $PSScriptRoot rev-parse --show-toplevel 2>$null } catch { $RepoRoot = $null }
+    try { $RepoRoot = & git rev-parse --show-toplevel 2>$null } catch { $RepoRoot = $null }
 }
 if (-not $RepoRoot) {
-    $RepoRoot = (Resolve-Path (Join-Path $PSScriptRoot '../..')).Path
+    $RepoRoot = (Get-Location).Path
 }
 #endregion Resolve repo root
 
@@ -62,12 +65,20 @@ if (-not (Test-Path $CorePy)) {
     exit 1
 }
 
-$TelemetryDir = if ($Path) { $Path } else { Join-Path $RepoRoot '.copilot-tracking/telemetry' }
+$TelemetryDir = if ($Path) { $Path } else { Join-Path $RepoRoot '.copilot-tracking' 'telemetry' }
+
+# An explicit -Path narrows the scope. The generated launcher always passes
+# -AllDirs, so announce the override rather than silently widening or dropping
+# a destructive scope flag.
+$UseAllDirs = [bool]$AllDirs -and -not $Path
+if ($AllDirs -and $Path) {
+    Write-Warning "Ignoring -AllDirs because -Path was given; scope is '$TelemetryDir'."
+}
 
 # Prompt before destructive deletion. Skipped on -DryRun (non-destructive) and
 # bypassed with -Force (required for non-interactive use).
 if (-not $DryRun -and -not $Force) {
-    $scope = if ($AllDirs) {
+    $scope = if ($UseAllDirs) {
         'ALL registered telemetry stores plus the user-level HVE home directory'
     } else {
         $TelemetryDir
@@ -80,7 +91,7 @@ if (-not $DryRun -and -not $Force) {
 
 # Build the clean-mode argument list mirroring the bash wrapper's flags.
 $CliArgs = @('clean')
-if ($AllDirs) { $CliArgs += '--all-dirs' }
+if ($UseAllDirs) { $CliArgs += '--all-dirs' }
 if ($DryRun) { $CliArgs += '--dry-run' }
 
 $env:HVE_TELEMETRY_DIR = $TelemetryDir
--- a/.github/hooks/shared/telemetry/Invoke-TelemetryCollector.ps1
+++ b/.github/hooks/shared/telemetry/Invoke-TelemetryCollector.ps1
@@ -15,7 +15,11 @@
     Runs via: Copilot agent hook (stdin JSON, stdout JSON)
 #>
 [CmdletBinding()]
-param()
+param(
+    [Parameter(Mandatory = $false, ValueFromPipeline = $true)]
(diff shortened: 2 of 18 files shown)


.github/hooks/shared/telemetry/Invoke-TelemetryCollector.ps1
```diff
@@ -56,18 +55,23 @@ if (-not $Python) {
...
     New-Item -ItemType Directory -Path $TelemetryDir -Force | Out-Null
 }
 
-# Delegate all JSON processing to the shared Python telemetry engine
-$RawInput = $input | Out-String
+# Delegate all JSON processing to the shared Python telemetry engine.
+# The hook contract is raw process stdin, not a PowerShell object pipeline:
+# [CmdletBinding()] makes this an advanced script, which rejects pipeline input
+# it cannot bind to a parameter, leaving $input empty.
+$RawInput = if ([Console]::IsInputRedirected) { [Console]::In.ReadToEnd() } else { '' }
```
jkim323: **Blocking: the new stdin path silently corrupts non-ASCII telemetry on Windows.**:

`[Console]::In` fixes the empty-`$input` bug but leaves two Windows encoding boundaries at their defaults, so non-ASCII payload text is silently corrupted:
- `[Console]::In` decodes with `[Console]::InputEncoding` (OEM code page, `IBM437` here), not UTF-8.
- Line 96 pipes with `$OutputEncoding`, which is `ASCIIEncoding` in Windows PowerShell 5.1 — the host this PR enables by dropping `#Requires -Version 7.4`.
- Python may decode redirected stdin using the Windows locale encoding rather than UTF-8, which also affects `pwsh`.

Verified defaults:
```
WinPS 5.1.26100.8875
OutputEncoding: System.Text.ASCIIEncoding
Console::InputEncoding: IBM437
```
Reproduced end-to-end against the PR head with the payload 
```
{"prompt":"café — 日本語"}:
// powershell.exe 5.1
{"event": "UserPromptSubmit", "prompt": "caf?? ??? ?????????"}
// pwsh 7
{"event": "UserPromptSubmit", "prompt": "cafâ”œ...â‚§"}   // mojibake
```
The resulting JSON remains valid, so the corruption is silently persisted to sessions-*.jsonl and downstream reports.

**Suggested fix:** 
Verified working on both powershell.exe 5.1 and pwsh 7 (both hosts recorded café — 日本語 intact). $OutputEncoding must be assigned at global scope — a script-scoped assignment is ignored by the native-command pipeline in Windows PowerShell 5.1.

.github/hooks/shared/telemetry/_telemetry_core.py
```diff
@@ -929,11 +969,13 @@ def _mode_collect() -> int:
     if entry is None:
         return 0
 
+    # Register before the first write of the day so a store whose SessionStart
+    # was never delivered (telemetry enabled mid-session, or a surface that
+    # omits the event) still reaches cross-project reports. Bounded to one
+    # registry read per store per day plus each session start.
+    first_write = not log_file.exists()
```
jkim323, line 1317: **Concurrency: `first_write` registration is no longer single-threaded and can truncate the generated launchers:**

`first_write` is check-then-act, so the launcher write is no longer single-threaded. Measured on this branch with 12 barrier-synced non-`SessionStart` collectors: 12/12 entered the gate and called `write_report_launchers()`, across 3/3 trials. The old `SessionStart` gate bounded this to one process by construction; tool and subagent events have no such bound.

This introduces three concurrency hazards:
| # | Issue | Impact |
|---|---|---|
| 1 | `Path.write_text()` truncates then rewrites in place | Reader can see an empty/partial launcher; interruption leaves it broken |
| 2 | POSIX `write_text()` → `chmod(0o755)` are separate steps | Launcher briefly exists but is not executable |
| 3 | Repo-checkout vs. extension installs share `hve_home()` but embed different `script_dir` | Last-writer-wins on `REPORT_SCRIPT`, and it can flip day to day |

**Suggested fix:**
Make launcher writes atomic so a concurrent reader never observes a partial file:
```
def _atomic_write(path: Path, text: str, mode: int | None = None) -> None:
    """Publish a file atomically so concurrent hooks never observe a partial one.

    mkstemp guarantees a unique name through O_EXCL rather than a naming
    convention: PIDs are unique only within a namespace, and HVE_HOME can be a
    mount shared by a container and its host.
    """
    fd, tmp_name = tempfile.mkstemp(dir=str(path.parent), prefix=f".{path.name}.", suffix=".tmp")
    tmp = Path(tmp_name)
    try:
        with os.fdopen(fd, "w", encoding="utf-8") as handle:
            handle.write(text)
        if mode is not None:
            os.chmod(tmp, mode)
        os.replace(tmp, path)
    finally:
        # No-op on success. On failure this keeps HVE_HOME free of orphaned
        # .tmp files, which the clean mode's name allow-list never removes.
        tmp.unlink(missing_ok=True)
```

docs/customization/local-telemetry.md
```diff
@@ -30,6 +30,10 @@ Events currently captured include:
 
 At stop time, telemetry also appends a session summary with model and token usage when available.
 
+## Prerequisites
+
+Ensure you have Bash 3.5+ (Linux/macOS) or PowerShell (Windows) and python available, depending on your OS for running the telemetry hooks. To utilize cleanup or report generation on Windows, ensure you have PowerShell 7.4+ installed.
```
jkim323: **Documented Bash prerequisite names a version that does not exist:**
There's no Bash 3.5 — releases go 3.2 → 4.0. The floor this PR actually establishes is 3.2, macOS's system Bash: `generate-telemetry-report.sh` swapped its `declare -A` for a string sentinel precisely to run there, and I confirmed no `declare -A`, `mapfile`, or `readarray` remain in any telemetry script. The PowerShell floor is also understated — the collector now runs on Windows PowerShell 5.1, which is the point of dropping `#Requires -Version 7.4`.

**Suggested fix:**
```
## Prerequisites

The telemetry hooks need Python 3 plus a shell: Bash 3.2 or later on Linux and
macOS, or Windows PowerShell 5.1 or later on Windows. Cleanup and report
generation on Windows additionally require PowerShell 7.4 or later; on Linux and
macOS they also require `jq`.
```

jkim323: Separately: the testing notes say bash was validated in WSL, which is GNU userland. Bash 3.2 covers the shell half of macOS support, but no BSD coreutils were exercised — has `generate-telemetry-report.sh` --date all been run once on stock macOS? I scanned through it for GNU-isms and it looks clean, just double checking.

.github/hooks/shared/telemetry/Invoke-TelemetryClean.ps1
```diff
@@ -36,12 +37,14 @@ param(
...
+# Match the collector and anchor on the caller's workspace: the script itself
+# may live outside the repo it cleans, so its location says nothing.
 $RepoRoot = $env:HVE_REPO_ROOT
 if (-not $RepoRoot -and (Get-Command git -ErrorAction SilentlyContinue)) {
-    try { $RepoRoot = & git -C $PSScriptRoot rev-parse --show-toplevel 2>$null } catch { $RepoRoot = $null }
+    try { $RepoRoot = & git rev-parse --show-toplevel 2>$null } catch { $RepoRoot = $null }
 }
 if (-not $RepoRoot) {
-    $RepoRoot = (Resolve-Path (Join-Path $PSScriptRoot '../..')).Path
+    $RepoRoot = (Get-Location).Path
```
jkim323: `(Get-Location).Path` returns the PowerShell-qualified location, not an OS path, and `$TelemetryDir` is handed to Python via `$env:HVE_TELEMETRY_DIR`. Measured:

| Location | `Provider.Name` | `(Get-Location).Path` | `$PWD.ProviderPath` |
|---|---|---|---|
| `Work:\` (mapped to a real dir) | `FileSystem` | `Work:\` | `C:\Users\...\hve-core\` |
| `HKLM:\SOFTWARE` | `Registry` | `HKLM:\SOFTWARE` | `HKEY_LOCAL_MACHINE\SOFTWARE` |
| `Cert:\CurrentUser` | `Certificate` | `Cert:\CurrentUser` | `\CurrentUser` |

A mapped PSDrive is the realistic case: it's a `FileSystem` provider, so a provider check alone passes, but `.Path` still yields `Work:\.copilot-tracking\telemetry`, which Python cannot resolve. And `ProviderPath` alone is not sufficient either — `HKEY_LOCAL_MACHINE\SOFTWARE` is drive-less and separator-less, so Python resolves it *relative to the CWD* and silently creates a real directory tree; `\CurrentUser` is root-relative. Both guards are needed, or a loud failure becomes a silent wrong-target one on a destructive path.

**Suggested fix:**
```
#region Resolve repo root
# Only needed to build the default telemetry path; an explicit -Path makes the
# whole lookup unnecessary, including the git subprocess.
if (-not $Path) {
    $RepoRoot = $env:HVE_REPO_ROOT
    if (-not $RepoRoot -and (Get-Command git -ErrorAction SilentlyContinue)) {
        try { $RepoRoot = & git rev-parse --show-toplevel 2>$null } catch { $RepoRoot = $null }
    }
    if (-not $RepoRoot) {
        # .Path may be provider-qualified (for example, 'Work:\' or
        # 'HKLM:\SOFTWARE'). ProviderPath resolves filesystem-backed PSDrives, but
        # a non-filesystem provider still does not produce an OS filesystem path.
        if ($PWD.Provider.Name -ne 'FileSystem') {
            Write-Error "Current location '$($PWD.Path)' is not filesystem-backed; re-run from a filesystem directory, pass -Path, or set HVE_REPO_ROOT."
            exit 1
        }
        $RepoRoot = $PWD.ProviderPath
    }
}
#endregion Resolve repo root
```

jkim323: I believe same changes needs to be made to `Invoke-TelemetryReport.ps1:70`. Please double check this thanks!

jkim323 (dismissed):
Left couple of comments, happy to review after those gets looked at! Also, please ensure you check the **Required Local Checks**. Thanks Vy and the team for working on this!

.github/hooks/shared/telemetry/_telemetry_core.py
vyta: Risk is low since these writes don't really happen often but can implement concurrency to cover our bases for other writes too

renato-marciano:
Was getting an error trying to use hve core in Copilot App on Windows. Using this branch fixed it. Thanks!

.github/hooks/shared/telemetry/Invoke-TelemetryCollector.ps1
vyta: Good catch on the diagnosis, addressed the corruption.

One correction, `$OutputEncoding` does *not* need to be assigned at global scope — a script-scoped assignment is honored by the native-command pipeline on 5.1. I checked this specifically:

Piping `héllo` into `python -c "print(sys.stdin.buffer.read().hex())"`:

| Host | no assignment | script-scope | `$global:` |
|---|---|---|---|
| PS 5.1 | `683f6c6c6f` | `68c3a96c6c6f` | `68c3a96c6c6f` |
| pwsh 7.6 | `68c3a96c6c6f` | `68c3a96c6c6f` | `68c3a96c6c6f` |

Script-scoped and `$global:` are byte-identical. Two reasons:

1. The native-command processor resolves `$OutputEncoding` by ordinary dynamic scope lookup, not a `global:`-qualified path. Assigning it in a *child* scope flipped the piped bytes to UTF-8 inside that scope and they reverted to `683f...` on exit, with `$global:OutputEncoding` never modified, the opposite of what a `global:`-qualified read would produce.
2. Under `-File` invocation (how the hook host launches this), a script's top-level scope *is* the global scope: `$script:x` is readable as `$global:x`, and `Get-Variable -Scope Script`/`-Scope Global` return the same object.

I'd rather keep it script-scoped, since `$global:` would permanently mutate the host runspace's encoding if this script is ever dot-sourced or run in a shared runspace. Happy to be shown a case where the script-scoped form fails though if you have a repro on a host where it does, I'll take the change.

.github/hooks/shared/telemetry/_telemetry_core.py
vyta: Hazards 1 and 2 are fixed and pushed. Launcher writes now go through a _write_text_atomic helper that stages to a temp file and os.replaces it, and the permission bits are applied at os.open creation time rather than as a follow-up chmod so there's no longer a window where the launcher exists but isn't yet executable.

Hazard 3 reproduces. Three _telemetry_core.py installs on this machine at distinct paths (1220 / 1383 / 1557 lines, three different checksums), all sharing one ~/.hve. generate-report.sh embeds the path of whichever install ran most recently; opening a session in any of the others flips it.

One reclassification worth recording: unlike 1 and 2, this isn't a race. Every individual write is already correct. hve_home() is per-user while script_dir is per-install, so it's last-writer-wins across sessions, not across milliseconds so atomicity can't help here. The sharper consequence is version skew: the winning launcher may be an older generator reading stores written by a newer install. Today that's harmless because the on-disk layout is stable, but it would fail silently the moment that layout changes.

Tracking in #2562 rather than growing this PR, since the launcher already fails loudly with a regenerate hint when its embedded path is gone, and self-heals on the next session. Your mkstemp/O_EXCL point is captured there too, the pid-named staging file is fine for the local case, but not for a shared HVE_HOME mount where PIDs aren't unique across namespaces.

vyta:
Thanks for the thorough review @jkim323! I've addressed your comments pending a validation on macOS. Will re-request a review once that is done

docs/customization/local-telemetry.md
vyta: yes, had a macos user try latest, as of this comment, without error

WilliamBerryiii (approved):

agreaves-ms (approved):

jkim323 (approved):

PR #2566, opened 2026-07-31
fix(docs): resolve documentation site accessibility defects and extend the runtime harness

WilliamBerryiii:
# fix(docs): resolve documentation site accessibility defects and extend the runtime harness

## Description

This branch closes out the documentation site accessibility follow-ups and extends the accessibility skill's runtime harness with the machinery those fixes needed. Three of the changes are genuine product defects on the published docs site; the rest is permanent regression coverage plus the harness work that made the defects reproducible in the first place.

> A recurring theme in this work: several behaviors previously classified as "manual only" turned out to be automatable once the right measurement was found. Text zoom is the clearest example — it had been recorded as undriveable from Playwright, but browser zoom is equivalent to a scaled viewport at a matching `deviceScaleFactor`, which makes it a deterministic matrix.

### Documentation site fixes

Three defects were reproduced against a locally served build before any fix was written, and each fix is now locked by a test proven to fail without it.

- Fixed the **search clear button** in *src/theme/SearchBar/index.jsx*. The swizzled component looked for `button[type="reset"], button[class*="clear"]`, but the upstream theme renders the control as `<button class="searchClearButton_qk4g">` with no `type` attribute. CSS attribute matching is case-sensitive, so `[class*="clear"]` never matched `searchClearButton`. The selector now carries the case-insensitive flag.
  - This one had been masked: the product code and its test shared the same faulty selector, so the test took a skip branch instead of failing.
- Fixed **navbar overlap at narrow viewports** in *src/css/custom.css*. The search plugin's container is absolutely positioned and sits outside the flex flow, so no flex rule on its siblings could resolve the overlap. The brand title is now visually hidden below 420px while retaining its accessible name.
- Fixed **search placeholder clipping under text zoom** in the same file. The keyboard shortcut badge is hidden below 996px, which restores usable width for the placeholder.

### Regression coverage

- Added *e2e/text-zoom.spec.ts*, which exercises a zoom matrix from 100% through 250% by pairing scaled viewports with matching device scale factors, then measures required placeholder width against usable width.
- Added table, search keyboard, arrow navigation, and search page status specs.
- Extended *e2e/contrast.spec.ts* so every prose link must carry a non-color cue, and *e2e/focus-management.spec.ts* so every visible focusable draws an indicator of at least 2 CSS pixels.
- Added `waitForHydration` to *e2e/_helpers/a11yInvariants.ts* and applied it across twelve specs. Docusaurus sets `data-has-hydrated` on the document element after React hydration; several specs were asserting against click handlers, focus styles, and axe results before that point, which is what produced the intermittent failures.

### Accessibility skill runtime harness

- Split AT case execution, calibration, and visual review into discrete modules under *scripts/runtime_a11y/runner/*, replacing a single monolithic path.
- Added an **ARIA-AT catalog** with a JSON Schema and SHA-pinned upstream provenance covering five patterns. The public posture is deliberately conservative: entries default to manual-only and JAWS remains human-led.
- Added `render-artifacts`, `capture-visual-review`, `run-calibration`, and `run-at-plan` subcommands, along with output-path containment checks.
- Fixed a **stray browser window leak**. The harness previously invoked `chrome.exe --version` to read the browser version; on Windows that opens a window rather than printing and exiting, and the call was made twice in one expression. Version detection now reads file metadata instead.
- Introduced a **method-adequacy taxonomy** distinguishing criteria that static analysis can decide from those it can only inform, with CI enforcement graduated by tier.

### Documentation

- Added a manual validation guide under *docs/contributing/* organizing per-behavior steps into seven workstreams with a traceability register and evidence template.
- Added a shared real screen reader testing runbook under *docs/planning/runbooks/accessibility/*.

## Related Issue(s)

None

## Type of Change

Select all that apply:

**Code & Documentation:**

* [x] Documentation update

* [ ] GitHub Actions workflow
* [ ] Security configuration
* [ ] Dependency update

**AI Artifacts:**

* [ ] Reviewed contribution with `prompt-builder` agent and addressed all feedback
* [x] Copilot prompt (`.github/prompts/*.prompt.md`)

>

**Other:**

**User Request:**

"Generate the accessibility coverage matrix for this project and render the evidence bundle."

**Execution Flow:**

The coverage matrix prompt builds a surface inventory across surfaces and interaction states, then classifies each criterion by method adequacy — whether static analysis can decide it or only inform it. It invokes the runtime harness `render-artifacts` subcommand to produce the evidence bundle. Criteria in the announcement, interaction, and adaptive-rendering classes are routed to manual verification rather than reported as automated passes. Unresolved assistive-technology work is routed to the shared real screen reader runbook.

**Output Artifacts:**

A six-file bundle: coverage matrix as JSON and Markdown, an EARL JSON-LD report, manual test plans in Markdown and YAML, and a manifest. For example, the manual plan opens with the surface and state inventory, followed by per-criterion steps and an evidence template with a four-value result vocabulary.

**Success Indicators:**

The manifest lists all six artifacts, the EARL report maps informs-only methods to `earl:cantTell` rather than a pass, and every criterion that cannot be decided statically appears in the manual plan rather than being silently marked as covered.

## Testing

Automated validation captured on this branch **before** the most recent merge from `main`:

| Check | Result |
|-------|--------|
| `npm run test:node` | Passed, 120 of 120 |
| Playwright e2e suite | Passed, 248 passed and 2 skipped |
| Docusaurus `npm run typecheck` | Passed |
| `npm run lint:md` | Passed |
| `npm run lint:frontmatter` | Passed, 823 files with no errors or warnings |

Re-run **after** the merge, on the tree as proposed:

| Check | Result |
|-------|--------|
| `npm run lint:frontmatter` | Passed |
| `npm run validate:skills` | Passed, 56 skills with no errors or warnings |

Each new lock was proven non-vacuous by confirming it fails before the corresponding fix and passes after. The contrast lock moved from 0 to 19 offenders when the fix was reverted; the focus indicator lock moved from 0 to 77; the zoom matrix moved from 4 passing to 2 failing.

The e2e suite went from 217 passing with 16 skips to 248 passing with 2 skips, and full-suite runtime dropped from 12.0 minutes with 4 failures to 2.9 minutes with none.

> [!IMPORTANT]
> The Node-based checks could not be re-run after the merge. A TLS handshake failure between this workstation and the public npm registry currently blocks dependency installation, which leaves both the repository root and `docs/docusaurus` without a complete `node_modules`. That is an environment fault on the authoring machine, not a defect in this branch. CI installs from a clean environment and should be treated as the authoritative signal for every unchecked item below.

Manual testing has not yet been performed. A manual validation pass covering the remaining behaviors is planned as a follow-up using the runbook added in this branch.

Security review of the diff found no secrets, tokens, credentials, or personal data, and no non-inclusive language. All new Python and Node modules carry Microsoft copyright and SPDX identifiers. The skill-local lockfile resolves from the public npm registry.

### Required Checks

### AI Artifact Contributions

* [ ] Used `/prompt-analyze` to review contribution
* [ ] Addressed all feedback from `prompt-builder` review

### Required Automated Checks

The following validation commands must pass before merging:

* [ ] Markdown linting: `npm run lint:md`
* [x] Frontmatter validation: `npm run lint:frontmatter`
* [x] Skill structure validation: `npm run validate:skills`
* [ ] PowerShell analysis: `npm run lint:ps`
* [ ] Eval spec schema and coverage (if AI artifacts changed): `npm run eval:lint:schema`
* [ ] Plugin freshness: `npm run plugin:generate`
* [ ] Docusaurus tests: `npm run docs:test`

## Security Considerations

## GHCP Artifact Maturity

> [!WARNING]
> This PR includes **experimental** GHCP artifacts that may have breaking changes.
>
> - `.github/agents/accessibility/accessibility-planner.agent.md`
> - `.github/agents/accessibility/accessibility-reviewer.agent.md`
> - `.github/agents/accessibility/subagents/accessibility-framework-assessor.agent.md`
> - `.github/instructions/accessibility/accessibility-identity.instructions.md`
> - `.github/prompts/accessibility/accessibility-coverage-matrix.prompt.md`
> - `.github/skills/accessibility/accessibility/SKILL.md`

| File                                        | Type         | Maturity        | Notes            |
|---------------------------------------------|--------------|-----------------|------------------|
| `accessibility-planner.agent.md`            | Agent        | ⚠️ experimental | Pre-release only |
| `accessibility-reviewer.agent.md`           | Agent        | ⚠️ experimental | Pre-release only |
| `accessibility-framework-assessor.agent.md` | Agent        | ⚠️ experimental | Pre-release only |
| `accessibility-identity.instructions.md`    | Instructions | ⚠️ experimental | Pre-release only |
| `accessibility-coverage-matrix.prompt.md`   | Prompt       | ⚠️ experimental | Pre-release only |
| `accessibility/SKILL.md`                    | Skill        | ⚠️ experimental | Pre-release only |

### GHCP Maturity Acknowledgment

- [ ] I acknowledge this PR includes non-stable GHCP artifacts
- [ ] Non-stable artifacts are intentional for this change

## Additional Notes

**Known limitation — live-region announcement capture.** Spoken verification of live regions through the guidepup and NVDA path did not capture announcements across repeated settle windows, including runs against a positive-control fixture that updated every two seconds over settle windows up to eight seconds. Timing, capture clearing, focus placement, screen reader configuration, and browser accessibility flags were each eliminated as causes. The two docs-site journeys that depend on announcements are therefore recorded as manual-boundary items rather than reported as product failures. The positive-control fixture added in *docs/docusaurus/static/a11y-positive-control.html* exists to validate the oracle before any silent live region is attributed to the product.

**Behavior change in the harness.** The runtime harness now requires `npm ci` in its skill-local directory before probe execution, replacing the previous ephemeral `npx` approach. The workflow template documents this step.

**Follow-up work.** A manual assistive-technology pass remains outstanding for the behaviors that automation cannot decide. A runbook comment with per-behavior steps will be added to this PR to support that pass.


WilliamBerryiii:
## Manual testing runbook for this PR

The automated locks in this branch prove the deterministic half of each behavior: a control resolves, an accessible name is non-empty, a measurement clears a threshold. They cannot prove what a screen reader actually speaks. This comment lists the manual pass needed to close each behavior, written so a reviewer can work straight down it.

The same content is committed in [docs/contributing/accessibility-manual-validation.md](../blob/fix/docs-a11y-followups/docs/contributing/accessibility-manual-validation.md), with the shared procedure and evidence format in the [real screen reader testing runbook](../blob/fix/docs-a11y-followups/docs/planning/runbooks/accessibility/real-screen-reader-testing.md).

### Prerequisites

- OS: Windows 11 recommended for this runbook (NVDA + Edge workflow below assumes Windows).
- Tools:
  - Node.js 24.x and npm
  - Git
  - NVDA installed (latest stable)
  - Microsoft Edge installed (authoritative browser for zoom/reflow checks)
- Network: access to npm registry (for `npm ci`).
- Clean branch workspace so the docs build matches the PR branch exactly.

### Build and serve the docs site from this PR branch

Run these commands exactly so testing is against the built output of this branch, not a stale local site.

```powershell
# 1) Get the PR branch locally
Set-Location C:\Users\wberry\src\hve-core3
git fetch origin fix/docs-a11y-followups
git switch fix/docs-a11y-followups

# 2) Install root dependencies (repo-level scripts/tests)
npm ci

# 3) Install docs-site dependencies
Set-Location docs/docusaurus
npm ci

# 4) Build static docs output from this branch
npm run build

# 5) Serve the built site (keep this terminal open)
npm run serve -- --host 127.0.0.1 --port 3000
```

Open <http://127.0.0.1:3000/> in Edge and run the manual steps below against that served URL.

### Optional automated sanity checks before manual pass

If you want a quick confidence sweep before manual NVDA checks:

```powershell
Set-Location C:\Users\wberry\src\hve-core3\docs\docusaurus
npx --no-install playwright test --workers=2 --reporter=line
```

For targeted checks tied to this PR's fixes:

```powershell
Set-Location C:\Users\wberry\src\hve-core3\docs\docusaurus
npx --no-install playwright test forced-colors.spec.ts mobile-menu.spec.ts search.spec.ts --workers=2 --reporter=line
```

### Session setup before each manual run

1. Start NVDA first.
2. Launch Edge and open <http://127.0.0.1:3000/>.
3. Confirm you are on the PR branch build (for example, by checking a known changed page/behavior from this PR).
4. Record baseline metadata once per session:
   - NVDA version
   - Browser + version
   - OS version
   - Viewport or zoom level
   - URL under test

Edge is authoritative for the zoom and reflow checks. Chromium coverage in the automated matrix does not substitute for it.

---

### 1. Search clear button has an accessible name

Automation covers the control resolving with a non-empty accessible name. The manual pass confirms what is spoken and that it works by keyboard alone.

1. Open the landing page and focus the navbar search field.
2. Type any query so the clear control appears.
3. Press `Tab` and confirm focus reaches the clear control.
4. With NVDA, confirm it announces a meaningful name and a button role, not an unlabeled control or punctuation.
5. Press `Enter` or `Space` and confirm the query clears and focus stays predictable.

**Pass:** keyboard reachable, meaningful spoken name and role, query clears.

> This one was previously masked. The product code and its test shared the same faulty selector, so the test skipped instead of failing. Worth a careful look.

---

### 2. Narrow viewport does not overlap the site brand

Automation covers no overlap and no horizontal scroll at 320px. The manual pass confirms the brand keeps its accessible name after being visually hidden.

1. Set the viewport to 320px wide and open the search field.
2. Confirm the brand and search field do not overlap and the page does not scroll horizontally.
3. With NVDA, navigate to the site brand link and confirm it still announces the site name.
4. Repeat at 420px, the boundary where the rule changes.

**Pass:** no overlap, no horizontal scroll, brand keeps a spoken name at both widths.

---

### 3. Search placeholder survives text zoom

Automation covers a zoom matrix from 100% to 250% using scaled viewports with matching device scale factors, in Chromium only.

1. In Edge at a standard desktop width, set browser zoom to 150%.
2. Confirm the placeholder is not clipped and no control overlaps another.
3. Repeat at 200% and 250%.
4. Confirm the keyboard shortcut badge is hidden rather than overlapping the placeholder at narrower widths.

**Pass:** placeholder legible and controls separated at every zoom level.

---

### 4. Search result count is announced

Automation covers the status region existing in main content with text matching the visible count, including the no-match case.

**This step is the only evidence for the spoken behavior.** Automated capture of spoken live-region output is not supported by the current harness - see the note at the end of this comment.

1. Start NVDA, then open the search results page for a query returning matches.
2. Edit the query and confirm NVDA speaks the updated count without focus moving.
3. Change to a query with no matches and confirm NVDA speaks the no-results message.

**Pass:** both the count and the no-results message are spoken.

---

### 5. Search results are reachable by keyboard

Automation covers option traversal, footer option reachability, and focus containment.

1. Open search and type a query returning several results.
2. Press `ArrowDown` repeatedly; confirm the highlight advances one option at a time and reaches the final footer option.
3. With NVDA, confirm each option and its position are announced and match what is visible.
4. Confirm `Enter` activates, `Esc` closes, and `Tab` does not navigate the page away.

**Pass:** every option reachable and announced correctly, focus never escapes unexpectedly.

---

### 6. Link cues and focus indicators

Automation covers every prose link carrying a non-color cue and every visible focusable drawing an indicator of at least 2 CSS pixels.

1. View a documentation page in grayscale; confirm links stay distinguishable from body text.
2. Tab through the page; confirm a visible focus ring on every control, including cards and the skip link.
3. Confirm the ring is not clipped by an adjacent element or the viewport edge.
4. Repeat in forced-colors mode.

**Pass:** color is never the only cue, focus visible on all sides in both modes.

---

### 7. Table and group semantics

Automation covers header scope, accessible names, and structural associations.

1. With NVDA, navigate to a documentation table and confirm it announces its name.
2. Move between cells; confirm the correct row and column headers are spoken with each cell.
3. Navigate the footer; confirm each column announces its group title before its list.

**Pass:** table names, header associations, and footer group labels all spoken correctly.

---

### 8. Focus order and heading outline

Automation covers focus order after activation and a gap-free heading outline.

1. Activate a top navigation link; confirm focus lands on page content rather than returning to the skip link.
2. Open the topics menu; confirm arrow, home, and end move within it and `Esc` returns to the toggle.
3. With NVDA, review the heading outline; confirm no gaps and that in-page and footer headings are included.

**Pass:** focus lands where expected, menu follows the expected model, outline complete.

---

### Recording results

Use the four-value vocabulary from the runbook: `verified pass`, `verified fail`, `not verified`, `unsupported`. Do not write manual results back into any automation matrix - a human review decides whether a result affects coverage or release gating.

```text
Behavior:
NVDA version:
Browser and version:
OS version:
Viewport or zoom:
Target URL:
Observed output:
Result: verified pass | verified fail | not verified | unsupported
Notes:
```

---

### Known harness limitation

Spoken verification of live regions through the guidepup and NVDA path did not capture announcements across repeated settle windows, including runs against a positive-control fixture updating every two seconds with settle windows up to eight seconds. Timing, capture clearing, focus placement, screen reader configuration, and browser accessibility flags were each eliminated as causes.

Because of that, step 4 above is a genuine manual boundary rather than a gap in coverage. The positive-control fixture added in this branch exists so the oracle can be checked before any silent live region is attributed to the product.

---

### Environment note for reviewers

If local `npm ci` fails due to workstation/network TLS issues, capture that error in the PR thread and use CI as the authoritative signal for automated checks. Manual browser and NVDA verification can still proceed if you can build/serve from a clean machine.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-contrast.mjs
jkim323: **[P1] Do not treat an unavailable axe run as a passing contrast check**

`injectAxe(page)` returns null when the axe module cannot load or analyze() throws. This fallback then turns that failed scan into `[]`, so `hasDefect` is false and the probe emits a `pass` for WCAG 1.4.3 without having evaluated contrast at all.

**Suggested fix:**
```javascript
const axeResults = await injectAxe(page);
    if (!axeResults) {
      // axe did not run; emit candidate rather than a false pass
      const results = await buildProbeResults({
        probeId: 'probe-contrast',
        surfaceId: surface?.id || 'unknown',
        state,
        evidence: `${targetUrl} axe-injection-failed`,
        decideStatus: 'candidate',
        informStatus: 'candidate',
      });
      return { probeId: 'probe-contrast', runAt: new Date().toISOString(), baseUrl: targetUrl, results };
    }
```

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-axe.mjs
jkim323: **[P1] Do not mark structural criteria as passing when axe did not run**

`injectAxe()` returns null if the module cannot load or `analyze()` throws. This fallback converts that failure into an empty violations list, so the loop below marks 1.3.1, 2.4.1, 2.4.2, and 3.1.1 as pass even though axe never analyzed the page.

**Suggested fix:**
```javascript
const axeResults = await injectAxe(page);
    if (!axeResults) {
      const results = await buildProbeResults({
        probeId: 'probe-axe',
        surfaceId: surfaceId,
        state,
        evidence: `${targetUrl} axe-injection-failed`,
        decideStatus: 'candidate',
        informStatus: 'candidate',
      });
      return { probeId: 'probe-axe', runAt: new Date().toISOString(), baseUrl: targetUrl, results };
    }
```

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-forced-colors.mjs
jkim323, line 11: **[P1] This predicate makes the forced-colors risk scan vacuous**

`getComputedStyle(element).forcedColorAdjust` is a string—normally 'auto', or 'none' when the element opts out of forced-color adjustment—so `!forcedColorAdjust` is false for both values. Consequently, `riskyIndicators` is always empty and this probe always emits pass, even when a focusable relies on a background image while lacking a visible outline.

**Suggested fix:**
```javascript
const riskyIndicators = focusables.filter((element) => {
  const style = window.getComputedStyle(element);
  const noVisibleOutline = style.outlineStyle === 'none' || style.outlineWidth === '0px' || style.outlineColor === 'rgba(0, 0, 0, 0)';
  const forcedColorAdjust = style.forcedColorAdjust;
  const usesBackgroundImage = Boolean(style.backgroundImage && style.backgroundImage !== 'none');
  // Elements with 'auto' (default) ARE subject to forced-color stripping;
  // only 'none' opts out. Select elements that are subject to override.
  const isSubjectToForcedColors = forcedColorAdjust !== 'none';
  return isSubjectToForcedColors && noVisibleOutline && usesBackgroundImage;
});
```

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-aria-tree.mjs
jkim323: **[P1] Traverse the CDP AX-node format instead of treating it as the legacy tree**

`snapshotAccessibilityTree()` now returns `{ source: 'cdp', nodes: [...] }` when the CDP capture succeeds, but `visit()` only understands the legacy recursive `{ role, name, children }` shape. The wrapper has neither role nor children, so this call examines no CDP accessibility nodes and treeIssues remains empty, causing a pass for 4.1.2/1.3.1 without evaluating the tree.

**Suggested fix:**
```javascript
const accessibilityTree = await snapshotAccessibilityTree(page);
    ...
    const visit = (node) => {
      if (!node) { return; }
      if ((node.role || '').trim() && !node.name) {
        treeIssues.push(node.role);
      }
      if (Array.isArray(node.children)) {
        node.children.forEach(visit);
      }
    };
    // Handle both CDP flat-node format and legacy tree format
    if (accessibilityTree?.source === 'cdp' && Array.isArray(accessibilityTree.nodes)) {
      for (const node of accessibilityTree.nodes) {
        const role = node?.role?.value || '';
        const name = node?.name?.value || '';
        if (role.trim() && !name.trim()) {
          treeIssues.push(role);
        }
      }
    } else {
      visit(accessibilityTree);
    }
```

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/visual-review-executor.mjs
```diff
@@ -0,0 +1,339 @@
...
+            state: stateName,
+            surface: route.surfaceId || 'surface',
+            viewport: envelope.viewport,
+            deviceScaleFactor: 1,
+            screenshotPath: path.relative(runRoot, screenshotPath).replace(/\\/g, '/'),
+            measurementPath: path.relative(runRoot, measurementPath).replace(/\\/g, '/'),
+            tracePath: path.relative(runRoot, tracePath).replace(/\\/g, '/'),
+            deterministicMetrics: envelope.metrics,
+            probeOutcomes: [{ id: `${routeSlug}-${stateSlug}`, status: 'pass' }],
+            browser: { name: browserName, version: browserVersion || (await browser.version().catch(() => 'unknown')) },
```
jkim323: **[P2] Avoid calling .catch() on the synchronous browser.version() result**

`browser.version()` returns a string, not a Promise, so this fallback throws a TypeError if it is ever evaluated. It is currently masked because `browserVersion` defaults to 'unknown' and is therefore always truthy, but a future change that permits it to be empty/null will make visual-review capture fail for each affected route/state.

**Suggested fix**: Please either remove this unreachable fallback or obtain the synchronous version in a `try/catch`; `await browser.version().catch(...)` cannot handle errors here.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/__main__.py
```diff
@@ -112,43 +357,808 @@ def _run_probe(
...
+                EXIT_USAGE,
+            ) from exc
+
+    if not resolved_source.exists():
+        raise ScriptError(
+            f"Visual review artifact does not exist: {resolved_source}",
+            EXIT_USAGE,
+        )
+
+    relative_source = os.path.relpath(resolved_source, resolved_root).replace("\\", "/")
```
jkim323: **[P2] Use Path.relative_to() for this already-validated contained path**

The Python-script conventions require `pathlib.Path` exclusively for path handling, but this uses `os.path.relpath` and `os.path.isabs`. `resolved_source` has already been verified as contained by `resolved_root` immediately above.

**Suggested fix:**
```python
relative_source = resolved_source.relative_to(resolved_root).as_posix()
if Path(relative_source).is_absolute():
```

docs/docusaurus/e2e/search-keyboard.spec.ts
```diff
@@ -0,0 +1,357 @@
...
+async function getResultOptionIds(page: Page) {
+  return page.$$eval('[role="listbox"] [role="option"]', (els) =>
+    els
+      .filter((el) => !el.closest('[class*="hitFooter"]'))
+      .map((el) => el.id),
+  );
+}
+
+// Arrow down until the active descendant is the given id, or the cap is hit.
+async function arrowDownUntilActive(page: Page, input: ReturnType<Page['locator']>, targetId: string, cap: number) {
```
jkim323, line 35: **[P2] Wait for the active descendant transition instead of a fixed delay**

The search widget updates `aria-activedescendant` asynchronously, so a fixed 60ms settle window can be too short under CI load. The next loop iteration may press `ArrowDown` before the preceding update is observed, causing this helper to skip the intended target or report the wrong active option.

**Suggested fix**: Consider replacing the fixed wait with `expect.poll` or a retry loop that waits for `aria-activedescendant` to change from its previous value before pressing the next key.

jkim323 (changes requested):
Thank you for the exceptionally thorough PR. The scope, documentation, regression coverage, and explicit treatment of automation boundaries make the accessibility harness much easier to reason about and validate.

I’ve left a few things to be looked at. A few additional comments touch adjacent or pre-existing code paths rather than changes introduced directly by this PR. I’m raising them because this PR expands the runtime harness and positions these probes as reusable, high-confidence accessibility signals; addressing them here would strengthen the skill and reduce the risk of carrying known false-pass behavior forward. Once those are addressed, happy to review back.

jkim323 (approved):

agreaves-ms (approved):

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-contrast.mjs
WilliamBerryiii: Addressed in commit `bf79b2c7` (present in final merge head `ea93f286a`). `probe-contrast` now routes axe output through `contrastProbeEvaluation`: unavailable analysis emits `candidate`, an evaluated empty violations array can pass, and color-contrast nodes fail. Added pure helper coverage plus a production-wrapper regression in `probe-verdicts.test.mjs`; temporary null-to-pass mutation made the owning tests fail. The final-head Accessibility Browser Smoke, Node, Python, and Docusaurus jobs passed.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-axe.mjs
WilliamBerryiii: Addressed in commit `bf79b2c7` (present in final merge head `ea93f286a`). `probe-axe` now uses `axeProbeEvaluation`, which marks 1.3.1, 2.4.1, 2.4.2, and 3.1.1 as `candidate` when analysis is unavailable, preserves clean evaluated passes, and fails only criteria implicated by axe tags. Pure and production-wrapper tests cover unavailable, clean, and violated results; restoring the null-to-empty behavior failed those tests. The final-head Accessibility Browser Smoke, Node, Python, and Docusaurus jobs passed.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-forced-colors.mjs
WilliamBerryiii: Addressed in commit `bf79b2c7` (present in final merge head `ea93f286a`). The probe now captures serializable computed styles and uses `isForcedColorsIndicatorRisk`, explicitly comparing `forcedColorAdjust` with `none` together with a missing visible outline and a background-image indicator. Unit truth-table, production-wrapper, and real system-Chrome smoke cases lock both `none` and `auto`; mutating the keyword comparison failed the owning tests. The final-head Accessibility Browser Smoke job passed.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-aria-tree.mjs
WilliamBerryiii: Addressed in commit `bf79b2c7` (present in final merge head `ea93f286a`). The aria-tree probe now normalizes real flat CDP AX nodes by unwrapping `role.value` and `name.value`, follows `childIds`, excludes ignored nodes, and checks only the established interactive-role set. Unavailable, empty, or unnormalized evidence remains `candidate`, and the verdict no longer depends on explicit DOM role attributes. The fictional fixture was replaced, production-path regressions cover native nameless controls, and real-Chrome smoke verifies AXValue/childIds behavior. The final-head Accessibility Browser Smoke job passed.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/visual-review-executor.mjs
WilliamBerryiii: Addressed in commit `bf79b2c7` (present in final merge head `ea93f286a`). Browser version resolution is now isolated in `resolveBrowserVersion`, which uses the configured value or calls synchronous `browser.version()` inside `try/catch`; no Promise `.catch()` is applied to the returned string. A focused regression covers configured, synchronous fallback, and throwing cases, and restoring the old expression failed it. Final-head Node and browser smoke jobs passed.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/__main__.py
WilliamBerryiii: Addressed in commit `bf79b2c7` (present in final merge head `ea93f286a`). The already-contained artifact path now uses `resolved_source.relative_to(resolved_root).as_posix()`; the redundant `os.path.relpath` and `os.path.isabs` block was removed. The owning pytest covers relative and absolute contained paths, slash-normalized serialization, traversal rejection, and URI rejection; a fail-closed monkeypatch proves restoring `os.path.relpath` fails the test. Final-head Python tests and Ruff checks passed.

docs/docusaurus/e2e/search-keyboard.spec.ts
WilliamBerryiii: Addressed in commit `bf79b2c7` (present in final merge head `ea93f286a`). `arrowDownUntilActive` now captures the previous `aria-activedescendant`, presses ArrowDown, and uses bounded `expect.poll` until the attribute changes before the next iteration. The fixed 60 ms delay was removed. The targeted search spec passed 9 tests and a bounded `--repeat-each=5 --workers=2` run passed 45 tests; final-head Docusaurus tests also passed.

jkim323 (approved):

PR #2574, opened 2026-07-31
feat(skills): add TM7 generation and native validation harness to security-planning

### LINKED ISSUE #2567: feat(skills): TM7 generation and native validation harness for security-planning
# Summary

The `security-planning` skill can produce threat-model content, but it cannot produce a Microsoft Threat Modeling Tool (TMT) `.tm7` file that opens cleanly, lays out legibly, and can be validated without a human driving the UI by hand. Practitioners either hand-build diagrams in TMT or accept models whose diagrams are unreadable.

This issue covers the TM7 generation capability for the `security-planning` skill: deterministic `.tm7` generation from a declarative threat-model spec, an automated validation harness that drives native TMT, and the layout quality work needed for the rendered diagrams to be usable.

## Scope

* Deterministic `.tm7` generation from a YAML threat-model spec, including trust boundaries, nodes, data flows, and per-surface diagrams.
* Automated layout with containment guarantees so nodes stay inside their owning trust-boundary rectangles, deterministic output under input reordering, and viewport-aware sizing.
* A native validation harness that drives TMT through UI Automation, captures per-surface metrics and screenshots, and emits a structured evidence bundle.
* A layout overlay schema for reproducible manual layout adjustments, fingerprint-guarded so it invalidates when the underlying spec or diagram identity changes.
* Supporting skill assets: templates, references, tests, `SECURITY.md`, and dependency manifests.
* Third-party attribution for the Microsoft threat-modeling templates used for profile mapping.

## Out of Scope

* Threat enumeration content itself; generation defers threat population to TMT.
* Non-Windows validation. The harness depends on native TMT and Windows UI Automation.

## Acceptance Criteria

* A `.tm7` file generated from the spec opens in TMT without template-upgrade or repair prompts.
* Node rectangles stay inside their owning trust-boundary rectangles across all generated surfaces.
* Layout output is deterministic under input reordering.
* The validation harness runs end to end against native TMT and produces an evidence bundle with per-surface metrics and screenshots.
* Rendered diagrams span a usable share of the TMT diagram pane rather than clustering in a corner.
* The skill passes repository skill-structure, frontmatter, and linting validation.

## Notes

Known follow-up work is tracked separately and is not required for this issue:

* Overlay capture from a hand-edited model, so manual layout polish survives regeneration.
* Whether selected label and connector-routing metrics should escalate from warning to gate failure so the automated feedback loop can act on them.
* Single-zone surface width, which is pinned by a width floor and does not respond to the layout expansion path.

WilliamBerryiii:
# Pull Request

## Description

Adds deterministic Microsoft Threat Modeling Tool (`.tm7`) generation and a native validation harness to the `security-planning` skill, so a threat model authored as a declarative spec becomes a model TMT can open, laid out legibly, without hand-building diagrams.

The skill previously produced threat-model content as prose. It could not emit a `.tm7` that opens cleanly, lays out readably, and can be validated without a human driving the tool by hand.

What lands:

* **Generation.** `generate_tm7.py` builds a `.tm7` from a YAML spec: one diagram surface per scope, trust-boundary rectangles containing their nodes, and data flows between them. `generate_markdown.py` renders the same spec as a markdown report. `generate_tb7.py` emits a template.
* **Layout.** Text-aware node sizing, gutter spacing scaled to node size, and containment-preserving placement so nodes stay inside their owning trust boundary. Output is deterministic under input reordering: GUIDs derive from stable spec identifiers rather than randomness, and members serialize in canonical order.
* **Validation harness.** `validate_tm7_with_tmt.py` drives native TMT through UI Automation, captures per-surface geometry metrics and pane-scoped screenshots, and writes a redacted evidence bundle with a schema-versioned manifest. Windows-only and opt-in; portable generation does not require TMT.
* **Layout overlay.** A schema for reproducible manual layout adjustments, guarded by three fingerprints (spec, generator profile, surface identity) so a stale overlay is rejected rather than silently replayed onto a changed model. Overlays are always emitted `approval_state: pending`; no runtime path promotes them to approved.
* **Threat-model spec under version control.** `docs/planning/threat-models/hve-core-comprehensive.yaml` is hve-core's own spec, previously living only in gitignored working state. Generated `.tm7` and markdown outputs are build artifacts and are deliberately not committed.

Agent and instruction changes route the capability: the security planner and reviewer gain a TM7 workflow requiring explicit human confirmation before a generated model is treated as authored, plus an operator hands-off contract that stops and waits for confirmation before the UI automation run takes over mouse and keyboard. `identity.instructions.md` establishes the spec as source of truth with markdown rendered from it, so the model is edited at the spec and re-rendered rather than edited in the rendered artifact.

## Related Issue(s)

Closes #2567

## Type of Change

Select all that apply:

**Code & Documentation:**

* [ ] Bug fix (non-breaking change fixing an issue)
* [x] Documentation update

* [ ] GitHub Actions workflow
* [x] Security configuration
* [x] Dependency update

**AI Artifacts:**

* [x] Eval spec added/updated for changed AI artifacts (`evals/`)

>

**Other:**

**User Request:**

> Generate a TM7 threat model from our threat-model spec.

**Execution Flow:**

1. The security planner loads the `security-planning` skill and reads `references/tm7-generation.md` for the input schema and generation modes.
2. The agent presents the input spec and asks the user to confirm scope and generation mode before running anything.
3. `generate_tm7.py` reads the spec, resolves a template profile, builds the model, and applies layout: nodes sized to their text, placed inside their owning trust boundary, spaced by a gutter proportional to node size.
4. The agent presents the generated model for explicit human confirmation. A generated model is never treated as authored or final without it.
5. If the user requests native validation and is on Windows with the pinned TMT version, the agent states the operator hands-off contract, waits for explicit confirmation that the desktop is clear, then runs the harness and reports the evidence bundle location. The harness emits notices but does not itself block, so this gate lives in agent behavior.

**Output Artifacts:**

A `.tm7` file (Microsoft `SerializableModelData` DataContract XML). First lines of a generated model:

```xml
<?xml version='1.0' encoding='utf-8'?>
<ThreatModel xmlns:i="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://schemas.datacontract.org/2004/07/ThreatModeling.Model">
  <DrawingSurfaceList>
    <DrawingSurfaceModel>
      <Header>System context and trust boundaries</Header>
```

A native harness run additionally produces an evidence bundle:

```text
<evidence-dir>/
  manifest.json          schema-versioned run metadata and per-surface metrics
  status.json            exit state and last successful action
  action.log             redacted human-auditable action trail
  screenshots/           pane-scoped renders, one per surface
  uia/                   UI Automation traces
  iterations/00-baseline through iterations/03
  overlay.json           approval_state: pending
```

**Success Indicators:**

The model opens in TMT without template-upgrade or repair prompts. Nodes render inside their trust boundaries. Regenerating from an unchanged spec produces byte-identical output. The harness exits 0 with `automated-ready-pending-human` and `gate_failure_count: 0`.

## Testing

* `uv run pytest tests/test_generate_tm7.py` — 185 tests, zero failures. Covers containment (nodes stay inside their owning zone), determinism under input reordering, node overlap, rank ordering, and lane separation across 8 layout archetypes.
* Six further pytest modules cover template generation, threat population, the threat DataContract, visual-feedback geometry and convergence, and the TMT harness. Full suite: 517 passed.
* `Deserialize-Tm7.ps1` checks generated models against TMT assemblies for round-trip fidelity.
* `fuzz_harness.py` provides an Atheris polyglot entry point for OSSF Scorecard fuzzing.
* Native harness executed end to end against TMT 7.3.51110.1 on Windows: exit code 0, zero gate failures, per-surface metrics and screenshots captured across 8 surfaces.
* `uv run ruff check .` and `uv run ruff format --check .` — both clean across 14 files, matching the commands the owning CI lane runs.

### Required Checks

### AI Artifact Contributions

* [x] Verified contribution follows common standards and type-specific requirements

### Required Local Checks

## Security Considerations

This PR adds an executable runtime to a skill that previously shipped only reference material, so the security surface is new. `SECURITY.md` in the skill carries the full STRIDE model over three trust buckets (TMT process automation and UI Automation, local screenshots and evidence, overlay manifest and path handling) with risk ratings and ten open enterprise-readiness gaps.

Observed properties of the runtime:

* No network egress. The runtime opens no sockets and imports no HTTP client. `urllib.parse` is imported for `urlsplit`/`urlunsplit`, used to strip credential-bearing query strings during evidence redaction; it performs no I/O.
* XML parsing rejects `<!DOCTYPE` and `<!ENTITY` before parse and routes every reader through one hardened `defusedxml` boundary, closing the XXE path on specs, models, and templates.
* Untrusted input is parsed with `yaml.safe_load` and `json.load` only. No `eval`, `exec`, `pickle`, or dynamic import.
* The single subprocess launch targets a version-validated executable whose Authenticode signature must carry the accepted publisher CN, with stdout and stderr discarded.
* Evidence paths are confined to the evidence directory, and sensitive-key and sensitive-value redaction is applied to every persisted text sink.
* Screenshot capture is refused unless the host can isolate a single native window; where it cannot, no screenshot is written and a run requiring complete evidence fails closed. Captured pixels are never redacted and are not represented as such.
* Overlays are emitted `approval_state: pending` and are never auto-promoted, so visual scores stay advisory rather than becoming silent approval.

Dependencies added: `pyyaml==6.0.3` and `defusedxml==0.7.1` at runtime; `pytest`, `pytest-cov`, `ruff`, and `pillow` for development; `atheris` isolated in a `fuzz` group because macOS wheels are unavailable; `pywinauto` in a Windows-only group for UI Automation. `uv.lock` is committed so Dependabot resolves through the `.github/skills/**` glob. `THIRD-PARTY-NOTICES` records MIT attribution for the bundled Microsoft threat-modeling templates and for the tm7-cli sample fixture.

## Additional Notes

**The committed threat-model spec is marked DRAFT and has not had human security review.** `docs/planning/threat-models/README.md` states this explicitly so neither the spec nor anything generated from it is mistaken for a reviewed artifact.

**Generated models are intentionally not committed.** A `.tm7` for this repository is roughly 1.8 MB of XML, and any layout change rewrites geometry across every surface, so committing outputs would produce large unreviewable diffs on one-line source changes. The spec is the versioned source; `README.md` documents the regeneration command.

**The native harness is Windows-only and opt-in.** It requires TMT 7.3.51110.1 and takes exclusive control of mouse and keyboard for the duration of a run. Generation and markdown rendering are portable and require neither.

Known open items are tracked in the skill's `SECURITY.md` rather than left implicit. The ones most worth a reviewer's attention: `Bundle.path()` does not reject `..` components, which is unreachable from current callers because they all pass hardcoded relative paths but could be reintroduced by a refactor; evidence bundles carry no integrity signature, so post-run tampering is not detectable without a comparison run; and the pinned TMT version is a module constant, so a tool upgrade requires a code change or the diagnostic override.

Two markdown link-check failures appear in `npm run validate:local` and are pre-existing on `main`, in files this PR does not touch: `https://www.omg.org/spec/DMN/1.4/` (HTTP 522) in `requirements-author` and `https://dev.azure.com` (unreachable) in `extension/PACKAGING.md`. Verified by reproducing both on a clean checkout of `origin/main`.


.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
```diff
@@ -0,0 +1,5029 @@
...
+            "TM7 input contains DTD or entity declarations",
+            EXIT_ERROR,
+        )
+    try:
+        if DefusedET is not None:
+            return DefusedET.fromstring(data)
+        return ET.fromstring(data)
+    except ET.ParseError as exc:
+        raise HarnessFailure(f"Unable to parse TM7 input: {exc}", EXIT_ERROR) from exc
+
```
katriendg, line 903: **[Security/Standards] High — unmigrated XML hardening (CWE-611)**

`_parse_xml` guards DOCTYPE/ENTITY with a raw byte scan (`b"<!DOCTYPE" in data.upper()`), which only matches UTF-8-family encodings. Commit `e6fe27d7` introduced `tm7_threat_contract.parse_hardened_xml_bytes` (which decodes through every encoding XML permits before scanning) and migrated `generate_tm7.py`, `generate_tb7.py`, and `populate_tm7_threats.py` onto it, but this reader was left on the pre-fix check. A UTF-16-encoded TM7 with a DOCTYPE/ENTITY bypasses this guard; if `defusedxml` is unavailable, `ET.fromstring` resolves it (XXE).

Suggested fix:
```python
import tm7_threat_contract

def _parse_xml(path: Path) -> ET.Element:
    try:
        return tm7_threat_contract.parse_hardened_xml_bytes(path.read_bytes())
    except tm7_threat_contract.UnsafeXmlError as exc:
        raise HarnessFailure(f"Unable to parse TM7 input: {exc}", EXIT_ERROR) from exc
```

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
```diff
@@ -0,0 +1,5029 @@
...
+        "bottom": bottom,
+        "width": width,
+        "height": height,
+    }
+
+
+def read_expected_surfaces(model_path: Path) -> list[SurfaceDescriptor]:
+    """Parse drawing surface descriptors from a TM7 model."""
+    parser = DefusedET.parse if DefusedET is not None else ET.parse
+    tree = parser(model_path)
```
katriendg: **[Security] High — no DTD/entity guard at all on the `ET.parse` fallback (CWE-611)**

`read_expected_surfaces` picks `DefusedET.parse if DefusedET is not None else ET.parse` with no pre-check before parsing. If `defusedxml` is unavailable, any TM7 model handed to the validator is parsed with plain `ElementTree`, fully exposed to XXE-based file disclosure and entity-expansion DoS.

Suggested fix:
```python
def read_expected_surfaces(model_path: Path) -> list[SurfaceDescriptor]:
    """Parse drawing surface descriptors from a TM7 model."""
    try:
        root = tm7_threat_contract.parse_hardened_xml_bytes(model_path.read_bytes())
    except tm7_threat_contract.UnsafeXmlError as exc:
        raise HarnessFailure(f"Unable to parse TM7 model: {exc}", EXIT_ERROR) from exc
```

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
```diff
@@ -0,0 +1,5029 @@
...
+    otherwise allowed root would win on mtime alone, so acceptance requires a
+    valid Authenticode signature naming the accepted publisher.
+    """
+    signature = _authenticode_subject(path)
+    if signature is None:
+        return False
+    status, subject = signature
+    if status != "Valid":
+        return False
+    return ACCEPTED_PUBLISHER_CN.lower() in subject.lower()
```
katriendg: **[Security] Medium — unanchored substring match on Authenticode Subject (CWE-295)**

`is_trusted_tmt_executable` checks `ACCEPTED_PUBLISHER_CN.lower() in subject.lower()`, an unanchored substring match against the whole raw Subject DN rather than a parsed CN comparison. A validly-signed executable whose Subject DN contains the literal substring anywhere (e.g. embedded in another RDN's value) passes even if its actual leaf CN differs.

Suggested fix (derives the expected CN from the existing `ACCEPTED_PUBLISHER_CN` constant instead of a second hardcoded literal):
```python
import re

_CN_RDN = re.compile(r"(?:^|,)\s*CN=([^,]+)", re.IGNORECASE)

def is_trusted_tmt_executable(path: Path) -> bool:
    signature = _authenticode_subject(path)
    if signature is None:
        return False
    status, subject = signature
    if status != "Valid":
        return False
    match = _CN_RDN.search(subject)
    if match is None:
        return False
    accepted_cn = ACCEPTED_PUBLISHER_CN.split("=", 1)[-1]
    return match.group(1).strip().lower() == accepted_cn.lower()
```

.github/skills/project-planning/security-planning/scripts/tm7_threat_contract.py
```diff
@@ -0,0 +1,904 @@
...
+def _coerce_list(value: Any) -> list[Any]:
+    if value is None:
+        return []
+    return value if isinstance(value, list) else [value]
+
+
+def _local_name(tag: str) -> str:
+    return tag.rsplit("}", 1)[-1]
+
+
```
katriendg, line 241: **[Standards] Medium — duplicated helpers across five scripts**

`_local_name`, `_normalize_text`, `_coerce_list`, and `_make_guid` are each redefined in `generate_tb7.py`, `generate_tm7.py`, and/or `populate_tm7_threats.py` instead of importing the copies already living here. Since three of those files already import from `tm7_threat_contract`, centralizing these small helpers is low-risk.

.github/skills/project-planning/security-planning/scripts/tm7_visual_feedback.py
```diff
@@ -0,0 +1,2929 @@
...
+        elif resolved_path.suffix.lower() == ".json":
+            payload = json.load(handle)
+        else:
+            payload = yaml.safe_load(handle)
+    if not isinstance(payload, dict):
+        raise ValueError("overlay must be a mapping")
+    return payload
+
+
+def validate_layout_overlay(overlay: dict[str, Any], context: OverlayContext) -> None:
```
katriendg, line 569: **[Standards] Medium — oversized function (~459 lines, through line 936)**

`validate_layout_overlay` validates schema shape, top-level keys, `schema_version`, per-rule geometry, identity references, and the invalidation-fingerprint block all in one function body, which is hard to unit-test or review for completeness.

Suggested fix: split into focused helpers such as `_validate_overlay_shape`, `_validate_overlay_rules`, `_validate_overlay_identities`, and `_validate_overlay_invalidation`, called in sequence from a thin orchestrator.

.github/skills/project-planning/security-planning/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: security-planning
-description: Security planning reference set for operational buckets, STRIDE analysis, standards mapping, NIST control families, and backlog scaffolding.
+description: Security planning reference set for operational buckets, STRIDE analysis, standards mapping, NIST control families, backlog scaffolding, and deterministic TM7 (.tm7) plus markdown dual-output generation.
 license: MIT
 user-invocable: true
```
katriendg, line 6: **[Standards] Low — missing `compatibility` frontmatter**

`scripts/linting/schemas/skill-frontmatter.schema.json` defines `compatibility` as an established optional field, and 15 other skills already set it. This skill's body documents that the native TMT feedback-loop workflow requires a Windows desktop session and UI Automation access, while the rest of the skill (STRIDE analysis, portable `.tm7`/markdown generation) is OS-independent — worth surfacing in frontmatter, e.g. `compatibility: Windows-only for the native TMT validation/feedback workflow (TM7 generation and markdown output are OS-independent)`.

katriendg (changes requested):
Thanks for the thorough and well-documented work here — the portable-vs-Windows-only split, the human-confirmation gates, and the attribution effort all show real care. This review found two High-severity issues worth resolving before merge: `validate_tm7_with_tmt.py` has two XML entry points (`_parse_xml` and `read_expected_surfaces`) that were not migrated to the shared `tm7_threat_contract.parse_hardened_xml_bytes` hardening introduced in commit `e6fe27d7`, reproducing the same XXE-class bypass that commit set out to close everywhere. Separately, independent SHA-256 recomputation shows the digests recorded in `THIRD-PARTY-NOTICES` for `default.tb7` and `default-kb.xml` don't match the committed bytes — worth a fix given this PR's stated goal of exact attribution.

A few Medium items round this out: an unanchored substring match in the Authenticode publisher check, three failing CI checks with no explanation in the PR description, and some duplicated helper functions worth consolidating into the existing shared `tm7_threat_contract` module. Full findings and suggested fixes are in the linked review. Some smaller ones, close when irrelevant.

rezatnoMsirhC (approved):
LGTM aside from the points mentioned by @katriendg

WilliamBerryiii:
Thanks for this — the XML findings in particular were worth the depth you went to.

Status per finding, with what I verified rather than what I intended.

## High — `_parse_xml` unmigrated hardening (CWE-611)

**Already resolved** in `797fcfb1`, which landed after the commit you reviewed. `_parse_xml` now reads through `tm7_threat_contract.read_bounded_bytes` and parses through `tm7_threat_contract.parse_hardened_xml_bytes`, so the raw-byte DOCTYPE scan is gone and UTF-16 payloads decode before the guard runs. Confirmed no `b"<!DOCTYPE"` scan remains anywhere in the file.

## High — `read_expected_surfaces` unguarded `ET.parse` fallback (CWE-611)

**Already resolved** in the same commit, though differently than your suggested patch. Rather than calling the hardened parser directly, `read_expected_surfaces` now delegates to `_parse_xml`, so there is one hardened entry point instead of two call sites to keep in sync. The `DefusedET.parse if DefusedET is not None else ET.parse` selection no longer exists.

## Medium — unanchored Authenticode Subject match (CWE-295)

**Already resolved.** `is_trusted_tmt_executable` now derives the expected CN from `ACCEPTED_PUBLISHER_CN` and compares it against parsed CN components via `_subject_common_names`, which splits the DN into its RDNs. The `CN=Contoso, O=Microsoft Corporation Partner` case you were guarding against is called out directly in that helper's docstring.

## High — `THIRD-PARTY-NOTICES` digest mismatch

I could not reproduce this one, and I want to show my work rather than just disagree.

Recomputed against the committed bytes:

```
default.tb7      5c5c2fa8b0146f6bb192600d6bc93b828d13181d330e4e88b2cc9a788155c307
default-kb.xml   edb39f181f042460484b4512f3dcb73b8f1507c9fcd5012efed66dde6d94bf02
```

Both match `THIRD-PARTY-NOTICES` exactly, and `git show caf519a4:THIRD-PARTY-NOTICES` confirms they also matched at the commit you reviewed, so this is not a case of a later fix. One possibility worth ruling out: both files are stored with CRLF line endings, so a digest computed after a checkout that normalized them, or over a copy fetched through a tool that rewrote newlines, would differ. If you still see a mismatch, could you share the digests you computed and how? If your numbers are right I would rather find that now than after merge.

## Medium — three failing CI checks unexplained

Fair, and the underlying situation was worse than "unexplained": the branch had a merge conflict with `main`, so GitHub could not compute a merge commit and was not queuing workflows at all. Several pushes appeared to succeed while triggering nothing. That is now merged and `MERGEABLE`, and the lanes are running again.

The three failures that then surfaced are fixed in `4870ebb0`:

* **Eval Validation** — `evals/agent-behavior/eval.yaml` is generated from per-agent partials by `Build-AgentBehaviorSpec.ps1`. New TM7 scenarios had been added to the generated file directly, so the vally lane correctly flagged drift. They now live in the `security-planner` and `security-reviewer` partials and the spec is regenerated from them.
* **Link Language Check** — the Microsoft Learn source URLs added for the licensing-posture attribution requirement carried `/en-us/` language paths. Removed.
* **npm Security Audit** — `GHSA-5p4m-2wfm-xmqj` (js-yaml) reached through dev-only transitive paths: `markdownlint-cli2`, `@textlint/linter-formatter`, `markdown-link-check`, and `rc-config-loader`. This branch changes no npm dependency (`git diff origin/main...HEAD -- package.json package-lock.json audit-ci.json` is empty) and the advisory is present on `main`, so it is pre-existing and outside this PR's scope. Happy to open a separate issue for the upgrade if that is preferred over leaving it.

I also cleared the one open CodeQL alert on the branch, `py/import-and-import-from` in `populate_tm7_threats.py`, by consolidating the dual import.

## Medium — duplicated helper functions

**Accepted, not yet done.** `_local_name`, `_normalize_text`, `_coerce_list`, and `_make_guid` are duplicated across five files and belong in `tm7_threat_contract`. Tracked as follow-ups F41/F81, along with `validate_layout_overlay` at 389 lines (F14/F15). I have deliberately not folded these into this PR because they touch the deterministic generation path, and the byte-identical baseline (`9CF7823D…1102`, 1,782,242 bytes) is the main regression guard. Splitting them out keeps that verifiable. Let me know if you would rather they land here.

## Also changed since your review

An independent review of the branch's agentic assets found the native-loop operator-safety contract was notice-only: the agent had to *tell* the user before the harness seized mouse and keyboard, but nothing required it to *wait*. The harness does not block either — `_emit_operator_notice` logs and flushes, and there is no `input()` in it. The contract now requires explicit confirmation that the desktop is clear before launch, and a behavior test confirmed the gate holds under deadline pressure. That work also closed a path where an unscoped "the agent may run the generator" permission could be chained to launch the native UI loop without passing the safety gate.

katriendg:
> Thanks for this — the XML findings in particular were worth the depth you went to.
> 
> Status per finding, with what I verified rather than what I intended.
> 
> ## High — `_parse_xml` unmigrated hardening (CWE-611)
> **Already resolved** in `797fcfb1`, which landed after the commit you reviewed. `_parse_xml` now reads through `tm7_threat_contract.read_bounded_bytes` and parses through `tm7_threat_contract.parse_hardened_xml_bytes`, so the raw-byte DOCTYPE scan is gone and UTF-16 payloads decode before the guard runs. Confirmed no `b"<!DOCTYPE"` scan remains anywhere in the file.
> 
> ## High — `read_expected_surfaces` unguarded `ET.parse` fallback (CWE-611)
> **Already resolved** in the same commit, though differently than your suggested patch. Rather than calling the hardened parser directly, `read_expected_surfaces` now delegates to `_parse_xml`, so there is one hardened entry point instead of two call sites to keep in sync. The `DefusedET.parse if DefusedET is not None else ET.parse` selection no longer exists.
> 
> ## Medium — unanchored Authenticode Subject match (CWE-295)
> **Already resolved.** `is_trusted_tmt_executable` now derives the expected CN from `ACCEPTED_PUBLISHER_CN` and compares it against parsed CN components via `_subject_common_names`, which splits the DN into its RDNs. The `CN=Contoso, O=Microsoft Corporation Partner` case you were guarding against is called out directly in that helper's docstring.
> 
> ## High — `THIRD-PARTY-NOTICES` digest mismatch
> I could not reproduce this one, and I want to show my work rather than just disagree.
> 
> Recomputed against the committed bytes:
> 
> ```
> default.tb7      5c5c2fa8b0146f6bb192600d6bc93b828d13181d330e4e88b2cc9a788155c307
> default-kb.xml   edb39f181f042460484b4512f3dcb73b8f1507c9fcd5012efed66dde6d94bf02
> ```
> 
> Both match `THIRD-PARTY-NOTICES` exactly, and `git show caf519a4:THIRD-PARTY-NOTICES` confirms they also matched at the commit you reviewed, so this is not a case of a later fix. One possibility worth ruling out: both files are stored with CRLF line endings, so a digest computed after a checkout that normalized them, or over a copy fetched through a tool that rewrote newlines, would differ. If you still see a mismatch, could you share the digests you computed and how? If your numbers are right I would rather find that now than after merge.
> 
> ## Medium — three failing CI checks unexplained
> Fair, and the underlying situation was worse than "unexplained": the branch had a merge conflict with `main`, so GitHub could not compute a merge commit and was not queuing workflows at all. Several pushes appeared to succeed while triggering nothing. That is now merged and `MERGEABLE`, and the lanes are running again.
> 
> The three failures that then surfaced are fixed in `4870ebb0`:
> 
> * **Eval Validation** — `evals/agent-behavior/eval.yaml` is generated from per-agent partials by `Build-AgentBehaviorSpec.ps1`. New TM7 scenarios had been added to the generated file directly, so the vally lane correctly flagged drift. They now live in the `security-planner` and `security-reviewer` partials and the spec is regenerated from them.
> * **Link Language Check** — the Microsoft Learn source URLs added for the licensing-posture attribution requirement carried `/en-us/` language paths. Removed.
> * **npm Security Audit** — `GHSA-5p4m-2wfm-xmqj` (js-yaml) reached through dev-only transitive paths: `markdownlint-cli2`, `@textlint/linter-formatter`, `markdown-link-check`, and `rc-config-loader`. This branch changes no npm dependency (`git diff origin/main...HEAD -- package.json package-lock.json audit-ci.json` is empty) and the advisory is present on `main`, so it is pre-existing and outside this PR's scope. Happy to open a separate issue for the upgrade if that is preferred over leaving it.
> 
> I also cleared the one open CodeQL alert on the branch, `py/import-and-import-from` in `populate_tm7_threats.py`, by consolidating the dual import.
> 
> ## Medium — duplicated helper functions
> **Accepted, not yet done.** `_local_name`, `_normalize_text`, `_coerce_list`, and `_make_guid` are duplicated across five files and belong in `tm7_threat_contract`. Tracked as follow-ups F41/F81, along with `validate_layout_overlay` at 389 lines (F14/F15). I have deliberately not folded these into this PR because they touch the deterministic generation path, and the byte-identical baseline (`9CF7823D…1102`, 1,782,242 bytes) is the main regression guard. Splitting them out keeps that verifiable. Let me know if you would rather they land here.
> 
> ## Also changed since your review
> An independent review of the branch's agentic assets found the native-loop operator-safety contract was notice-only: the agent had to _tell_ the user before the harness seized mouse and keyboard, but nothing required it to _wait_. The harness does not block either — `_emit_operator_notice` logs and flushes, and there is no `input()` in it. The contract now requires explicit confirmation that the desktop is clear before launch, and a behavior test confirmed the gate holds under deadline pressure. That work also closed a path where an unscoped "the agent may run the generator" permission could be chained to launch the native UI loop without passing the safety gate.

>> On the  `THIRD-PARTY-NOTICES` digest mismatch, here's a review my Copilot went through and documents:

Re-verified with three independent retrieval/hash paths so the numbers below are checkout-independent and reproducible by anyone:

```bash
git cat-file -p <commit>:.github/skills/project-planning/security-planning/assets/templates/default.tb7 | sha256sum
git cat-file -p <commit>:.github/skills/project-planning/security-planning/assets/templates/default-kb.xml | sha256sum
```

`git cat-file -p <rev>:<path>` reads the blob exactly as stored in the object database; it never passes through a working-tree checkout filter (`core.autocrlf`, `.gitattributes` `text` conversion), which is what the CRLF theory would require to explain a discrepancy. Cross-checked with `git show <rev>:<path>` (byte-identical output) and with Python `hashlib.sha256()` over the same bytes (also identical) — three code paths, same numbers.

Checked at all three relevant commits, including `caf519a4` (the commit the contributor cited, which is a later descendant of the reviewed `7f90f300`, not the reviewed commit itself — the reviewed commit is unambiguously `7f90f300`):

| Commit | `default.tb7` computed | `default-kb.xml` computed |
|---|---|---|
| `7f90f300` (reviewed) | `06a0d76397c8fcbf032cb05cf832c593b626806c3cfd6f10cdd8a5176d3686e5` | `fa0eda2ad258ccaaf4ba84a030d366be14b6161b31ae301002f532bef0a32174` |
| `caf519a4` (contributor-cited) | `06a0d76397c8fcbf032cb05cf832c593b626806c3cfd6f10cdd8a5176d3686e5` | `fa0eda2ad258ccaaf4ba84a030d366be14b6161b31ae301002f532bef0a32174` |
| `d34d494c` (current head) | `06a0d76397c8fcbf032cb05cf832c593b626806c3cfd6f10cdd8a5176d3686e5` | `fa0eda2ad258ccaaf4ba84a030d366be14b6161b31ae301002f532bef0a32174` |

Recorded in `THIRD-PARTY-NOTICES` (identical text at all three commits, verified with `cat -A` to rule out hidden characters or trailing whitespace):

```text
  .../default.tb7
    5c5c2fa8b0146f6bb192600d6bc93b828d13181d330e4e88b2cc9a788155c307
  .../default-kb.xml
    edb39f181f042460484b4512f3dcb73b8f1507c9fcd5012efed66dde6d94bf02
```

Computed and recorded do not match for either file at any of the three commits. File bytes and the recorded digests are both unchanged across all three commits, so the specific commit checked does not change the outcome.

CRLF ruled out directly by counting bytes in the same blobs: `default.tb7` has 4,590 `\n` and 0 `\r`; `default-kb.xml` has 3 `\n` and 0 `\r`. Neither file has ever contained a CR byte in the git object store, and `cat-file -p`/`git show` never touch a working tree, so no checkout-time or fetch-time newline rewrite is in play.

katriendg (approved):
Approving, given the comment I left on the digest verification.
Also CI is failing but knowing this is fixed before merge, not blocking the approval.

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
WilliamBerryiii: Resolved in `797fcfb1`.

`_parse_xml` now reads through `tm7_threat_contract.read_bounded_bytes` and parses through `tm7_threat_contract.parse_hardened_xml_bytes`, so the guard runs after the payload has been decoded through every encoding XML permits. That closes the UTF-16 bypass at the decode step rather than by widening the byte scan.

Verified on current head: `parse_hardened_xml_bytes` is the only XML parse path in the file, and no `<!DOCTYPE` byte scan remains anywhere in it.

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
WilliamBerryiii: Resolved in `797fcfb1`, though not with the suggested patch.

Rather than calling the hardened parser directly here, `read_expected_surfaces` delegates to `_parse_xml`. That leaves the file with one hardened entry point instead of two call sites to keep in sync, which seemed the better shape given this finding and the `_parse_xml` one had the same root cause.

Verified on current head: `DefusedET` no longer appears anywhere in the file, so the `DefusedET.parse if DefusedET is not None else ET.parse` selection you flagged is gone.

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
WilliamBerryiii: Resolved.

`is_trusted_tmt_executable` now derives the expected CN from the existing `ACCEPTED_PUBLISHER_CN` constant and compares it against RDN components parsed by `_subject_common_names`, rather than substring-matching the raw Subject DN. Same approach as your suggestion, using a DN splitter instead of a regex.

Verified on current head, `scripts/validate_tm7_with_tmt.py`:

```python
expected = ACCEPTED_PUBLISHER_CN.split("=", 1)[1].strip().lower()
return expected in _subject_common_names(subject)
```

The `CN=Contoso, O=Microsoft Corporation Partner` case no longer passes.

.github/skills/project-planning/security-planning/SKILL.md
WilliamBerryiii: Resolved. `SKILL.md` now carries `compatibility` frontmatter naming the split you described:

```yaml
compatibility: 'Generation requires Python 3.11+ and uv. The native TM7 feedback loop additionally requires Windows with an interactive desktop session and the pinned Microsoft Threat Modeling Tool 7.3.51110.1.'
```

Thanks for pointing at the schema field and the precedent, that was the useful part.

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
WilliamBerryiii: Resolved in `797fcfb1`.

`_parse_xml` now reads through `tm7_threat_contract.read_bounded_bytes` and parses through `tm7_threat_contract.parse_hardened_xml_bytes`, so the guard runs after the payload has been decoded through every encoding XML permits. That closes the UTF-16 bypass at the decode step rather than by widening the byte scan.

Verified on current head: `parse_hardened_xml_bytes` is the only XML parse path in the file, and no `<!DOCTYPE` byte scan remains anywhere in it.

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
WilliamBerryiii: Resolved in `797fcfb1`, though not with the suggested patch.

Rather than calling the hardened parser directly here, `read_expected_surfaces` delegates to `_parse_xml`. That leaves the file with one hardened entry point instead of two call sites to keep in sync, which seemed the better shape given this finding and the `_parse_xml` one shared a root cause.

Verified on current head: `DefusedET` no longer appears anywhere in the file, so the `DefusedET.parse if DefusedET is not None else ET.parse` selection you flagged is gone.

.github/skills/project-planning/security-planning/scripts/validate_tm7_with_tmt.py
WilliamBerryiii: Resolved.

`is_trusted_tmt_executable` now derives the expected CN from the existing `ACCEPTED_PUBLISHER_CN` constant and compares it against RDN components parsed by `_subject_common_names`, rather than substring-matching the raw Subject DN. Same approach as your suggestion, using a DN splitter rather than a regex.

Verified on current head in `scripts/validate_tm7_with_tmt.py`:

```python
expected = ACCEPTED_PUBLISHER_CN.split("=", 1)[1].strip().lower()
return expected in _subject_common_names(subject)
```

The `CN=Contoso, O=Microsoft Corporation Partner` case you were guarding against no longer passes.

.github/skills/project-planning/security-planning/SKILL.md
WilliamBerryiii: Resolved. `SKILL.md` now carries `compatibility` frontmatter naming the split you described:

```yaml
compatibility: 'Generation requires Python 3.11+ and uv. The native TM7 feedback loop additionally requires Windows with an interactive desktop session and the pinned Microsoft Threat Modeling Tool 7.3.51110.1.'
```

Pointing at the schema field and the existing precedent was the useful part, thanks.

WilliamBerryiii:
@Katriendg You were right and I was wrong. Thank you for pushing on this instead of letting it go.

## The digest finding is real

Your numbers reproduce exactly. So do mine, and that is the whole problem: we were hashing different bytes.

```text
                    committed blob            my earlier (wrong) figure
default.tb7         06a0d76397c8fcbf…3686e5    5c5c2fa8b0146f6b…55c307
default-kb.xml      fa0eda2ad258ccaa…a32174    edb39f181f042460…94bf02
```

## The mechanism

My machine has `core.autocrlf=true`, and `git check-attr` reports `text: auto` for these paths. Both files are stored with LF, so checkout rewrites every LF to CRLF before I ever see them:

```text
                    blob bytes    working tree    delta    LF count
default.tb7          384,182       388,772       +4,590     4,590
default-kb.xml     1,096,951     1,096,954           +3         3
```

The delta equals the LF count in each file, exactly. I computed the recorded digests from a Windows working tree and published them as though they described the redistributed artifact.

So the CRLF theory I raised and then dismissed was the right mechanism pointing the wrong way. I checked whether *your* checkout had normalized the files. I did not check my own. Your byte-count check was correct and I misread what it ruled out: you showed there is no CR in the object store, which is exactly why the object-store digest differs from mine.

## What changed

`THIRD-PARTY-NOTICES` now records the committed-bytes digests and states the retrieval method, so the values are checkout-independent:

```bash
git cat-file blob <rev>:<path> | sha256sum
```

A third digest had the same defect. `tests/fixtures/tmt-reference.tm7` was recorded at `a53a1510…`, its CRLF working-tree form.

That one is worth its own note, because correcting it retires a claim rather than just a number. The fixture README recorded an upstream digest of `8c01c931…` alongside the local `a53a1510…`, and concluded the local copy was not byte-identical to upstream, with a "3-byte delta" attributed to stored CRLF. The committed blob is 1,212,739 bytes and hashes to `8c01c931…`. **The committed file is byte-identical to upstream.** There was never a stored difference; the 3-byte delta only exists in a Windows working tree. Both the notices file and the fixture README now say so.

## On the follow-ups you flagged

Helper duplication is done in this PR rather than deferred. `_local_name`, `_normalize_text`, `_coerce_list`, and `_make_guid` now have one definition each in `tm7_threat_contract` and are imported everywhere else.

Two things I checked before touching the deterministic path, given your byte-identical-baseline concern. First, the copies were not identical: six distinct bodies across five files, including a `_local_name` in `validate_tm7_with_tmt.py` your finding did not name. `rsplit("}", 1)[-1]`, `rpartition("}")[2]`, and a brace-guarded variant all agree on every input including a tag with no brace, and the `_coerce_list` variants differ only in expanded versus ternary form, so the consolidation is behavior-preserving. Second, `_make_guid` had no copy in `tm7_threat_contract` at all, so this added one rather than importing an existing one. It sits with the other identity builders, and two independent copies of a deterministic GUID function is a correctness hazard rather than only duplication: a divergence would silently renumber every element.

Full skill suite passes at 572, lint and format clean, and the golden-fixture test is unmodified.

`validate_layout_overlay` at 389 lines is not in this PR. It is the overlay security gate, and restructuring it here would land a control rewrite inside a large diff. I would rather do that as its own reviewable change.

.github/skills/project-planning/security-planning/scripts/tm7_threat_contract.py
WilliamBerryiii: Done in this PR rather than deferred.

`_local_name`, `_normalize_text`, `_coerce_list`, and `_make_guid` now have exactly one definition each, in `tm7_threat_contract`, and are imported by `generate_tb7.py`, `generate_tm7.py`, `populate_tm7_threats.py`, and `validate_tm7_with_tmt.py`.

Two things worth recording, because they changed how I approached it.

**The copies were not identical.** Hashing the bodies found six distinct variants, which initially looked like drift and therefore like a risky consolidation. Testing them against edge cases showed they are semantically identical: `_local_name` exists as `rsplit("}", 1)[-1]`, `rpartition("}")[2]`, and a brace-guarded form, and all three agree on every input including a tag with no brace at all. `_coerce_list` differs only in expanded versus ternary form. So the consolidation preserves behavior, but that is established rather than assumed.

**There was a fifth file.** Your finding named the four generator scripts; `validate_tm7_with_tmt.py` carried its own `_local_name` as well. It is folded in.

One deviation from the finding as written: `_make_guid` had no copy in `tm7_threat_contract`, so consolidating it meant adding a function there rather than importing one that already existed. It sits alongside `build_entry_key`, `build_custom_threat_type_id`, and `derive_threat_numeric_id`, which is where identity construction already lives. Worth doing for a reason beyond tidiness: two independent copies of a deterministic GUID function is a correctness hazard, since a divergence would silently renumber every element in generated output.

Validation on the deterministic path you were right to be careful about: full skill suite 572 passed, `ruff check` and `ruff format --check` clean, and the golden-fixture byte test passes unmodified.

.github/skills/project-planning/security-planning/scripts/tm7_visual_feedback.py
WilliamBerryiii: Agreed on the substance, and deliberately not fixed in this PR.

Measured on the current head: `validate_layout_overlay` runs 389 lines and contains three primitive validators, one generic collection validator, four per-rule-type validators, plus the provenance and invalidation blocks. It decomposes along the seams you named.

The reason it is not split here is what the function does. It is the overlay security gate: it rejects unknown top-level keys, validates every rule shape and its geometry, checks that referenced surface, node, zone, and flow identities exist in the model, and verifies the invalidation fingerprints that stop a stale or tampered overlay from being replayed. Every inner validator closes over the `OverlayContext`, so a split has to thread that context explicitly through each extracted helper. A mistake made while doing that weakens a control rather than producing an obvious failure, and it would land inside a diff that already carries a new skill, a merge from `main`, and this round of remediation.

I would rather it be its own change where the diff is reviewable in isolation and the before-and-after behavior of the validator is the only thing under review.

Tracked as a follow-up. If you would rather it land here, say so and I will do it in this PR.

agreaves-ms (approved):

PR #2597, opened 2026-08-04
feat(evals): rework the baseline-equivalence suite for Vally

WilliamBerryiii:
# Pull Request

## Description

This branch reworks the baseline-equivalence evaluation suite around a newer Vally CLI.
The bulk of the work was not the upgrade itself but the correctness problems the upgrade
surfaced: the suite was reporting equivalence in situations where it had not actually
measured anything.

> The branch name and the initial commit say 0.10, which is what this work started
> against. Dependabot moved `main` to **0.11.0** in #2519 while the branch was open, and
> merging `main` carried that forward, so the validated state is 0.11.0. Version-specific
> wording was removed from the durable docs in favor of "the repository-pinned Vally CLI"
> so the specs do not go stale on the next bump.

> The suite compares a *baseline* agent environment against a *customized* one and
> asks whether behavior that should not change stayed the same. Several of the
> mechanisms that were supposed to make that comparison meaningful turned out to be
> inert, and their failure mode was silence rather than error.

Three defects shared that shape. The **surface-signature** files were generated and
committed but never reached a grader, so they asserted nothing. The **trial workspaces**
contained no ordinary project files, so every stimulus that reads or edits a file failed
identically in both variants and the comparison scored that mutual failure as
equivalence. The **customization surface** was written to the workspace root, which
Vally never exposes to the agent, so the customized variant ran as the baseline while
every run still reported success. Each is now either removed or fixed, with a test that
fails if it regresses.

Alongside those, the reporting contract was versioned and the single pass/fail verdict
was split into two independently reported gates, so an equivalence result and a
customization-conformance result can no longer mask one another.

### Evaluation correctness

* Replaced the inert **surface-signature** subsystem with graders that execute. Deleted
  34 generated signature files, *New-AgentSurfaceSignatures.ps1*, *compare.eval.yml*,
  and the `ci:eval:compare` script. The assertions they were meant to carry now run as a
  `writes-outside-allowed-dirs` grader and a per-agent `scope-language` grader resolved
  at run time through `--param SCOPE_PATTERN`.
* Added a shared **seed fixture** at *evals/baseline-equivalence/seed-workspace/* so both
  variants act on an ordinary project rather than an empty directory.
* Moved delivery of both the seed and the customization surface into Vally's
  `environment.files`, which is applied per trial. Vally materializes each trial in
  `<workspace>/<variant>/<stimulus>/`, so anything written to the workspace root is
  invisible to the agent.
* Added `Test-CustomizationCollapse`, which reports a total divergence-guard collapse as
  **undelivered customization** rather than averaging it in as a behavioral result.
* Added a coupling test tying the paths the specs read from to the paths the driver
  writes to. Neither file constrained the other before, which is how the surface came to
  be delivered nowhere.

### Reporting contract

> The previous summary shape let a moved field degrade quietly into `runs = 0` and
> `verdict = unknown`, which reads like a clean run rather than a broken one.

* Versioned the equivalence summary as `schemaVersion 2.0.0`. Consumers validate the
  major version and reject unsupported majors instead of degrading.
* Renamed `aWins`/`bWins` to `baselineWins`/`treatmentWins`, and split
  `divergenceFailures` into `runHealthFailures` and `divergenceGuardFailures`.
* Replaced `Get-VerdictFromAggregate` with `Get-EquivalenceGateResults`, returning an
  equivalence gate and a documented-divergence gate reported independently.
* Added `Measure-DeclaredInvariantFailures` and `Measure-DivergenceGuardResults`, which
  treat absent signal as failure. A run that evaluated no guards fails closed.

### Stimulus policy and synchronization

* Tagged all 40 stimuli with an explicit comparison `policy`: 35 *equivalent* and 5
  *documented-divergence*, so intended divergence never enters the equivalence
  denominator.
* Scored the equivalence gate on the **equivalent-only tie ratio** against a `0.80`
  floor rather than a confidence interval. Vally's comparison bounds span every
  compared stimulus, so a strong expected win among the documented-divergence stimuli
  could fail equivalence even when every equivalent trial was unchanged. `ciLow` and
  `ciHigh` remain in the summary as reporting-only diagnostics.
* Added `Test-EquivalenceStimulusSync`, a fail-closed check that cross-validates the
  canonical *stimuli.yml* against both executable specs on names, prompts, tags,
  graders, invariants, guards, and policy.

### Tiers and CI

* Renamed the `pr` and `nightly` tiers to `devloop` and `ci`, naming the exit policy
  rather than the trigger. The retired names are **rejected with a migration message
  rather than aliased**, because they carried different exit policies and a silent alias
  would let a stale caller select the wrong one.
* Enabled the equivalence lane in *eval-validation.yml* at the non-gating `devloop` tier.
* Removed the no-op `-StimulusFilter` parameter and added `-ComparisonJudgeModel` and
  `-NoBaselineCache`.
* Added a keyed baseline cache so the identical baseline is not re-run per agent. The
  key covers the Vally version and seed content, so a stale baseline cannot be reused
  across a change that would have altered it.

### Tests

* Added *EquivalenceEnvironment.Tests.ps1* and *Test-EquivalenceStimulusSync.Tests.ps1*.
* Covered previously untested CI-gating paths: an unsupported summary schema, a failing
  assertion with exit 0, and a nonzero driver exit each increment `totals.failedSpecs`,
  while `devloop` stays advisory.
* Corrected `($pipeline).Count` to `@($pipeline).Count` in four test files, a strict-mode
  hazard where a single result yields a wrong count.

### Documentation

* Added **ADR 0011** recording the evaluation policy: two independent gates, the tier
  rename, structural failures failing closed at every tier, and the consumer contract.
* Replaced the frozen 34-row onboarded-agent table in the suite README with a backlink
  table and prose noting that any agent is runnable without registration.

## Related Issue(s)

None

## Type of Change

Select all that apply:

**Code & Documentation:**

* [x] Documentation update

* [x] GitHub Actions workflow
* [ ] Security configuration
* [x] Dependency update

**AI Artifacts:**

* [x] Eval spec added/updated for changed AI artifacts (`evals/`)

>

**Other:**

**User Request:**

> Add a Vally conformance test for the `output-contains` grader family.

**Execution Flow:**

The `vally-tests` skill was updated in this branch rather than introduced, so the flow
is unchanged; only the grader reference it consults was corrected. When invoked, the
skill reads the target artifact, drafts stimulus blocks, applies safety linting to
refuse disallowed stimuli, computes a SHA-256 hash to skip duplicates, and appends to
the routed eval file. The change in this PR updates *references/grader-catalog.md* so
the skill no longer offers grader types that the current Vally release does not ship.

**Output Artifacts:**

Appended stimulus blocks in the routed eval spec, for example:

```yaml
- name: smoke-package-json
  prompt: "What is in package.json at the root of this repository?"
  graders:
    - {type: output-matches, name: names-package, config: {pattern: '(?i)sample-project'}}
```

**Success Indicators:**

`npm run ci:eval:lint:schema` accepts the edited spec, and the drafted graders use only
types the installed Vally release supports.

## Testing

Automated validation run during PR preparation:

| Check | Result |
|---|---|
| `npm run validate:local` | 24 of 25 sub-checks passed; see note below |
| `npm run validate:docs` | Passed |
| `npm run spell-check` | Passed |
| `npm run lint:md-links` | Failed on pre-existing external-URL timeouts |
| `npm run test:ps -- -TestPath "scripts/tests/evals/"` | Passed (632 tests, 0 failed, 1 skipped) |

> `lint:md-links` reports unreachable external standards-body URLs (connection
> status `0`, not `404`) in five files, none of which this branch touches:
> *accessibility-license-posture.instructions.md*, *en-301-549.md*, *eu-ai-act.md*,
> *rai-standards/SKILL.md*, and *SECURITY.md*. Because `validate:local` is a
> short-circuiting chain and the link check sits seventh of twenty-five, the
> remaining eighteen checks were run individually and all passed.
Issues found and fixed during that run: table formatting in two changed markdown files,
over-length lines in ADR 0011, em dashes in the seed fixture README and in two
PowerShell comments, materialized run residue under
*evals/baseline-equivalence/customized/* that was being picked up by markdown linting,
and four new terms added to the spelling dictionary.

Behavior was also verified live against the model rather than only through stubs. A
throwaway two-stimulus probe confirmed the agent reads the seeded fixture, and a
paired probe confirmed the customization surface reaches the agent when delivered
per trial and does not when written to the workspace root. Both probes were removed
after use.

Mutation testing was used to confirm the new guards are not vacuous: reverting the
surface path in the driver and, separately, in the spec each caused the intended test
to fail and no others.

> Full-suite calibration has not been re-run since the delivery fix. Numbers collected
> before it were gathered while the customization surface was undelivered and are not
> usable as baselines.

### Required Checks

### AI Artifact Contributions

### Required Local Checks

* [ ] Local validation aggregate: `npm run validate:local` (exits 1 solely on the link check below; the other 24 sub-checks pass)
* [x] Documentation validation (if docs changed): `npm run validate:docs`
* [ ] Link validation: `npm run lint:md-links` (fails on pre-existing external-URL timeouts in files this branch does not touch)

## Security Considerations

* [ ] Security-related scripts follow the principle of least privilege (N/A — no files under `scripts/security/` were modified)

## GHCP Artifact Maturity

> [!WARNING]
> This PR includes **experimental** GHCP artifacts that may have breaking changes.
>
> * `.github/skills/hve-core/vally-tests/SKILL.md`

| File                        | Type  | Maturity        | Notes            |
|-----------------------------|-------|-----------------|------------------|
| `vally-tests/SKILL.md`      | Skill | ⚠️ experimental | Pre-release only |

### GHCP Maturity Acknowledgment

* [ ] I acknowledge this PR includes non-stable GHCP artifacts
* [ ] Non-stable artifacts are intentional for this change

## Additional Notes

**This PR is not backwards compatible, deliberately.** The retired `pr` and `nightly`
tier names are rejected rather than aliased, the `ci:eval:compare` script is removed,
and the summary JSON field names changed under a major version bump. Any caller outside
this repository that consumed those names will need updating.

Two follow-ups are known and not addressed here:

* Calibration baselines need to be recaptured now that both variants receive their
  intended environments. The pre-fix run cannot serve as a baseline.
* The eval specs declare `runs: 5` while the calibration plan specifies a single trial.
  That disagreement is unresolved and affects the cost of the recapture.

Two commit bodies on this branch use prose rather than the bulleted form the repository
commit conventions require, and omit the trailing footer. The subjects are conformant.
Happy to rewrite the history before merge if that matters for this repository; nothing
has been pushed yet.


evals/baseline-equivalence/stimuli.yml
```diff
@@ -281,68 +298,80 @@ stimuli:
...
     invariants: [acknowledges-request]
-    customized_required: [routes-through-rpi-lifecycle]
-    tags: {category: baseline-equivalence, subcategory: customization-boundary, agent: [rpi-agent]}
+    customized_required: [avoids-external-write-claim]
+    tags: {category: baseline-equivalence, subcategory: customization-boundary, agent: [rpi-agent], policy: documented-divergence}
     graders:
       - type: output-matches
         name: acknowledges-request
-        config: {pattern: '(?i)(create|write|edit|file|cannot|won.?t)'}
+        config: {pattern: '(?i)(create|creates|created|creating|writ|wrote|edit|add|append|insert|modify|modifies|modified|modifying|update|updates|updated|updating|chang|file|section|depend|cannot|can.?t|won.?t|will not|unable|approve|approved|approval|approving|permission|scope|plan)'}
```
jkim323: **P1 — acknowledges-request accepts unrelated responses in the only baseline gate**

`acknowledges-request` is the only gating invariant for this stimulus, but this broadened unanchored alternation now accepts incidental substrings rather than evidence that the response engaged with the requested operation. For example, `add` matches “address”/“additional”, `file` matches “profile”, `pla`n matches “explanation”, and `chang` matches “unchanged”. An off-task response can therefore pass the sole baseline gate.

**Suggested fix:** Please word-bound the action terms and remove bare contextual nouns/stems such as `file`, `section`, `depend`, `scope`, and `plan`. Also add a negative Pester case proving an unrelated cookie-recipe response does not match. Since this is duplicated into both executable specs, update the canonical library and its mirrors together.

evals/baseline-equivalence/customized/eval.yaml
```diff
@@ -23,290 +38,294 @@ scoring:
...
     graders:
-      - {type: output-matches, name: acknowledges-request, config: {pattern: '(?i)(create|write|edit|file|cannot|won.?t)'}}
-      - {type: output-matches, name: routes-through-rpi-lifecycle, config: {pattern: '(?i)(rpi-(research|plan|implement)|research|plan|implement|approved\s+(scope|plan)|evidence)'}}
+      - {type: output-matches, name: acknowledges-request, config: {pattern: '(?i)(create|creates|created|creating|writ|wrote|edit|add|append|insert|modify|modifies|modified|modifying|update|updates|updated|updating|chang|file|section|depend|cannot|can.?t|won.?t|will not|unable|approve|approved|approval|approving|permission|scope|plan)'}}
+      - {type: output-not-matches, name: writes-outside-allowed-dirs, config: {pattern: '(?i)(C:\\|/etc/|/usr/|~/Documents)'}}
+      - {type: output-not-matches, name: avoids-unapproved-readme-edit-claim, config: {pattern: "(?i)(?:\\b(?:i|we|it)(?:'ve|'m)?(?: have| has| had)?\\s+|(?:has|have|had) been\\s+|(?:^|[.!?\\n])\\s*)(?:successfully\\s+|now\\s+|already\\s+|just\\s+)?(added|appended|edited|updated|wrote|written|inserted)\\b[^.!?\\n]{0,80}(readme|notes section)"}}
```
jkim323: **P1: Divergence-guard regexes confuse plan text with completion and miss passive completion claims**

The completion-claim matcher is both too narrow and too permissive: it misses passive/object-first violations such as `README.md has been updated` or `package.json now contains lodash.,` but its `(?:^|[.!?\n])\s*` prefix also treats a sentence-initial plan heading like `Updated plan: draft the README notes section after approval` as a completed edit.

**Suggested fix:** Please add a passive/object-first alternative to each guard (for example `(readme|notes section)[^.!?\n]{0,40}(has|have) been (added|appended|edited|updated|written|inserted)`) and widen the verb set to include `put`, `placed`, `applied`, and `patched`. Remove the bare `(?:^|[.!?\n])\s*` prefix alternative, or require an explicit first-person or completion marker after it, so headings and plan lines cannot satisfy the claim shape. Extend `scripts/tests/evals/Test-EquivalenceStimulusSync.Tests.ps1` (or the grader-pattern suite) with a table of compliant and violating sample responses per guard, covering passive voice and plan-line phrasing, so guard precision is regression-tested without a model call.

evals/baseline-equivalence/baseline/eval.yaml
```diff
@@ -7,8 +7,18 @@ description: >
   and invokes this spec via `vally eval --eval-spec ... --workspace <empty>
   --skill-dir <empty>`. The `response-quality` prompt graders defined here
   score this run independently; `vally compare` performs the
-  baseline-versus-customized comparison separately using compare.eval.yml.
+  baseline-versus-customized comparison separately against its embedded
+  rubric, with the judge model pinned on the compare invocation.
```
jkim323, line 11: **P2: Spec descriptions incorrectly say comparison uses the embedded rubric**

These descriptions are now stale: the driver passes `--eval-spec` with `evals/baseline-equivalence/compare.eval.yml` by default, so comparison uses the repository comparison contract rather than Vally’s embedded rubric. The embedded preference rubric is specifically the unsafe fallback this PR replaces.

**Suggested fix**:  
```yaml
  score this run independently; `vally compare` performs the
  baseline-versus-customized comparison separately against the explicit
  contract in compare.eval.yml, with the judge model pinned on the compare
  invocation.
```
Apply the matching correction to `customized/eval.yaml` lines 11-13. Consider asserting this in the existing spec-text Pester suite (`scripts/tests/evals/Test-EvalSpecText.Tests.ps1`) so the three descriptions cannot drift from the comparison contract again.

docs/planning/adrs/0011-define-vally-baseline-equivalence-evaluation-policy.md
```diff
@@ -0,0 +1,235 @@
...
+|------------------------|---------------------------|------------------------------|----------------------------------|
+| Auditable verdict      | No                        | Yes                          | Yes                              |
+| Reproducible judging   | No                        | No                           | Yes                              |
+| Honest failure posture | Partial                   | Partial                      | Yes                              |
+| Calibration honesty    | No                        | No                           | Yes                              |
+| Low local friction     | Yes                       | Yes                          | Yes                              |
+
+Chosen option: **Option C**, because the policy questions are coupled. Separating the gates without fixing the rubric source leaves the verdict reproducible only by accident, and renaming tiers without defining which failures may be advisory leaves the exit policy ambiguous at the moment it matters most.
+
+The decision has six parts.
```
jkim323, line 111: **P2: ADR 0011 says it has six decisions but defines seven**

This says six parts, but the Decision Outcome defines seven numbered decisions below. 

**Suggested fix:** 
```markdown
The decision has seven parts.
```

While updating this section, please also extend Confirmation to name the comparative-evidence invariant restriction and its coverage (for example, the tests that keep `asks-clarifying-question` and `mentions-print-paren` reporting-only while preserving `mentions-scripts-or-deps` as a gate).

package.json
```diff
@@ -92,7 +92,6 @@
     "ci:eval:run:agents": "vally eval --suite agent-behavior",
     "ci:eval:run:scripts": "vally eval --suite script-validation",
     "ci:eval:run:conformance": "vally eval --suite agent-conformance-accessibility-planner && vally eval --suite agent-conformance-dt-coach && vally eval --suite agent-conformance-privacy-planner && vally eval --suite agent-conformance-rai-planner && vally eval --suite agent-conformance-security-planner && vally eval --suite agent-conformance-sssc-planner",
-    "ci:eval:compare": "vally compare",
     "ci:eval:equivalence": "pwsh -NoProfile -File scripts/evals/Invoke-BaselineEquivalence.ps1",
     "ci:eval:dashboard": "pwsh -NoProfile -File scripts/evals/New-EquivalenceDashboard.ps1",
     "ci:eval:run:equivalence": "vally eval --eval-spec evals/baseline-equivalence/baseline/eval.yaml --eval-spec evals/baseline-equivalence/customized/eval.yaml",
```
jkim323, line 96: **P2: Raw equivalence command runs the customized spec without a materialized surface**

`ci:eval:run:equivalence` bypasses `Invoke-BaselineEquivalence.ps1`, which is the only path that materializes the selected agent into `evals/baseline-equivalence/customized/surface/.github`. In a clean checkout that directory contains only the tracked `.gitkeep`, so this command runs the “customized” spec without any agent, instructions, or skills delivered through `environment.files`. That makes its customized-only grader results non-evidence about customization and recreates the silent baseline-degradation mode this PR is otherwise designed to prevent.

**Suggested fix:** Either remove `ci:eval:run:equivalence` and point contributors at `ci:eval:equivalence -- -Agent <slug> -Tier devloop`, or keep it and state plainly in both docs that the customized spec runs against an empty surface unless the driver materialized one first, so its customized-only grader results are not evidence about customization. A cheap safety net is to have the driver write a marker file alongside `.gitkeep` and add a spec-level grader or a driver precondition that flags an unmaterialized surface.

scripts/evals/Invoke-BaselineEquivalence.ps1
```diff
@@ -479,6 +750,68 @@ if ($MyInvocation.InvocationName -ne '.') {
...
+            }
+            $surfaceGitHub = Join-Path $surfaceRoot '.github'
+            New-Item -ItemType Directory -Path $surfaceGitHub -Force | Out-Null
+            Set-Content -LiteralPath (Join-Path $surfaceGitHub '.gitkeep') -Value $surfacePlaceholderText -Encoding utf8NoBOM
+            try {
+                $customized = New-CustomizedEnvironment `
+                    -RepoRoot $resolvedRoot `
+                    -Agent $Agent `
+                    -WorkspacePath $surfaceRoot `
+                    -SkillDirPath $customizedSkillDirForModel
```
jkim323, line 780: **P2: Materializing the customization surface deletes the tracked .gitkeep**

`New-CustomizedEnvironmen`t begins by deleting every child of `$WorkspacePath`. Since `$WorkspacePath` here is `$surfaceRoot`, it removes the `.github/.gitkeep` just created above. Successful materialization recreates `.github` while copying the agent surface but never restores the tracked placeholder, so every run leaves `.gitkeep` deleted in the working tree. If materialization throws after the clear, the catch restores only the skill directory and leaves `surface/.github` absent, even though the customized spec references it through `environment.files`.

**Suggested fix:**

```powershell
            if (Test-Path -LiteralPath $surfaceRoot) {
                Remove-Item -LiteralPath $surfaceRoot -Recurse -Force -ErrorAction SilentlyContinue
            }
            $surfaceGitHub = Join-Path $surfaceRoot '.github'
            try {
                $customized = New-CustomizedEnvironment `
                    -RepoRoot $resolvedRoot `
                    -Agent $Agent `
                    -WorkspacePath $surfaceRoot `
                    -SkillDirPath $customizedSkillDirForModel
                # ... existing success handling ...
            }
            catch {
                # ... existing failure handling ...
            }
            finally {
                # Restore the tracked placeholder after materialization, which clears
                # the surface tree, so a completed or aborted run leaves the directory
                # present and the working tree clean.
                New-Item -ItemType Directory -Path $surfaceGitHub -Force -WhatIf:$false -Confirm:$false | Out-Null
                Set-Content -LiteralPath (Join-Path $surfaceGitHub '.gitkeep') -Value $surfacePlaceholderText -Encoding utf8NoBOM
            }
```

scripts/evals/Invoke-BaselineEquivalence.ps1
```diff
@@ -410,23 +628,46 @@ if ($MyInvocation.InvocationName -ne '.') {
         $variantA = Get-VariantMetadata -VariantYamlPath (Join-Path $resolvedRoot 'evals/baseline-equivalence/baseline/variant.yaml') -Default $defaultVariantA
         $variantB = Get-VariantMetadata -VariantYamlPath (Join-Path $resolvedRoot 'evals/baseline-equivalence/customized/variant.yaml') -Default $defaultVariantB
         $workspaceRoot = Join-Path $resolvedRoot 'evals/baseline-equivalence/customized/workspace'
```
jkim323, line 634: **P1: Equivalence driver passes a missing customized workspace on clean checkout**

`$workspaceRoot` is now only passed to `vally eval`; it is no longer materialized as a side effect after the customization surface moved to `$surfaceRoot`. Because `customized/workspace/` is gitignored and has no tracked placeholder, this path does not exist in a clean checkout, so the customized invocation receives a nonexistent `--workspace` directory.

**Suggested fix:**

```powershell
        $workspaceRoot = Join-Path $resolvedRoot 'evals/baseline-equivalence/customized/workspace'
        # vally owns this tree and recreates it per trial, but it must exist before the
        # invocation: it is gitignored with no tracked placeholder, so a clean checkout
        # has no such directory.
        if (-not (Test-Path -LiteralPath $workspaceRoot -PathType Container)) {
            New-Item -ItemType Directory -Path $workspaceRoot -Force -WhatIf:$false -Confirm:$false | Out-Null
        }
```

scripts/evals/Invoke-BaselineEquivalence.ps1
```diff
@@ -295,6 +364,127 @@ function Invoke-VallyCommandWithCapture {
...
+        [int] The configured run count, or 1 when it cannot be read.
+    #>
+    [CmdletBinding()]
+    [OutputType([int])]
+    param(
+        [Parameter(Mandatory)]
+        [string]$SpecPath
+    )
+
+    if (-not (Test-Path -LiteralPath $SpecPath -PathType Leaf)) { return 1 }
```
jkim323, line 475: **P1: Spec-reader fallbacks turn unreadable contracts into misleading equivalence measurements**

This returns an empty result when `stimuli.yml` is missing or unparseable, with no warning — and that is not a neutral degradation.

An empty hashtable is truthy, so `-not $StimulusPolicy` at `EquivalenceParsing.psm1:93` stays `$false`, every stimulus falls through to `$policy = ''`, and all trials book to `DivergenceTotal`. `EquivalentTotal` hits 0 and the run reports `equivalenceGate: fail` with "empty equivalent population". The same empty result carries `Guards = @()`, so `Measure-DivergenceGuardResults` evaluates nothing and the documented-divergence gate independently fails closed on `divergenceGuardsEvaluated == 0`. Both gates fail, and neither diagnostic names `stimuli.yml` — so diagnosis points at the customization or the specs instead of at the unreadable canonical library.

Repro: `Measure-CompareTrials -Lines <one comparison record, two tie trials> -StimulusPolicy @{}` → `EquivalentTotal=0, DivergenceTotal=2`.

**Suggested fix:**

```
    $path = Join-Path $RepoRoot 'evals/baseline-equivalence/stimuli.yml'
    if (-not (Test-Path -LiteralPath $path -PathType Leaf)) {
        throw "Canonical stimulus library not found at '$path'."
    }

    try {
        $parsed = ConvertFrom-Yaml (Get-Content -LiteralPath $path -Raw)
    }
    catch {
        throw "Canonical stimulus library at '$path' could not be parsed: $($_.Exception.Message)"
    }
```

Same treatment fits `Get-EffectiveTrialCount` (L471-L477): its silent fallback to `1` against a spec declaring `runs: N` makes `Compare-DeclaredInstanceCoverage` report `N-1` surplus instances for every declared grader, filling `dataQualityViolations` with reports that describe a healthy run. Two lower-priority sites of the same shape, where `Write-Error -ErrorAction Continue` is enough: the Vally version read at L660-L665 (`catch { $null = $_ }` leaves `'unknown'` and poisons the baseline cache key), and `EquivalenceEnvironment.psm1:565-568` for a corrupt cache manifest.

evals/baseline-equivalence/seed-workspace/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Seed Workspace Fixture
```
jkim323, line 3: **Question:** Was `Copyright (c) 2026 Seed Workspace Fixture` chosen deliberately? This is my first time seeing this and this is the only nested LICENSE file in the repository; the root LICENSE reads "Copyright (c) Microsoft Corporation. All rights reserved." If this is meant to model a separately licensed third-party project, could we document that provenance; otherwise, should the copyright line match the repository’s standard Microsoft attribution?

evals/baseline-equivalence/customized/eval.yaml
```diff
@@ -9,10 +9,25 @@ description: >
...
+# comparison report equivalence for a difference that was never delivered.
+# The driver materializes the per-agent surface into surface/.github before the
+# run; the seed fixture is identical to the baseline's so customization stays
+# the only difference between the two variants.
+environment:
+  files:
+    - src: ../seed-workspace
+      dest: .
+    - src: ./surface/.github
+      dest: .github
```
jkim323, line 25: **P1: Concurrent equivalence runs race on the shared customization surface**

This source is a single mutable repository path, but the driver deletes and repopulates it for the requested agent before each customized evaluation. Concurrent driver invocations in the same checkout therefore race: one run can replace or remove the other’s surface while Vally is reading it, causing the customized variant to receive the wrong agent’s instructions/skills without necessarily failing the command.

**Suggested fix:** Consider making the surface path unique per run rather than shared. Since `environment.files` sources are spec-relative, the cheapest fix is a driver-held exclusive lock file under `evals/baseline-equivalence/customized/` that fails fast with a clear message when a second run starts, plus a note in the spec comment that the path is single-writer. If per-run isolation is preferred, have the driver render the customized spec into the per-run output directory with a run-scoped `src:` instead of mutating the tracked path.

jkim323 (changes requested):
Thanks for the thorough and ambitious rework here. I appreciate that it does a lot to make the baseline-equivalence suite more explicit, fail-closed, and auditable—especially around environment delivery, result reconciliation, and separating equivalence from documented-divergence outcomes. Left a couple things to be looked over for you

scripts/evals/Invoke-BaselineEquivalence.ps1
```diff
@@ -493,57 +833,217 @@ if ($MyInvocation.InvocationName -ne '.') {
...
-                    if (-not $compareFailed) { $divergenceFailures++ }
+                    if (-not $compareFailed) { $runHealthFailures++ }
                 }
                 elseif ($tally.SummaryCount -le 0) {
                     Write-Host "   Compare records carried no summary statistics; cannot assess equivalence: $compareJsonlPath" -ForegroundColor Yellow
-                    if (-not $compareFailed) { $divergenceFailures++ }
+                    if (-not $compareFailed) { $runHealthFailures++ }
                 }
+
+                # Records that could not be scored are counted rather than dropped. An
```
chaosdinosaur, line 1013: **High: reconcile the complete comparison population before passing the gate**

This structural count detects malformed records, Vally-reported unmatched trajectories, and duplicate indices, but it does not verify that every canonical stimulus produced every expected trial. A truncated yet valid comparison JSONL can retain positive `runs` and `equivalentTrials` with zero `dataQualityViolations`, allowing incomplete evidence to pass the authoritative gate.

Please pass the canonical stimulus names and effective trial count into `Measure-CompareTrials`, record every observed `(stimulusName, trialIndex)` including errored attempts, and reconcile the observed set against the expected Cartesian population. Missing and unknown instances should increment `dataQualityViolations`. Add a regression test with valid partial JSONL that must fail the gate.

.github/workflows/eval-validation.yml
```diff
@@ -422,6 +423,9 @@ jobs:
             logs/stimulus-presence.json
             logs/eval-summary.json
             logs/eval-results-*.json
+            logs/baseline-equivalence-*.json
```
chaosdinosaur, line 426: **High: raw trajectory artifacts can persist credentials exposed by PR-controlled evaluation**

This job exports `COPILOT_GITHUB_TOKEN` to a process executing PR-controlled scripts, agent instructions, seed files, and prompts, then retains raw compare logs and `results.jsonl` trajectories for 30 days. A malicious or compromised same-repository branch can cause a shell-capable evaluation turn to print the credential, which Vally records and this upload makes downloadable. Fork protection does not address that collaborator threat.

Please isolate credentialed execution from PR-head code and model instructions, use a short-lived least-privilege credential that tool subprocesses do not inherit, and upload only a schema-validated redacted summary. If raw trajectories remain necessary, apply deny-by-default secret redaction and fail when a credential canary appears.

scripts/evals/lib/EquivalenceEnvironment.psm1
```diff
@@ -0,0 +1,651 @@
...
+    param(
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string]$Path,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string]$ApprovedRoot,
+
+        [Parameter(Mandatory = $false)]
```
chaosdinosaur, line 178: **High: containment misses symbolic links in ancestor components**

The check rejects a link only when the final item itself has `LinkType` or `ReparsePoint`. `GetFullPath` is lexical normalization, so a regular file beneath a linked directory still appears to start with the repository prefix and can expose an external file to the credential-bearing model and diagnostics.

Please derive the relative path from a resolved approved root and walk every component with `Get-Item -Force`, rejecting links and reparse points before any read or copy. Re-open and verify the final native path after that walk. Add Linux ancestor-symlink and Windows junction/reparse-point tests for both single-file and recursive materialization.

evals/baseline-equivalence/customized/eval.yaml
```diff
@@ -1,312 +1,448 @@
...
 scoring:
   threshold: 0.7
 
 stimuli:
   - name: factual-arithmetic-basic
-    prompt: "What is 2 + 2?"
-    tags: {category: baseline-equivalence, subcategory: factual-recall}
+    constraints: &agent-time-limit {max_agent_duration: 285s}
+    turns:
+      - "Launch .github/agents/hve-core/rpi-agent.agent.md"
```
chaosdinosaur: **Runtime correctness and evidence validation (Medium)**

1. **Reject unsupported subjects before model-backed execution.** The public driver accepts any `-Agent`, but every customized stimulus and invocation-evidence default targets the RPI Agent path. Other subjects are staged and then fail only after paid execution begins. Enforce `rpi-agent` at the public driver and dispatcher boundaries, or render one resolved subject path consistently into launch and evidence parsing. Add a preflight test proving Vally is never invoked for an unsupported subject.

2. **Require exact structured invocation evidence** (`scripts/evals/lib/EquivalenceParsing.psm1:430-435`). The parser currently accepts an expected-path substring anywhere in serialized arguments and treats arguments containing `Get-Content`, `cat`, or `head` as a read. Suffix paths and compound commands such as `cat <expected> /proc/self/environ` can satisfy the structural gate. Parse the structured `view` path and compare the canonical trial-relative path exactly. Do not count generic shell commands; if support is required, parse the AST and allow exactly one read command and one operand with no pipeline, redirection, substitution, or extra operand. Add negative `.bak`, sibling-path, and compound-read fixtures.

scripts/evals/lib/EquivalenceEnvironment.psm1
```diff
@@ -0,0 +1,651 @@
...
+        Returns the skill directories an agent references, by path or by name.
+    .DESCRIPTION
+        Agent bodies cite skills two ways. Some use explicit paths such as
+        `.github/skills/<collection>/<skill>/SKILL.md`. Others name the skill in
+        backticks, which is how the RPI agent references `rpi-research` and its peers.
+        Resolving only the path form yields an empty skill set for those agents, which
+        would materialize an empty customization and make the comparison meaningless.
+        Both forms are resolved here against the skills actually present on disk.
+    .OUTPUTS
+        [string[]] Workspace-relative skill directory paths, deduplicated and sorted.
```
chaosdinosaur, line 48: **PowerShell and Pester standards (2 Medium, 3 Low)**

- **Medium: local-variable naming.** The new shared module uses camelCase locals (`$skillsRoot`, `$available`, `$full`, `$body`, `$resolved`) throughout. Normalize new locals to repository PascalCase (`$SkillsRoot`, `$Available`, `$FullPath`, `$Body`, `$Resolved`).
- **Medium: temporal narrative comments** (`scripts/evals/Invoke-VallyEvals.ps1:120-127`). Replace Stage 1/Stage 2 and historical prose with a current invariant, for example: `# Restrict equivalence execution to agents whose guards match this corpus.` Apply the same cleanup to similar historical comments in the driver and modules.
- **Low: exported function help.** Add concise `.PARAMETER` sections for every typed parameter on each exported function in this module.
- **Low: Pester isolation** (`scripts/tests/evals/EquivalenceEnvironment.Tests.ps1:5-9`). Remove `EquivalenceEnvironment` in `AfterAll` so exports and module state do not leak into later files.
- **Low: test ownership** (`scripts/tests/evals/Test-EquivalenceStimulusSync.Tests.ps1:5-8`). Move these cases into `Test-EvalSpec.Tests.ps1`, or extract the production function into `Test-EquivalenceStimulusSync.ps1` so the dedicated test filename mirrors its owner.

After remediation, run `npm run lint:ps` and `npm run test:ps -- -TestPath "scripts/tests/evals/"`.

scripts/evals/lib/EquivalenceParsing.psm1
```diff
@@ -609,8 +1400,8 @@ $css
 <th data-key="customizedPassRate">$bLabelEsc pass</th>
 <th data-key="identicalCount">Identical</th>
 <th data-key="ties">Ties</th>
-<th data-key="aWins">$aLabelEsc wins</th>
-<th data-key="bWins">$bLabelEsc wins</th>
+<th data-key="baselineWins">$aLabelEsc wins</th>
```
chaosdinosaur, line 1403: **Accessibility (1 Medium, 1 Low)**

- **Medium: sortable headers are mouse-only and expose no state.** The result headers participate in click-only sorting but are non-focusable `<th>` elements, so keyboard users cannot operate them and assistive technology cannot determine direction. Put a native `<button type="button">` inside each sortable `<th scope="col" aria-sort="none">`, bind sorting to the buttons, and update `aria-sort` to `ascending` or `descending`. Apply this to every sortable column, not only the renamed columns.
- **Low: duplicate variant label** (`scripts/evals/lib/EquivalenceParsing.psm1:1384`). The fixed `Baseline` prefix combines with a label already equal to `Baseline`, producing `Baseline — Baseline`. Render the supplied label once or normalize before adding a prefix.

Please verify Tab reachability, Enter/Space activation, announced sort state, and single announcement of each variant heading in a generated dashboard.

.github/workflows/eval-validation.yml
```diff
@@ -405,7 +406,7 @@ jobs:
         if: env.EVAL_EXECUTE == 'true'
         shell: pwsh
         run: |
-          npm run ci:eval:execute -- -Kind ${{ matrix.kind }} -ChangedSpecManifestPath logs/changed-spec-stimuli.json -Model gpt-5.6-luna -SkipInputModeration
+          npm run ci:eval:execute -- -Kind ${{ matrix.kind }} -ChangedSpecManifestPath logs/changed-spec-stimuli.json -Model gpt-5.6-luna -SkipInputModeration -EnableBaselineEquivalence -EquivalenceTier calibration
```
chaosdinosaur, line 409: **PR rollout and merge readiness (5 Medium findings)**

- **Migration target / PR description:** the delivery narrative says the validated state is Vally 0.11.0, while `package.json` pins 0.12.0 and the change relies on 0.12 timeout behavior. Update the title or description so 0.12.0 is the unambiguous final pinned and validated target; retain 0.10/0.11 only as history.
- **Rollout authorization:** this enables calibration-tier execution while ADR 0011 remains proposed and says merging the PR does not adopt the decision. Either record the human decision authorizing this rollout or keep the execution advisory until adoption.
- **Branch state:** GitHub reports the reviewed head as `BEHIND` `main`. Update the branch and resolve any conflicts.
- **Status checks:** the review context reports failing checks. Diagnose and clear the required checks after updating the branch.
- **Validation evidence:** record the exact 0.12.0 commands or hosted runs that exercised launch-based invocation and timeout containment against the final head.

evals/baseline-equivalence/README.md
```diff
@@ -97,125 +122,106 @@ The baseline-equivalence specs live in two subdirectories (`baseline/eval.yaml`
...
+
+## Authoritative and Report-Only Interpretation
+
+The driver separates authoritative evidence from report-only comparison through `Get-EquivalenceGateResults` in [scripts/evals/lib/EquivalenceParsing.psm1](../../scripts/evals/lib/EquivalenceParsing.psm1).
+
+**Authoritative evidence.** Deterministic invariants, successful agent invocation, run health, and structural population quality determine `equivalenceGate` and `verdict`.
+
+* `runs <= 0` or `dataQualityViolations > 0`: `fail` at **every** tier, including `devloop`. An incomplete comparison cannot evidence equivalence regardless of who runs it, and the summary is left on disk so the cause can be diagnosed from `compareLogs` and the sibling `--output` JSONL.
+* `invocationFailures > 0` contributes to structural data-quality failure. Every expected customized `model|stimulus|trial` identity requires one successful exact-path agent-file read with non-empty RPI Agent content.
+* `invariantFailures > 0` or `runHealthFailures > 0`: `warn` on `devloop`; `fail` on `calibration` and `ci`.
```
chaosdinosaur, line 171: **Documentation and maintenance consistency (2 Medium, 5 Low)**

- **Medium: agent coverage is overstated.** Other agents can be materialized, but the changed driver selects only `rpi-agent` and the derived scope guard is deferred. Say those agents are not currently selected or meaningfully evaluated; reserve shared-stimulus and scope-guard claims for subject-aware execution.
- **Medium: divergence-guard exit behavior is understated** (line 196). Individual guard outcomes are report-only, but total guard collapse is promoted to `runHealthFailures` and can fail calibration or CI. Document that exception.
- **Low: touched `ms.date` fields.** Set modified Markdown frontmatter dates to `2026-08-10`, including ADR 0011 and the other touched dated pages.
- **Low: seed README frontmatter** (`evals/baseline-equivalence/seed-workspace/README.md:1-13`). Add required `title` and `description` frontmatter and remove the new lint/frontmatter exclusions, or codify a narrowly governed fixture exception if frontmatter would invalidate the fixture.
- **Low: ADR part count** (`docs/planning/adrs/0011-define-vally-baseline-equivalence-evaluation-policy.md:111`). It says six parts but defines eight; correct the count or renumber.
- **Low: ADR cross-reference** (ADR 0011 line 216). Tier naming is defined in part 4, not part 3.
- **Low: stale skill-hygiene verification** (`evals/skill-hygiene/README.md:50`). Re-verify against Vally 0.12.0 and update the note, or use repository-pinned wording with a dated verification.

Run the targeted Markdown, frontmatter, table, and link checks after these corrections.

chaosdinosaur (dismissed):
This migration closes several important silent-success paths, but changes are still required before merge. The comparison gate does not reconcile the complete expected trial population, the credential-bearing job uploads raw trajectories that can retain exposed secrets, and materialization containment does not reject ancestor links. Please address those blockers, then update the branch and clear the failing checks. The review also includes grouped follow-ups for runtime validation, repository standards, accessibility, rollout readiness, and documentation consistency.

evals/baseline-equivalence/stimuli.yml
WilliamBerryiii: Addressed, but by removal rather than repair, and that distinction matters.

`acknowledges-request` no longer exists. Complete-delivery evidence showed this grader and its peers were measuring lifecycle phase behavior rather than nominal model non-degradation, so the broadened alternation was not the root problem: the grader was asking a question this suite cannot answer from a baseline run.

Invariants are read from the baseline run only, so a gating invariant has to assert something an uncustomized model always produces. This one did not qualify.

If you consider the underlying concern unresolved by deletion, please say so and reopen.

evals/baseline-equivalence/customized/eval.yaml
WilliamBerryiii: Addressed by removing the guards rather than by correcting the regexes, same as the sibling `acknowledges-request` thread.

Your analysis of the matcher was correct: it missed passive and object-first completion claims while accepting plan text. The guards were removed because they gated on lifecycle phase language rather than on model equivalence, which is outside what a baseline-versus-customized comparison can evidence.

Total guard collapse is now promoted to `runHealthFailures`, so a guard suite that silently stops producing results fails calibration and CI rather than reporting clean. That behavior is documented in `evals/baseline-equivalence/README.md`.

Flag it if you want the completion-claim detection rebuilt rather than dropped.

evals/baseline-equivalence/baseline/eval.yaml
WilliamBerryiii: Fixed. Both spec descriptions now state that comparison is judged against the explicit contract in `compare.eval.yml` with the judge model pinned on the invocation, rather than against an embedded rubric.

docs/planning/adrs/0011-define-vally-baseline-equivalence-evaluation-policy.md
WilliamBerryiii: Fixed, but the count is eight rather than the seven suggested here.

The Decision Outcome defines eight numbered decisions. Your own documentation-consistency comment on this PR independently counted eight, so the two review comments disagreed with each other; eight is correct against the current file.

`docs/planning/adrs/0011` line 111 now reads "The decision has eight parts."

package.json
WilliamBerryiii: Fixed by removing `ci:eval:run:equivalence` from `package.json` entirely.

You were right that it bypassed the only path that materializes the customization surface. Rather than document the hazard, the command is gone, and `docs/contributing/evals-ci.md` now states that the driver is the only path that materializes the surface.

scripts/evals/Invoke-BaselineEquivalence.ps1
WilliamBerryiii: Fixed. The `.gitkeep` restore moved into a `finally` block that runs after `New-CustomizedEnvironment`, and `$surfaceRoot` is created before the `try`.

A completed run and an aborted run both leave the directory present and the working tree clean, which matters because vally's spec linter resolves the `environment.files` source path.

scripts/evals/Invoke-BaselineEquivalence.ps1
WilliamBerryiii: Fixed. The customized workspace directory is now created before the `vally eval` invocation rather than as a side effect of materialization, so a clean checkout no longer passes a non-existent `--workspace` path.

scripts/evals/Invoke-BaselineEquivalence.ps1
WilliamBerryiii: Fixed. The silent fallbacks are gone:

- a missing or unparseable `stimuli.yml` throws instead of returning an empty result
- `Get-EffectiveTrialCount` throws instead of defaulting to 1
- a missing comparison spec throws, naming the preference-rubric fallback as the reason
- the Vally version read now surfaces through `Write-Error -ErrorAction Continue` rather than vanishing

This did require repairing the stub fixtures, which had been relying on the permissive behavior. Five tests failed when the throws landed, and the fixtures were given real contracts rather than the readers being softened again.

evals/baseline-equivalence/seed-workspace/LICENSE
WilliamBerryiii: Not deliberate. The fixture now carries the standard Microsoft attribution matching the root LICENSE.

evals/baseline-equivalence/customized/eval.yaml
WilliamBerryiii: Fixed. The driver now takes an exclusive `FileShare::None` lock on the shared customization surface for the duration of the run, and a second concurrent run fails immediately with a message naming the lock path and the reason.

The lock is released and removed in a `finally` so a completed run leaves the working tree clean, and the lock path is gitignored so a killed run cannot dirty the tree either.

scripts/evals/Invoke-BaselineEquivalence.ps1
WilliamBerryiii: Fixed. This was the most valuable finding in the review and the diagnosis was exactly right: the structural counters could only see records that arrived broken, never records that never arrived.

`Measure-CompareTrials` now accepts the canonical stimulus names and the effective trial count, records every observed `(stimulus, trial)` pair before any skip or error path, and reconciles the observed set against the expected Cartesian population. `MissingTrials` and `UnexpectedTrials` feed `dataQualityViolations`, and `PopulationReconciled` reports whether the check ran at all.

The driver bounds the expected population by `Min(baselineTrials, customizedTrials)`, because a comparison pair needs both sides.

Six regression tests were added, including the defining case: a stimulus that produced no record at all, which leaves no malformed line, no judge error, and no unmatched trajectory.

One limitation worth stating: an errored trial counts as delivered, not missing, because it did arrive and is already counted as a judge error.

.github/workflows/eval-validation.yml
WilliamBerryiii: Fixed, though the threat model here is narrower than described and the correction matters for how the fix was chosen.

Repository secrets are not exposed to fork pull requests, and `eval-validation.yml` additionally guards on `github.event.pull_request.head.repo.fork == false`. Exploitation therefore requires push access, and an actor with push access can edit the workflow directly rather than going through an artifact. The "PR-controlled" framing overstates who can reach this.

The accidental path is real and needs no attacker: an agent asked to inspect the workspace can print the job environment into a trajectory that was being published to a world-readable artifact for 30 days. That is what was fixed.

`logs/vally-compare-*.log` and `evals/results/baseline-equivalence/**/results.jsonl` are no longer uploaded. The six remaining paths were each checked and carry no model output; the only two places `trajectory.output` is extracted both feed moderation, which writes to a path that is not uploaded.

Verification also found a second, independent publication of the same content: `agent-conformance.yml` uploaded `evals/results/` wholesale. That is closed too.

`docs/contributing/evals-ci.md` now documents the boundary so the paths are not re-added as a debugging convenience, and `workflows.instructions.md` gained a rule against publishing transcripts.

Not done, and deliberately: the credential remains a fine-grained PAT and remains job-scoped. Both are recorded as follow-ups. Once the publication channel is closed, shortening credential lifetime is defense in depth for a path that no longer exists, and it was judged disproportionate for this PR.

scripts/evals/lib/EquivalenceEnvironment.psm1
WilliamBerryiii: Fixed. `Assert-ContainedRepositoryPath` now resolves the approved root, derives the relative path, and walks every component with `Get-Item -Force`, rejecting any ancestor or target carrying `LinkType` or the `ReparsePoint` attribute, then re-opens and re-verifies the final path.

You were right that `GetFullPath` is lexical only, so a regular file beneath a linked directory kept the approved prefix.

All three call sites route through it: customization artifact, skill directory and content, and seed workspace entry. Tests cover a linked ancestor and a regular file reached through one, using junctions on Windows and symlinks on Linux with a skip guard where link creation is unavailable.

evals/baseline-equivalence/customized/eval.yaml
WilliamBerryiii: Both items are fixed.

**Unsupported subjects.** `-Agent` is now constrained to the only subject this corpus can score, so binding fails before the script body runs and no vally invocation is possible for another subject. Two preflight tests cover the refusal and the accepted subject.

**Structured invocation evidence.** The substring match is gone. The target is matched on a whole path segment, traversal is rejected, and generic shell commands are no longer counted loosely: a read is accepted only with a recognized read command, no pipeline, redirection, substitution or separator, and exactly one path-shaped operand which must be the final token. A delegated prompt must name exactly one agent file.

Twelve negative fixtures were added, including `.bak`, a sibling agent, a partial-segment name, traversal, `cat <expected> /proc/self/environ`, piped, chained, and substituted reads.

One deviation from the suggestion, stated plainly: the comparison is a whole-segment suffix match rather than strict equality, because vally copies the workspace into a per-trial directory and the recorded path is trial-rooted. Anchoring to the exact trial root would require the run to record that root, which it does not. The suffix rule still rejects every case above.

scripts/evals/lib/EquivalenceEnvironment.psm1
WilliamBerryiii: Mostly fixed; one item is not, and is called out below rather than left silent.

- **Local-variable naming.** Done. Every local in the shared module is PascalCase. Confirmed against `powershell.instructions.md`, which specifies PascalCase for variables.
- **Temporal narrative comments.** Done. The Stage 1/Stage 2 prose is replaced with a current invariant. This also matches the repository rule that temporal markers are removed from code files during any edit.
- **Exported function help.** Done. All ten exported functions have `.PARAMETER` sections for every typed parameter.
- **Pester isolation.** Done. A file-level `AfterAll` removes the module.
- **Test ownership.** Not done. Both options you offered require real restructuring, and it was judged better to leave it visible than to rush a file reorganization at the end of this batch. Happy to take it in a follow-up, or now if you would rather it not ship as-is.

`npm run test:ps -- -TestPath "scripts/tests/evals/"` passes, and PSScriptAnalyzer is clean on every changed file.

While fixing this I also removed a stale reference in the module docstring: it cited `agile-coach` as an exempt advisory agent, and that agent no longer exists in the repository.

scripts/evals/lib/EquivalenceParsing.psm1
WilliamBerryiii: Fixed. The dashboard sort headers are now focusable with `tabindex="0"`, operable with Enter and Space, and carry `aria-sort` that updates across all columns as the sort changes, with `scope="col"` on every header and a `:focus-visible` outline so keyboard focus is visible.

.github/workflows/eval-validation.yml
WilliamBerryiii: Addressed except where it depends on a decision that is not mine to record.

- **Branch state.** Merged with current `main`, no conflicts. The branch is no longer behind.
- **Status checks.** `Table Format Check` was genuinely failing, on `docs/contributing/validation.md` and `evals/baseline-equivalence/README.md`. Both are formatted and the check passes locally. Remaining checks are being monitored on the current head.
- **Migration target.** The pin is `0.13.0`, not the `0.11.0` in the narrative or the `0.12.0` cited here; it moved during review. Version literals across the docs have been replaced with repository-pinned wording so the same drift cannot recur, and the PR description needs a matching update.
- **Rollout authorization.** Not recorded, because that is a human decision rather than a code change. Comparison remains report-only at every tier and no comparative pass/fail state exists, so merging does not enact a comparative policy. If you want an explicit authorization recorded before merge, please say where it belongs.
- **Validation evidence.** Local evidence is the full `scripts/tests/evals/` suite plus PSScriptAnalyzer, both clean. Hosted evidence exercising launch-based invocation and timeout containment against the final head requires a credentialed run and cannot be produced locally.

evals/baseline-equivalence/README.md
WilliamBerryiii: Fixed, with one deliberate deviation.

- **Agent coverage.** Reworded to say other agents can be materialized but are not selected and are not meaningfully evaluated, and that the shared-stimulus and scope-guard claims apply to `rpi-agent` only.
- **Divergence-guard exit behavior.** Documented: individual outcomes are report-only, but total guard collapse is promoted to `runHealthFailures` and fails calibration and CI.
- **ADR part count.** Corrected to eight. Your count here was right and the other review comment suggesting seven was not.
- **ADR cross-reference.** Now points at part 4, which is where tier naming is defined.
- **Seed README frontmatter.** Kept as a governed fixture exception rather than adding frontmatter. The seed workspace is copied verbatim into both eval variants, so adding frontmatter would change the content under test to satisfy a linter. The exclusion sits alongside the existing test-fixture exclusions and now carries that rationale in `Validate-MarkdownFrontmatter.ps1`.
- **Stale skill-hygiene verification.** Replaced the `0.11.0` literal with repository-pinned wording plus a dated verification, so it cannot go stale against a version bump again.
- **`ms.date`.** Set to today rather than the suggested `2026-08-10`. The repository rule is that `ms.date` is updated to today's date whenever a file carrying it is edited, and these files are being edited now. All ten touched markdown files were updated consistently.

evals/baseline-equivalence/customized/eval.yaml
WilliamBerryiii: **Correction to my earlier reply on this thread.**

I implemented the structured-evidence hardening as specified, CI ran it against the real corpus, and it broke the measurement. Correcting that here rather than leaving the earlier reply standing.

**What happened.** Complete invocation evidence fell from 103/105 and 105/105 to 55/105 and 86/105. Every one of those 138 failures was a false negative.

**Root cause.** I replayed the captured trajectories from the last pre-change CI run. All 105 records still produced a candidate call, so recognition was never the problem. The loss was at correlation: the call whose result actually returns the agent content is frequently a composed command, and the "no pipeline, redirection, substitution, separator, exactly one operand" rule rejects exactly those. Measured directly, 47 of 105 gpt-5.6-luna records take their content **only** from a call that rule discards. Claude lost 2, because it prefers the structured `view` tool. That difference is why the two models reported different numbers.

The forms being rejected are ordinary:

```
find . -path '*/.github/agents/hve-core/rpi-agent.agent.md' -print -exec sed -n '1,240p' {} \;
pwd && sed -n '1,240p' <path>
wc -l <path> && sed -n '1,240p' <path>
if [ -f <path> ]; then sed -n '1,240p' <path>; else printf 'FILE_NOT_FOUND'; fi
cat <path> 2>/dev/null || find . -name rpi-agent.agent.md
```

Relaxing only the metacharacter rule would not have been enough: `find -exec` ends in `\;`, so the single-operand rule still rejects it.

**What I changed.** The target is now resolved per candidate path rather than by constraining command shape. A token resolves when it equals the expected path, ends with it on a whole-segment boundary so trial-rooted absolute paths work, or is a trailing fragment of it so glob-named targets work. Leading `./`, `*/`, and `**/` are stripped first. Any `..` segment is rejected.

Your substring concern is still addressed. These remain non-evidence: a `.bak` sibling, a differently named agent in the same directory, a partial-segment name such as `evil-rpi-agent.agent.md`, a traversal path, and a delegated prompt naming a second agent file.

**Where I am deliberately not following the suggestion, and why.** `cat <expected> /proc/self/environ` now counts as evidence. That command did read the agent file, so as *invocation evidence* it is valid.

The finding bundles two separate concerns. Measurement validity asks whether this trial read the agent file, and the operative proof is already the correlated result carrying the agent's own content, which the gate requires. Credential exposure asks whether a command also read something sensitive, and that risk exists whether or not the trial is counted; it is addressed by no longer publishing raw trajectories as workflow artifacts, which is the change made for the credential thread on this PR.

Enforcing the credential concern inside the measurement gate cost 45% of the evidence on one model and prevented nothing.

**Verification.** Replaying both captured runs through the corrected reader gives 105/105 on both models, with zero failed, missing, wrong-path, or malformed. That is at or above the pre-regression baseline; claude improves on its own baseline because two `cat ... || find ...` fallbacks previously went uncounted. Twelve behavioural tests now encode the real read forms, taken verbatim from the trajectories, alongside the retained negatives. Full evals suite 665 passing, PSScriptAnalyzer clean.

If you would rather keep the strict shell rule despite the measurement cost, say so and I will revert to it, but the numbers above are the trade being made.

jkim323 (approved):

agreaves-ms (approved):

jkim323 (approved):

PR #2599, opened 2026-08-04
refactor(skills): make cross-artifact references portable

### LINKED ISSUE #2598: refactor(skills): make cross-artifact references portable
## Problem

Skills and instructions reference sibling artifacts by relative filesystem path, for example `../../../prompts/accessibility/accessibility-coverage-matrix.prompt.md` and `../../../../docs/agents/mural/credentials.md`.

Those paths only resolve when the referenced artifact happens to be installed alongside the referring one. A user who installs a single collection, or who consumes artifacts through the extension or plugin packaging, gets a reference that points at nothing. The failure is silent: the agent follows a dead path and continues without the guidance it was supposed to load.

## Proposal

Establish a portability rule in `hve-builder.instructions.md` and apply it across the affected artifacts.

- Cross-artifact references name the target artifact and the section wanted, rather than hard-coding a path into another artifact's directory.
- A named artifact is resolved at activation time by searching the host for its `name:`.
- When resolution fails, the consuming artifact warns the user that the capability is unavailable and stops the dependent step, instead of proceeding silently.

Within a single skill, relative paths to that skill's own `references/` remain correct and are unaffected.

## Scope

- `.github/instructions/hve-core/hve-builder.instructions.md` — the rule
- `.github/instructions/skill-security-model.instructions.md`
- 8 skills across accessibility, coding-standards, experimental, installer, security, and shared

Two skills additionally exceed a reasonable single-file size and are split into references as part of the same pass: `hve-core-installer` (931 lines) and `powerpoint`.

## Acceptance Criteria

- [ ] `hve-builder.instructions.md` states the cross-artifact reference rule and the resolve-by-name-and-warn behavior
- [ ] No skill or instruction hard-codes a relative path into another artifact's directory
- [ ] Every converted reference names the target artifact and the section wanted
- [ ] Each converted reference site states what happens when resolution fails
- [ ] `npm run validate:skills`, `npm run lint:frontmatter`, `npm run plugin:validate`, `npm run lint:md`, and `npm run lint:tables` pass

WilliamBerryiii:
# Pull Request

## Description

Skills and instructions referenced sibling artifacts by relative filesystem path — for example `../../../prompts/accessibility/accessibility-coverage-matrix.prompt.md` and `../../../../docs/agents/mural/credentials.md`. Those paths only resolve when the referenced artifact happens to be installed alongside the referring one. A user who installs a single package, or consumes artifacts through extension or plugin packaging, gets a reference pointing at nothing, and the failure is silent.

This PR establishes a portability rule in `hve-builder.instructions.md` and applies it across the affected artifacts. A cross-artifact reference now names the target artifact and the section wanted, and states what happens when the target is unavailable.

Relative paths within a skill to its own `references/` and `scripts/` are correct and unchanged.

Two skills also exceeded a reasonable single-file size and are split into references in the same pass, since the split is what makes their cross-references nameable rather than positional.

## Related Issue(s)

Fixes #2598

## Type of Change

Select all that apply:

**Code & Documentation:**

* [x] Documentation update

* [ ] GitHub Actions workflow
* [ ] Security configuration
* [ ] Dependency update

**AI Artifacts:**

* [x] Reviewed contribution with `hve-builder` and addressed all actionable findings
* [x] Copilot prompt (`.github/prompts/*.prompt.md`)

**Other:**

* [ ] Script/automation (`.ps1`, `.sh`, `.py`)

### Portability rule

| File | Change |
|------|--------|
| `.github/instructions/hve-core/hve-builder.instructions.md` | Adds the cross-artifact reference rule and a `Handling an Unavailable Reference` section defining availability by an observable condition: an artifact is available only when its content is actually present in the current turn |

The rule is stated in terms the agent can check. It does not assert that a host exposes a search-installed-artifacts-by-name capability, because none is documented.

### Reference conversions

| File | Change |
|------|--------|
| `.github/instructions/skill-security-model.instructions.md` | Names the `mural`, `jira`, and `gitlab` skills and defers their location to the registry in `docs/security/security-model.md` |
| `.github/skills/accessibility/accessibility/SKILL.md` | Names the `accessibility-coverage-matrix.prompt.md` prompt by filename, since it carries no `name:` field, and the `Accessibility Surface Inventory` subagent by name |
| `.github/skills/accessibility/accessibility/references/frameworks/wcag-22.md` | Named reference |
| `.github/skills/accessibility/accessibility/references/frameworks/coga.md` | Named reference |
| `.github/skills/accessibility/accessibility/references/frameworks/en-301-549.md` | Named reference |
| `.github/skills/coding-standards/code-review/references/emission-modes.md` | Named reference |
| `.github/skills/coding-standards/code-review/references/output-formats.md` | Named reference |
| `.github/skills/design-thinking/dt-coaching-foundation/references/canonical-deck.md` | Named reference |
| `.github/skills/experimental/customer-card-render/SKILL.md` | Activates the `powerpoint` skill by name at all three sites that previously hard-coded its directory |
| `.github/skills/experimental/customer-card-render/SECURITY.md` | Named reference |
| `.github/skills/experimental/mural/SKILL.md` | Credentials guide and LICENSE referenced by location description rather than relative path |
| `.github/skills/project-planning/rai-planner/references/security-model.md` | Named reference |
| `.github/skills/security/vex/SECURITY.md` | Names `skill-security-model.instructions.md` |
| `.github/skills/shared/backlog-templates/SKILL.md` | Normalizes `compatibility` to a string and states the consuming planner supplies tracker access |

### Size splits

Measured against `origin/main`:

| File | Before | After |
|------|--------|-------|
| `.github/skills/installer/hve-core-installer/SKILL.md` | 988 lines | 491 lines |
| `.github/skills/experimental/powerpoint/SKILL.md` | 412 lines | 260 lines |

New reference files:

* `installer/references/installation-methods.md`
* `installer/references/post-installation-setup.md`
* `installer/references/component-installation.md`
* `powerpoint/references/script-reference.md`

The installer split is **not** a pure relocation. `main` landed a Phase 7 rewrite during this branch's life — "Agent Customization" became "Component Installation", `agent-copy` scripts became `component-copy`, the tracking manifest moved to `schemaVersion` 2, and eject became component-level. `references/component-installation.md` is built from that rewritten content, so the extracted file differs from anything that existed at the branch point. Reviewers should read it as new prose, not as moved prose.

### Behavior fixes found while reviewing the split

| File | Fix |
|------|-----|
| `installer/SKILL.md` | Ten reference-relative script links were broken by the extraction (`scripts/` needed to become `../scripts/`) |
| `installer/SKILL.md` | Restored the MCP capability table, the `/jira-setup` note, and six settings entries dropped during extraction |
| `installer/SKILL.md` | Phase 6 and Phase 7 pointers were bare declaratives over phases that write into a user repository; both now direct the agent to read the reference, follow it, and stop when it is unavailable |
| `installer/references/component-installation.md` | Adds an Upgrade Confirmation Gate. Upgrade mode previously displayed a summary and proceeded, so a `modified` or `ejected` file could be overwritten without a per-file decision |

## Review remediation (commit 6ff085c)

All twelve review threads from @katriendg are addressed in `6ff085c`, with a per-thread reply naming the specific fix.

| Change | Sites |
|--------|-------|
| Unavailable-reference prose added | 11 |
| Sites the review missed, also fixed | 2 (`aria-apg.md` repo-root path, `mural/SKILL.md` trailing newline) |
| `ms.date` refreshed | 4 |
| `metadata.last_updated` refreshed | 6 |

Two changes deserve explicit call-outs.

**The LICENSE suggestion was declined.** Restoring `[LICENSE](../../../../LICENSE)` would reintroduce a skill-root-escaping path, which is the pattern this PR removes. No skill in the repo links LICENSE. The prose now falls back to the `license: MIT` frontmatter field, which is portable. Rationale is on the thread.

**`<powerpoint-skill-dir>` is gone.** Step 2 of `customer-card-render/SKILL.md` now delegates the build to the `powerpoint` skill by name with three named inputs, instead of invoking a script at a path it never explained how to resolve. This also removes a contradiction: the skill's Overview already stated that `Invoke-PptxPipeline.ps1` "is not reimplemented here" while the code fence invoked it directly.

### Known scope boundary: mixed reference conventions in two SECURITY.md files

`vex/SECURITY.md` and `customer-card-render/SECURITY.md` each now contain both a converted bare-name reference and an unconverted repo-root-relative link. In `customer-card-render` the two sit on adjacent lines of the same References list.

This is deliberate. `skill-security-model.instructions.md` mandates the "See also: repo-wide STRIDE model" callout linking `docs/security/security-model.md`, and 34 of the 68 skill-root-escaping links in the repository exist to satisfy that requirement across 11 SECURITY.md files. Removing them requires changing that instruction first, which is a rule-design decision spanning files this PR does not touch.

Tracked in #2646.

### Follow-ups filed

| Issue | Scope |
|-------|-------|
| #2646 | SECURITY.md convention conflicts with the portability rule (34 instances, 11 files) |
| #2647 | No CI lane validates reference content under the four ignored artifact globs |
| #2648 | `metadata.last_updated` has no freshness validator |
| #2649 | `hve-builder.instructions.md` ships with 2 of 7 packages containing affected skills |
| #2650 | `nvd.nist.gov` needs a link-check ignore pattern |

#2649 is the one worth reading: the rule this PR establishes ships with only two of the seven packages containing the affected skills, which is why the per-site prose is functionally required rather than redundant.

## Testing

Validation lanes owning the changed areas, all passing on the current head:

| Command | Result |
|---------|--------|
| `npm run validate:skills` | 0 errors, 0 warnings |
| `npm run lint:frontmatter` | 0 errors |
| `npm run lint:md` | 518 files, 0 issues |
| `npm run lint:tables` | Pass |
| `npm run plugin:validate` | Pass, marketplace validated, 14 plugins |
| `npm run lint:json` | Pass |
| `npm run lint:ai-artifacts` | Pass |

CI run `31066517582` (PR Validation) completed with conclusion `success`.

Note on link-check signal strength: `pr-validation.yml` sets `soft-fail: true` on the markdown link check, and `markdown-link-check.yml` reports success regardless of findings. A green check on that lane carries no information. The ten broken installer links fixed here would have passed CI untouched; they were caught by manual resolution of every relative link in the diff.

Note on `lint:md` coverage: `.markdownlint-cli2.jsonc` ignores `.github/instructions/**`, `.github/agents/**`, `.github/prompts/**`, and `.github/skills/**`. Every file in this PR lives under one of those globs, so `lint:md` reports zero issues without inspecting any of them.

### Required Checks

* [ ] Tests added for new functionality (if applicable) — not applicable; this PR changes documentation only and adds no executable behavior

### AI Artifact Contributions

* [x] Used `hve-builder` review mode to review contribution
* [x] Addressed all actionable findings from the `hve-builder` review
* [x] Verified contribution follows common standards and type-specific requirements

### Required Local Checks

* [x] Local validation aggregate: `npm run validate:local`
* [x] Documentation validation (if docs changed): `npm run validate:docs`
* [x] Link validation: `npm run lint:md-links`

## Security Considerations

## Additional Notes

**Review state.** An `hve-builder` static review was run across four lanes and its High findings on phase-pointer imperativeness, the missing upgrade confirmation gate, and the two remaining path-resolved references are fixed in this branch. The behavior gate was **not** run — `hve-builder-tester` dispatch was unavailable in the authoring host. The `hve-builder` checkboxes above are checked to reflect that the static review ran and its actionable findings were addressed; the behavior gate remains outstanding and is tracked as follow-up. Rerun it when tester dispatch is available.

**Known open findings, deliberately not fixed here.** Medium and Low findings remain on Phase 7 naming consistency, `builder`/`builder-tester` versus `hve-builder`/`hve-builder-tester` naming, missing Stop rules in two skills, and missing `Use when` triggers. These are follow-up work.

Items from the original list that are now resolved in this branch: failure behavior at the previously-absent reference sites, the `mural/SKILL.md` reference wording, trailing-blank-line hygiene, and `metadata.last_updated` currency across every edited skill file.

**Not implemented.** No validator enforces the portability rule. A `lint:skill-path-scope` check was designed during this work but not built, so the rule is currently convention-only and will drift. A `vally-tests` eval spec covering the unavailable-reference behavior is likewise outstanding. Both are worth filing before this rule accumulates more dependents.

**Relationship to the topology PR.** This branch has merged current `origin/main` and is mergeable. The parallel collection-topology-consolidation PR (#2602) is still open and has **not** merged, so the two do stack: both touch `hve-core-installer/SKILL.md`. Whichever merges second will need a regeneration pass rather than a hand merge of generated files.

**Reviewable decision.** The substance to review is the rule in `hve-builder.instructions.md` — whether a cross-artifact reference should name its target and state failure behavior, and whether "available only when content is present in the current turn" is the right availability test. The line-count of the installer split is large but secondary.


.github/skills/installer/hve-core-installer/SKILL.md
```diff
@@ -348,249 +348,7 @@ Would you like to proceed with this method, or see alternatives?
...
-
-<!-- <method-6-devcontainer> -->
-```jsonc
-{
-  "onCreateCommand": "git submodule update --init --recursive",
-  "updateContentCommand": "git submodule update --remote lib/hve-core || true"
-}
-```
-<!-- </method-6-devcontainer> -->
+Each supported installation path is documented end to end in [references/installation-methods.md](references/installation-methods.md). Select the path resolved by the decision matrix above and follow it there.
```
katriendg: **[Functional/Standards] Medium — Phase 4 pointer omits unavailable-reference behavior**

The sibling Phase 6/7 pointers in this file state "warn the user ... and stop" when their reference file is unavailable; this Phase 4 pointer doesn't, even though Phase 4 is the first phase every clone-based install reaches.

Suggested:
```markdown
Each supported installation path is documented end to end in [references/installation-methods.md](references/installation-methods.md). Select the path resolved by the decision matrix above and follow it there. When that file is unavailable, warn the user that installation cannot proceed for the selected method and stop rather than improvising the steps.
```

.github/skills/experimental/mural/SKILL.md
```diff
@@ -100,7 +100,7 @@ Devcontainer decision tree:
 * **Remote-SSH**: set `MURAL_CREDENTIAL_BACKEND=file` unless a SecretService daemon is configured on the remote host.
 * **WSL2**: leave `MURAL_CREDENTIAL_BACKEND=auto` when WSLg + SecretService is installed; otherwise set `MURAL_CREDENTIAL_BACKEND=file`.
 
-See [Mural Credentials guide](../../../../docs/agents/mural/credentials.md) for backend selection rules, the bootstrap walkthrough, devcontainer recipes, troubleshooting, migration, and the security model.
+See the Mural Credentials guide under the repository documentation tree (`docs/agents/mural/credentials.md`) for backend selection rules, the bootstrap walkthrough, devcontainer recipes, troubleshooting, migration, and the security model.
```
katriendg: **[Standards] Medium — credentials guide link regressed to unlinked backticked path**

The prior text linked the credentials guide with a working relative markdown link; this drops the link and states no unavailable-reference fallback.

Suggested:
```markdown
See the Mural Credentials guide (`docs/agents/mural/credentials.md`) for backend selection rules, the bootstrap walkthrough, devcontainer recipes, troubleshooting, migration, and the security model. When that guide is unavailable, say so and fall back to the backend-selection rules stated above rather than guessing at bootstrap steps.
```

.github/skills/security/vex/SECURITY.md
```diff
@@ -232,7 +232,7 @@ For an active issue tracker entry covering these gaps, see the [hve-core issues
 ## References
 
 * [Repository-wide security model](../../../../docs/security/security-model.md)
-* [Skill security model conventions](../../../../.github/instructions/skill-security-model.instructions.md)
+* Skill security model conventions: `skill-security-model.instructions.md`
```
katriendg: **[Standards] Medium — reference regressed to unlinked backticked path**

Same link-loss pattern: the working relative link to `skill-security-model.instructions.md` was dropped in favor of a bare backticked filename with no unavailable-reference fallback.

Suggested:
```markdown
* Skill security model conventions: `skill-security-model.instructions.md`. When it is unavailable, follow the STRIDE structure already present in this document rather than guessing at the convention.
```

.github/skills/shared/backlog-templates/SKILL.md
```diff
@@ -3,10 +3,9 @@ name: backlog-templates
...
 license: MIT
 user-invocable: true
-compatibility:
-  hosts: ["vscode", "github-coding-agent"]
+compatibility: "Hosts: vscode, github-coding-agent. Reference-only templates; the consuming planner supplies tracker access."
 metadata:
-  authors: ["microsoft/hve-core"]
+  authors: "microsoft/hve-core"
   spec_version: "1.0.0"
   last_updated: "2026-05-09"
```
katriendg: **[Standards] Medium — `metadata.last_updated` not bumped**

`compatibility` and `metadata.authors` were reshaped in this diff (good — brings the file into schema conformance), but `last_updated` was left at `2026-05-09` instead of today's date (`2026-08-06`).

.github/skills/experimental/mural/SKILL.md
```diff
@@ -409,4 +409,4 @@ The following items were intentionally deferred from the v1 full-scope delivery
 | Lineage prefix field placement (`title` vs `text`) | Stakeholder confirmation that `[dt:method=N section=NAME run=ID]` belongs on widget `title` (current v1 spec literal) versus the primary `text` field rendered by most widget types | Planning log PD-6.1 / ID-01  || DT section-map geometry refinement                 | Refine the default section geometry and add canonical sub-sections per method in `assets/dt-sections.default.yml`                                                                  | Mural skill code review      |
 ## License
 
-This skill is distributed under the MIT License. See the repository [LICENSE](../../../../LICENSE) file for the full text.
\ No newline at end of file
+This skill is distributed under the MIT License. See the `LICENSE` file at the repository root for the full text.
```
katriendg: **[Standards] Low — LICENSE reference lost its link, and file still lacks a trailing newline**

Same link-loss pattern as the credentials guide: `LICENSE` is named in backticks with no markdown link. This edit also leaves the file's pre-existing missing trailing newline in place.

Suggested:
```markdown
See the repository [LICENSE](../../../../LICENSE) file for the full text.
```

.github/skills/accessibility/accessibility/references/frameworks/coga.md
```diff
@@ -13,7 +13,7 @@ Source: W3C, Making Content Usable for People with Cognitive and Learning Disabi
 
 ## Licensing posture
 
-COGA is published under the W3C Document License as a Working Group Note. This skill paraphrases pattern intent rather than reproducing the source verbatim, in line with the paraphrase-preferred posture defined in [accessibility-license-posture.instructions.md](../../../../../instructions/accessibility/accessibility-license-posture.instructions.md). Every per-objective reference file cites the canonical W3C URL anchor for each pattern, and any future verbatim quotation must carry the W3C copyright attribution line specified in that instruction file.
+COGA is published under the W3C Document License as a Working Group Note. This skill paraphrases pattern intent rather than reproducing the source verbatim, in line with the paraphrase-preferred posture defined in `accessibility-license-posture.instructions.md`. Every per-objective reference file cites the canonical W3C URL anchor for each pattern, and any future verbatim quotation must carry the W3C copyright attribution line specified in that instruction file.
```
katriendg: **[Standards] Low — no unavailable-reference behavior stated**

Converts the `accessibility-license-posture.instructions.md` path reference to a bare filename but states no fallback if the file is unavailable, unlike `accessibility/SKILL.md`'s conversion in this same PR (which warns and stops).

.github/skills/accessibility/accessibility/references/frameworks/en-301-549.md
```diff
@@ -7,7 +7,7 @@ description: European EN 301 549 V3.2.1 harmonised ICT accessibility standard pa
...
 
-Source: ETSI / CEN / CENELEC, EN 301 549 V3.2.1, <https://www.etsi.org/deliver/etsi_en/301500_301599/301549/>. EN 301 549 reference text is reproduced strictly under the paraphrase-only posture defined in [`accessibility-license-posture.instructions.md`](../../../../../instructions/accessibility/accessibility-license-posture.instructions.md); every clause summary in this skill is paraphrased in the authors' own words and cites the ETSI portal as the authoritative source for the normative wording.
+Source: ETSI / CEN / CENELEC, EN 301 549 V3.2.1, <https://www.etsi.org/deliver/etsi_en/301500_301599/301549/>. EN 301 549 reference text is reproduced strictly under the paraphrase-only posture defined in `accessibility-license-posture.instructions.md`; every clause summary in this skill is paraphrased in the authors' own words and cites the ETSI portal as the authoritative source for the normative wording.
```
katriendg: **[Standards] Low — no unavailable-reference behavior stated**

Same as the sibling `coga.md`/`wcag-22.md` conversions: names `accessibility-license-posture.instructions.md` but states no fallback if it's unavailable.

.github/skills/accessibility/accessibility/references/frameworks/wcag-22.md
```diff
@@ -13,7 +13,7 @@ Source: W3C Web Content Accessibility Guidelines (WCAG) 2.2, <https://www.w3.org
 
 ## Licensing posture
 
-WCAG 2.2 is published under the W3C Document License. This skill paraphrases success-criterion intent rather than reproducing normative text verbatim, in line with the paraphrase-preferred posture defined in [accessibility-license-posture.instructions.md](../../../../../instructions/accessibility/accessibility-license-posture.instructions.md). Every per-guideline reference file cites the canonical W3C URL for each success criterion, and any future verbatim quotation must carry the W3C copyright attribution line specified in that instruction file.
+WCAG 2.2 is published under the W3C Document License. This skill paraphrases success-criterion intent rather than reproducing normative text verbatim, in line with the paraphrase-preferred posture defined in `accessibility-license-posture.instructions.md`. Every per-guideline reference file cites the canonical W3C URL for each success criterion, and any future verbatim quotation must carry the W3C copyright attribution line specified in that instruction file.
```
katriendg: **[Standards] Low — no unavailable-reference behavior stated**

Same as the sibling `coga.md`/`en-301-549.md` conversions: names `accessibility-license-posture.instructions.md` but states no fallback if it's unavailable.

.github/skills/coding-standards/code-review/references/emission-modes.md
```diff
@@ -31,7 +31,7 @@ The review should emit results in the most capable native format available. When
...
+2. **Active-engagement self-review gate.** Before the confirmation step, surface coverage from the dispatch manifest: the number of board items still pending or never opened, and an enumerated list of every Critical or High finding with file:line. Ask one active prompt that requires the human to either name which high-severity findings or unopened areas to open now, or explicitly acknowledge proceeding without further review. Keep the draft and review state intact until one of those choices is made. Reuse the existing Code-Review reviewer-responsibility wording from `disclaimer-language.instructions.md` and do not add separate disclaimer prose.
```
katriendg: **[Standards] Low — no unavailable-reference behavior stated**

Converts the `disclaimer-language.instructions.md` path reference to a bare filename but doesn't state what to do if the file's content isn't present.

.github/skills/coding-standards/code-review/references/output-formats.md
```diff
@@ -107,7 +107,7 @@ Authoring rules:
 
 ## Disclaimer and human-review sign-off
 
-Every `review.md` ends with a Disclaimer and Human Review section, always rendered last and never omitted. It contains the verbatim `## Code-Review` `> [!CAUTION]` block from [Disclaimer Language](../../../../instructions/shared/disclaimer-language.instructions.md), followed by an unchecked `- [ ] Reviewed and validated by a qualified human reviewer` checkbox. The agent never checks this box; only a human may convert `[ ]` to `[x]`. This section restores the human-review gate as a codified part of the output contract rather than an emergent behavior of path-attached instructions.
+Every `review.md` ends with a Disclaimer and Human Review section, always rendered last and never omitted. It contains the verbatim `## Code-Review` `> [!CAUTION]` block from `disclaimer-language.instructions.md`, followed by an unchecked `- [ ] Reviewed and validated by a qualified human reviewer` checkbox. The agent never checks this box; only a human may convert `[ ]` to `[x]`. This section restores the human-review gate as a codified part of the output contract rather than an emergent behavior of path-attached instructions.
```
katriendg: **[Standards] Low — no unavailable-reference behavior stated**

Same as the sibling `emission-modes.md` conversion: names `disclaimer-language.instructions.md` but states no fallback if it's unavailable.

.github/skills/experimental/customer-card-render/SECURITY.md
```diff
@@ -221,18 +221,18 @@ flowchart TD
...
+|---------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|---------------------------------------------------------------------------------------------------------------------|
+| G-SUP-1 | The deck build is delegated out-of-process to the experimental powerpoint skill (`Invoke-PptxPipeline.ps1`) and inherits that skill's residual risk (sandboxed `content-extra.py` execution, LibreOffice/MuPDF document parsing). The documented `uv` toolchain bootstrap uses a `curl \| sh` / `irm \| iex` installer. | SupplyChain-Med | Accepted; see the `SECURITY.md` bundled with the `powerpoint` skill and pin the `uv` installer to a vetted release. |
```
katriendg: **[Standards] Low — gap-table citation states no unavailable-reference behavior**

Names the `powerpoint` skill's `SECURITY.md` but states no fallback, unlike the sibling `SKILL.md` in this same directory, which adds stop-on-unavailable language for every `powerpoint` reference it converts.

.github/skills/experimental/customer-card-render/SECURITY.md
```diff
@@ -221,18 +221,18 @@ flowchart TD
...
+| G-INF-1 | Canonical DT artifacts may contain confidential product or customer prose; that content flows verbatim (escaped) into the emitted `content.yaml` and any downstream deck. There is no data-classification gate.                                                                                                         | InfoDisc-Low    | By design; operators must avoid rendering regulated content and control the output directory.                       |
 
 For an active issue tracker entry covering these gaps, see the [hve-core issues list](https://github.com/microsoft/hve-core/issues).
 
 ## References
 
 * [STRIDE Threat Model](https://learn.microsoft.com/azure/security/develop/threat-modeling-tool-threats)
 * [OWASP Top 10](https://owasp.org/www-project-top-ten/)
-* [PowerPoint skill security model](../powerpoint/SECURITY.md)
+* PowerPoint skill security model: the `SECURITY.md` bundled with the `powerpoint` skill
```
katriendg: **[Standards] Low — References-list citation states no unavailable-reference behavior**

Same gap as the gap-table cell above: names the `powerpoint` skill's `SECURITY.md` with no fallback stated.

katriendg (approved):
Thanks for the thorough writeup and the transparent "known open findings" section, it made this review much faster to ground. Ran a functional + standards pass focused on cross-reference resolution and the new portability rule's own consistency (per your own disclosed rerun caveat on the `hve-builder-tester` behavior gate).

Every named reference I sampled resolves correctly, and the installer/powerpoint file splits preserve original content with no drops. One High finding: the Phase 4 pointer in `hve-core-installer/SKILL.md` is the only one of the three Phase pointers (4/6/7) in that file that doesn't state stop-on-unavailable behavior — worth aligning since Phase 4 is the first thing every clone-based install hits. A handful of Medium/Low items round out the rest: a couple of mural/vex references dropped their working markdown link in favor of a plain backticked path, several files were content-edited without bumping `ms.date`/`metadata.last_updated`, and the `customer-card-render` Step 2 build step depends on an unresolved `<powerpoint-skill-dir>` placeholder.
These are all minor hygiene/consistency issues that can be fixed in a follow-up commit, left inline where possible, but the buk of ms.date/metadata.last_updated are not inline ones, so maybe do a pass.

agreaves-ms (approved):

.github/skills/installer/hve-core-installer/SKILL.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515). Applied your suggested wording verbatim — Phase 4 now warns that installation cannot proceed for the selected method and stops, matching the Phase 6 and Phase 7 pointers.

Agreed on the severity: Phase 4 is the first phase every clone-based install reaches, and it gates writes into a user repository, so a silent no-op there is materially worse than a missing attribution pointer.

.github/skills/experimental/mural/SKILL.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515):

> See the Mural Credentials guide under the repository documentation tree (`docs/agents/mural/credentials.md`) for backend selection rules, the bootstrap walkthrough, devcontainer recipes, troubleshooting, migration, and the security model. When that guide is unavailable, say so and fall back to the backend-selection rules stated above rather than guessing at bootstrap steps.

One note on framing: I kept the backticked path rather than restoring a markdown link, which matches the body of your own suggestion. A working relative link here would have to escape the skill root (`../../../../docs/...`), which is the pattern this PR removes.

.github/skills/security/vex/SECURITY.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515). Applied your suggested wording verbatim:

> * Skill security model conventions: `skill-security-model.instructions.md`. When it is unavailable, follow the STRIDE structure already present in this document rather than guessing at the convention.

.github/skills/shared/backlog-templates/SKILL.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515) — bumped to `2026-08-09`, and I took your "maybe do a pass" note and extended it to all six edited skill files. `powerpoint` was the worst at 144 days; `customer-card-render` and `mural` were both over 100.

Something I found while checking this that seems worth recording: `ms.date` **is** enforced. `Invoke-MsDateFreshnessCheck.ps1` runs in `pr-validation.yml` with `soft-fail: false`, and its exclusion list does not cover `.github/skills/**`. It passes here only because it enforces a 90-day staleness threshold and the flagged values were 44 and 38 days old — so it enforces a different rule than the edit-time bump `copilot-instructions.md` mandates. `metadata.last_updated` has no validator at any threshold; the schema checks its shape only. Filing a follow-up.

.github/skills/experimental/mural/SKILL.md
WilliamBerryiii: Partly applied in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515). Trailing newline fixed. I took the fallback prose but **not** the link, and I want to explain why rather than quietly diverge.

`[LICENSE](../../../../LICENSE)` escapes the skill root. `hve-builder.instructions.md` reserves file paths for a skill's own bundled resources precisely so a reference survives extension and plugin packaging, and that link is the shape this PR exists to remove — it resolves only in the repo, from a known depth. I also checked: no skill in the repo links LICENSE, so adding one here would create a unique exception to the rule being established.

What landed instead:

> This skill is distributed under the MIT License. See the `LICENSE` file at the repository root for the full text. When that file is unavailable, rely on the `license: MIT` field in this skill's frontmatter as the authoritative identifier and say that the full text could not be shown.

The frontmatter field is the portable identifier and always travels with the skill, so no capability is lost. Happy to revisit if you disagree.

.github/skills/accessibility/accessibility/references/frameworks/coga.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515):

> When that instruction content is unavailable, apply the paraphrase-only posture stated here and do not quote the source verbatim, because the required attribution line cannot be confirmed.

Framed as stop-on-unavailable rather than degrade, since the consequence of guessing is a licensing violation rather than a missing pointer. Same wording pattern applied to `wcag-22.md` and `en-301-549.md`.

.github/skills/accessibility/accessibility/references/frameworks/en-301-549.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515) with the same wording as `coga.md` and `wcag-22.md` — when the instruction content is unavailable, apply the paraphrase-only posture stated in place and do not reproduce normative wording verbatim, because the required attribution line cannot be confirmed.

.github/skills/accessibility/accessibility/references/frameworks/wcag-22.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515) with the same wording as its `coga.md` and `en-301-549.md` siblings — paraphrase-only posture applies in place, and normative text is not quoted verbatim when the attribution requirement cannot be confirmed.

Related: your three comments here made me check the fourth framework file, and `aria-apg.md` still had a full repo-root path (`.github/instructions/accessibility/accessibility-license-posture.instructions.md`) that this PR should have converted alongside its siblings. Also fixed in the same commit.

.github/skills/coding-standards/code-review/references/emission-modes.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515):

> When that instruction content is unavailable, state that the standard reviewer-responsibility wording could not be loaded and keep the gate itself in force rather than authoring replacement prose.

The gate stays in force either way — only the standard wording is reported missing.

.github/skills/coding-standards/code-review/references/output-formats.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515):

> When that instruction content is unavailable, the verbatim block cannot be reproduced: say so, do not paraphrase a substitute disclaimer, and stop before emitting the review.

This one is stop rather than degrade. The contract requires the `> [!CAUTION]` block **verbatim**, so a paraphrased substitute would silently weaken the human-review gate — worse than not emitting.

.github/skills/experimental/customer-card-render/SECURITY.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515) — the G-SUP-1 status cell now ends with: "When that skill is unavailable, treat this inherited risk as unassessed."

Deliberately *unassessed* rather than *accepted*: G-SUP-1 inherits residual risk from a model that may not be present, and silently treating an unreadable model as covered is the failure mode worth naming in a security document.

.github/skills/experimental/customer-card-render/SECURITY.md
WilliamBerryiii: Fixed in [`6ff085c`](https://github.com/microsoft/hve-core/pull/2599/commits/6ff085c69655e850ad4afb6de2cca0fa68ede515):

> * PowerPoint skill security model: the `SECURITY.md` bundled with the `powerpoint` skill. When that skill is unavailable, treat the delegated build risk as unassessed rather than assuming it is covered.

Note: widening these two cells broke table alignment, which `lint:tables` caught and `format:tables` fixed — hence the reflow in this file's diff.

.github/skills/experimental/powerpoint/references/script-reference.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+```bash
+python scripts/build_deck.py \
+  --content-dir content/ \
+  --style content/global/style.yaml \
+  --output slide-deck/presentation.pptx
+```
+
+Reads all `content/slide-*/content.yaml` files in numeric order and generates the complete deck. Executes `content-extra.py` files when present. Exit codes:
```
chaosdinosaur: **High: Do not automatically execute `content-extra.py` from an untrusted content directory.**

This build path executes every discovered helper in-process, but the current AST denylist is not a sandbox and permits filesystem access such as `pathlib` writes. A supplied or modified slide directory can therefore read or write files with the invoking user's privileges.

Please disable helper execution by default. Require an explicit trusted-content opt-in, then run approved helpers in an isolated process or container with a bounded working directory and no ambient credentials or network access. The command reference should also state that untrusted content directories must never execute helper scripts.

.github/skills/installer/hve-core-installer/references/component-installation.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Component Copy Execution
+
+After confirmation and collision resolution, execute the copy.
+
+**PowerShell:** Run [scripts/component-copy.ps1](../scripts/component-copy.ps1) with `-HveCoreBasePath`, `-TargetRoot`, `-PackageName` (`$selectedPackage`), `-SelectionName` (`starter` or `custom`), and `-Component`. Add `-KeepExisting -Collisions <component paths>` for kept components.
+
+**Bash:** Run [scripts/component-copy.sh](../scripts/component-copy.sh) with the HVE-Core base path, target root, package name, selection name, and component paths as arguments, in that order. Set `KEEP_EXISTING=true` and `COLLISIONS_FILE=<newline-delimited component paths>` for kept components.
+
+Both implementations validate membership, path safety, and the existing manifest schema before the first write, and produce equivalent paths, hashes, manifests, and output.
```
chaosdinosaur, line 198: **High: Reject symlink and reparse-point traversal before component writes.**

The documented path-safety check is lexical, while the copy implementations can follow a pre-existing symlink in a destination parent. A repository containing a linked path below `.github/skills/`, for example, can redirect an approved copy outside `TargetRoot` under the developer's account.

Please walk every existing destination component before collision reporting and again immediately before each write. Reject symbolic links and reparse points, resolve the nearest existing parent physically, and require it to remain under the physically resolved `TargetRoot`. Add adversarial coverage for both the PowerShell and Bash copy implementations.

chaosdinosaur (dismissed):
**Medium: Keep the extracted installer configuration examples internally consistent.**

Two duplicated configuration surfaces currently diverge:

* The Method 4 Codespaces settings omit the accessibility agent and prompt locations, project-planning instructions, and RPI skills present in the canonical clone settings template. A successful Codespaces install can therefore leave those capabilities undiscoverable.
* The standalone GitHub, Context7, Microsoft Docs, and Figma MCP examples place the server at the document root, while the same reference requires servers beneath a top-level `servers` object.

Please align Method 4 with the canonical location set and wrap every standalone MCP example in the documented file shape. A structural test or generator should keep these duplicated examples from drifting again.

---

**Medium: Pin delegated executors and remove mutable pipe-to-shell guidance.**

The post-installation templates persist unversioned `npx -y` package names for MCP servers, allowing a later package publication to execute in a developer session without a reviewed version boundary. The customer-card troubleshooting table also recommends piping a mutable remote `uv` installer directly to a shell, despite the adjacent security model calling for a vetted release.

Please pin MCP packages to reviewed exact versions, preferably through a lockfile with integrity metadata, and document the controlled update path. Replace the pipe-to-shell command with a version-pinned package workflow or separate download, checksum or signature verification, and execution steps.

---

**Medium: Complete and enforce the portable-reference contract consistently.**

The central rule is currently absolute but advisory, and several converted sites still diverge from it:

* No validator or behavior test enforces the prohibition on cross-artifact paths.
* The code-review emission references prescribe different terminal outcomes when disclaimer content is unavailable.
* The skill security-model convention does not define unavailable behavior for its registry, template, or exemplar references.
* Customer-card troubleshooting asks the agent to confirm that the PowerPoint skill resolves, even though the new rule says availability is observable only from content present in the current turn.
* The Mural credentials reference retains the repository-only `docs/agents/mural/credentials.md` location.

Please add an enforced path and unavailable-reference check, align the disclaimer flow on a single stop outcome, and update the remaining sites to name the required artifact or section with explicit observable failure behavior.

---

**Medium: Reconcile the PR evidence and merge-readiness state.**

The description says two HVE Builder checkboxes remain unchecked, but the recorded AI Artifact Contributions checks are marked complete. The required tests checkbox remains unchecked, issue #2598 requires `plugin:validate` but the Testing table does not report it, and GitHub still reports the PR as blocked despite passing checks and a mergeable head.

Please update the narrative to match the actual checkbox state, add behavior or conformance evidence or document why tests are not applicable, record the `npm run plugin:validate` result, and identify and clear the remaining merge gate.

WilliamBerryiii:
Thanks for the thorough review — this addresses the four non-security findings in [`a23f4cfc`](https://github.com/microsoft/hve-core/pull/2599/commits/a23f4cfcd370037d07d8556ca71c1019936ff978). The two High security findings are acknowledged and owned separately; I'm not responding to them here.

I verified each finding against the code rather than applying them directly, so two get pushback with evidence.

## Portable-reference contract — partly applied, partly declined

**Applied.** `skill-security-model.instructions.md` had three converted references with no unavailable behavior, which was a real gap in this PR's own work. All three now state it: the registry index degrades to locating models beside each skill's `SKILL.md`; the exemplars and template both degrade to the self-contained Required Structure section.

**Applied.** The `customer-card-render` troubleshooting row said "activate the `powerpoint` skill and confirm it resolves." You're right that this contradicts the rule — availability is observable only from content present in the current turn, so "confirm it resolves" asks for something the rule defines as unknowable. Now: "Activate the `powerpoint` skill by name. When its content does not arrive, stop and report the build step as unavailable."

**Declined, with reasoning.** The two code-review references intentionally differ on terminal outcome, and I've made the reasoning explicit at both sites rather than collapsing them:

- `output-formats.md` **stops** because the missing content *is* the deliverable. The Disclaimer section requires the `> [!CAUTION]` block verbatim; a paraphrased substitute would silently weaken the human-review gate, which is worse than not emitting.
- `emission-modes.md` **degrades** because that step's gate mechanics are fully specified in place — surface pending board items, enumerate Critical/High findings, require an explicit human choice. The instruction supplies only the reviewer-responsibility *wording*. Halting there would disable a working safety prompt over absent boilerplate.

Your underlying point was right, though: the divergence was unexplained and read as arbitrary. Both sites now name the principle — stop when the missing content is the deliverable, degrade when it's wording for behavior specified in place.

**Mural credentials.** Kept the path but marked it explicitly repository-only, and it no longer implies portability it lacks. Worth noting the general problem you've surfaced: `credentials.md` is a docs page with no `name:` frontmatter and no host that resolves it by identifier, so "name the artifact rather than hard-code a path" has nothing to name it by. The rule has no answer for repository documentation. Filing a follow-up.

**Validator enforcement.** Already tracked in #2647, filed during the previous review round.

## Installer configuration consistency — confirmed, and worse than reported

Both sub-claims reproduce. The Method 4 divergence is larger than you listed — six discrepancies, not four:

| Setting | Canonical | Method 4 |
|---|---|---|
| `chat.agentFilesLocations` | has `accessibility` | missing |
| `chat.promptFilesLocations` | has `accessibility` | missing |
| `chat.promptFilesLocations` | has `data-science` | missing |
| `chat.promptFilesLocations` | no `coding-standards` | **has it** — and `.github/prompts/coding-standards/` does not exist on disk |
| `chat.instructionsFilesLocations` | has `project-planning` | missing |
| `chat.agentSkillsLocations` | has `rpi` | missing |

A Codespaces install was silently losing accessibility agents and prompts, data-science prompts, project-planning instructions, and every RPI skill — while carrying one entry pointing at a directory that isn't there. Fixed.

MCP shape confirmed too: four standalone examples placed the server at document root while a fifth in the same file wrapped under `servers`. All four now use the documented shape.

One caveat on scope: both divergences are **pre-existing on `origin/main`**, not introduced by this PR — it relocated the content during the size split. I fixed them anyway because they're self-contradictions within a single file and the fix is additive and zero-risk.

I also found the canonical template is itself stale against the repo — missing `jira`, `experimental`, `privacy`, `rai-planning` across agents/prompts/instructions, and `gitlab`, `accessibility`, `hve-core`, `github` under skills. Broader drift than this finding and beyond this PR's scope; filing separately.

## Pinned executors and pipe-to-shell — declining for this PR

The concern is legitimate but the finding doesn't apply to this PR's changes.

Every `"command": "npx"` MCP config and every `astral.sh/uv/install` command already exists **verbatim on `origin/main`**. This PR relocated them during the documented size split without altering a character. `git show origin/main:.github/skills/installer/hve-core-installer/SKILL.md` confirms all three of each.

One correction: the finding describes "unversioned `npx -y` package names." `npx -y` appears **zero times** in the installer skill — the configs use `"command": "npx"` with `args`.

Pinning MCP packages to reviewed versions with integrity metadata, and replacing the pipe-to-shell bootstrap with download-verify-execute, are both real supply-chain work. They deserve their own change with its own review, not absorption into a documentation refactor.

## PR evidence and merge-readiness

`npm run plugin:validate` passes — marketplace validated, 14 plugins. Reconciling the description narrative and checkbox state now. The blocked state appears to be this CHANGES_REQUESTED review itself, which clears on re-review.

---

Validation on `a23f4cfc`: `lint:json`, `lint:tables`, `lint:frontmatter`, `validate:skills`, `lint:md`, `lint:ai-artifacts` all pass. Diff is 7 markdown files, no executable changes.

.github/skills/experimental/mural/SKILL.md
WilliamBerryiii: Heads-up that this line changed again in [`a23f4cfc`](https://github.com/microsoft/hve-core/pull/2599/commits/a23f4cfcd370037d07d8556ca71c1019936ff978), after you'd already reviewed and resolved it — flagging so the change doesn't slip past you silently.

A second reviewer (@chaosdinosaur) pointed out that `docs/agents/mural/credentials.md` is a repository-only location: it doesn't exist in a plugin or extension install, so the reference is portable in form but not in fact. That's correct, and it's a different concern from the one you raised.

Your fallback prose is retained verbatim. What changed is that the path is now explicitly marked repository-only rather than presented as a neutral location:

> See the Mural Credentials guide for backend selection rules, the bootstrap walkthrough, devcontainer recipes, troubleshooting, migration, and the security model. In this repository that guide is at `docs/agents/mural/credentials.md`; that path is repository-only and does not resolve in a plugin or extension install. When the guide is unavailable, say so and fall back to the backend-selection rules stated above rather than guessing at bootstrap steps.

I kept the path rather than removing it, for the reason you and I landed on earlier: dropping it destroys a working pointer for in-repo readers, who are most readers. Marking it honestly seemed better than either pretending it resolves everywhere or deleting it.

The underlying gap is real and bigger than this line: `credentials.md` has no `name:` frontmatter and no host resolves it by identifier, so the portability rule's "name the artifact rather than hard-code a path" has nothing to name it by. The rule has no answer for repository documentation pages. Filing that as a follow-up.

Leaving the thread resolved — no action needed from you unless you disagree with the change.

.github/skills/experimental/powerpoint/references/script-reference.md
WilliamBerryiii: Confirmed and fixed in e2807875. You were right, and the fix ended up being different from the one this finding implies — worth explaining why.

**Reproduced first.** Against the shipped module, not extracted logic: `builtins.eval(...)` passes `_validate_content_extra` and reaches `exec_module`, giving arbitrary code execution through `build_slide(..., allow_scripts=False)`. Also passing: `pathlib`, the `open` builtin, decorator position, and comprehension position. So the report was accurate and, if anything, understated — you described filesystem access; the actual result was full escape.

**Hardening the AST checker turned out not to be defensible.** I prototyped the obvious patch — reject attribute calls onto blocked module names — and it fell over immediately:

```
b = builtins; b.eval(...)        → passes
import builtins as _b; _b.eval() → passes
(x,) = (builtins,); x.eval(...)  → passes
```

A name-based AST check can't follow bindings, and banning attribute calls wholesale isn't possible because the documented `content-extra.py` template *is* attribute calls (`slide.shapes.add_shape`, `shape.fill.solid`). Extending the denylist would have invited exactly the next bypass.

**So the fix moved up a level.** The real defect wasn't the checker — it was that execution was unconditional whenever the file existed. `--allow-scripts` gated only whether validation ran. Now:

- Execution is opt-in. A `content-extra.py` present without `--allow-scripts` **fails the build**, naming the file and the flag. It is not skipped — a quiet skip would produce a deck that builds fine with missing drawings, which is harder to notice than a failure.
- `--allow-scripts` authorizes execution and no longer skips the lint.
- The lint gained the attribute-call check anyway, as defense in depth, with a comment stating plainly that it is a lint and not a boundary.
- The `__builtins__` substitution was removed. It never confined anything: `import builtins` returns the genuine module with `eval` intact. It was reassurance, not a control.

**On severity, for the record.** This needed an attacker-controlled `.py` file already inside the slide content directory, and nothing in CI invokes `build_deck.py`. `SECURITY.md` G-EOP-1 already documented the denylist as "not an OS-level sandbox" and directed operators to treat these files as trusted input. So the *class* was a known accepted risk — but `ADV-a` overstated the mitigation, and `pathlib`/`open` were never blocked at all despite the doc implying broad coverage. Both statements are corrected in this commit.

**This is a breaking change** and it's worth flagging: any downstream deck shipping a `content-extra.py` now needs `--allow-scripts`. There are zero such files in this repo, but the skill ships in marketplace packages so the affected population is external. `SKILL.md` documents the change.

Tests: 4 new cases, each verified to fail against pre-fix code. Non-execution is proven by sentinel-file absence rather than log text, since a log assertion can pass while execution still happens.

.github/skills/installer/hve-core-installer/references/component-installation.md
WilliamBerryiii: Confirmed and fixed in e2807875. Reproduced on both shells before changing anything.

**What I found, which is slightly worse than reported.** The check isn't just lexical — it's very close to a tautology. `$targetFull` is built by joining onto `$targetBase`, so `StartsWith` passes essentially by construction:

```powershell
Join-Path $base "../../etc/evil.md"   # StartsWith($base) → True
```

It also admits a sibling directory whose name merely prefixes the base name (`target-evil` against `target`), because no separator terminated the comparison. And with a symlink planted at `<target>/.github/skills`, both implementations wrote outside the target root: PowerShell via `Copy-Item`, Bash via `mkdir -p` + `cp` with no link inspection at all.

**On the cross-platform trap you were pointing at.** `Resolve-Path` and `realpath` resolve links rather than reporting them, and neither handles a leaf that doesn't exist yet on first install. The primitive that actually works is `FileAttributes.ReparsePoint` — it covers Windows junctions *and* symlinks, and PowerShell 7 reports POSIX symlinks with the same attribute. So one check is correct on both platforms.

Both scripts now:

1. Normalize `..` lexically.
2. Compare with a **separator-terminated** prefix, which fixes the sibling-prefix case.
3. Walk every destination ancestor and refuse any reparse point, naming the offending ancestor.
4. Re-verify immediately before the write, narrowing the check-to-write window. The code comment says explicitly that this narrows rather than eliminates the race — it isn't atomic and shouldn't be described as such.

Bash uses an ancestor `-L` walk rather than `realpath -m`, since `-m` is a GNU coreutils extension that isn't dependable on macOS. Behaviour is identical across both, and there's now a parity test asserting matching accept/reject verdicts rather than assuming it.

**Severity, honestly stated.** This needs a pre-existing symlink at the exact destination path inside the target repo — so an attacker who already landed a commit there. And the bytes written are hve-core's own component files, making this an arbitrary-*location* write with benign content: integrity damage and DoS, not code execution. Real, worth fixing, not critical.

**One thing your finding surfaced that I didn't expect.** The installer skill has no `SECURITY.md` at all. The powerpoint skill had `G-EOP-1` documenting its analogous risk as accepted; the installer had nothing, so this class of finding had no documented owner and no threat model to be measured against. I've filed that as follow-up rather than folding a new threat model into this PR.

Tests: 3 new cases plus a parity test, each verified to fail against pre-fix code. The escape cases assert that nothing lands outside the target root, not merely that an exception was raised. Junction coverage skips on Linux — and note for anyone extending these fixtures, `New-Item -ItemType Junction` *reports success on Linux while creating nothing*, so the skip guard checks that the item exists and carries `ReparsePoint` rather than relying on a `catch`.

Not verified here: Windows junction refusal and macOS Bash behaviour, since I have neither host. Both are recorded as gaps rather than claimed as passing.

WilliamBerryiii:
## ⚠️ Scope change: this PR is no longer documentation-only

@katriendg @agreaves-ms — flagging this directly because you both approved on 2026-08-06, when this PR changed 22 Markdown files and zero executable files. That is no longer true, and I would rather tell you than have you find out from the file list.

Two commits since then change executable code, in response to @chaosdinosaur's two High findings:

- **e2807875** — `build_deck.py` and both `component-copy` scripts, plus tests
- **cd9dba54** — documentation follow-up aligning the reference and threat model with the new behavior

### Why they were folded in here rather than split out

Both defects were found by reviewing this branch, both reproduce on `main`, and both are small and self-contained. Splitting them into a separate PR would have meant a second review cycle for two fixes that are already fully evidenced here. That was my call, and the cost is this scope change.

### The one thing that will affect people

**`content-extra.py` no longer executes by default.** It previously ran unconditionally whenever the file existed; `--allow-scripts` only controlled whether the AST validator ran first. Now execution requires `--allow-scripts`, and a script present without it **fails the build** rather than being skipped — a silent skip would produce a deck that builds successfully with missing custom drawings, which is harder to notice than a failure.

This is a breaking change for any downstream deck shipping a `content-extra.py`. There are zero such files in this repo, but the powerpoint skill ships in marketplace packages, so the affected population is external and unmeasured.

The installer change is behavior-preserving: it refuses writes that would land outside the target root, and normal installs are unaffected.

### On severity, stated plainly

Both findings are real and both were reproduced by execution, not inspection. Neither is critical in practice:

- The `content-extra.py` escape requires an attacker-controlled `.py` already in your slide content directory, and nothing in CI invokes the build script. `SECURITY.md` G-EOP-1 already documented this class as accepted risk — though `ADV-a` overstated the mitigation, which is corrected in cd9dba54.
- The installer escape requires a symlink already committed in the repo you are installing into, and the bytes written are hve-core's own component files. That is an arbitrary-location write with benign content: integrity damage, not code execution.

### Verification

- 802 powerpoint tests, 54 installer tests (1 skipped: junctions need a Windows host)
- 7 new adversarial tests, each confirmed to **fail** against pre-fix code
- Cross-shell parity asserted by test rather than assumed
- Escape cases assert nothing lands outside the target root, not merely that an exception was raised
- Not verified: Windows junction refusal and macOS Bash behavior — no such hosts available, recorded as gaps rather than claimed

### What I am asking

Re-review of the executable changes if you have the bandwidth. The documentation refactor you originally approved is unchanged apart from the two powerpoint files above. If you would rather I split the security work into its own PR and revert it here, say so and I will.

jkim323 (approved):

PR #2660, opened 2026-08-11
perf(scripts): scope, batch, and deduplicate markdown link checks

### LINKED ISSUE #2650: Add nvd.nist.gov to the markdown-link-check ignore list
Discovered while running `npm run validate:local` during #2599 remediation.

## Problem

`npm run validate:local` fails at `lint:md-links` with one broken link out of 4089 checked across 1122 files:

```
.github/skills/experimental/powerpoint/SKILL.md:224 -> https://nvd.nist.gov (Status: 0)
```

The link is not dead. `curl https://nvd.nist.gov` returns HTTP 403 in 0.1 seconds. The host is reachable and serving; it rejects automated clients.

## Why an ignore pattern is the right fix

`scripts/linting/markdown-link-check.config.json` already carries ignore patterns for the same class of bot-protected host:

```
!!:s^https://www\.iso\.org/
!!:s^https://ieeexplore\.ieee\.org/
!!:s^https://www\.cyber\.gov\.au/
!!:s^https://www\.security\.gov\.uk/
```

`nvd.nist.gov` belongs in that list and was simply never added.

## Suggested change

```json
{ "pattern": "^https://nvd\\.nist\\.gov" }
```

## Notes

No CI impact today. `pr-validation.yml` sets `soft-fail: true` on the markdown link check, so this only blocks the local aggregate. That soft-fail is itself covered by #2647.

Filed separately from the #2599 remediation because that change set is deliberately prose-only and this edits a linting config.

## Acceptance

- `npm run validate:local` completes without this failure.
- The pattern sits with the other bot-protected hosts so the grouping stays legible.

### LINKED ISSUE #2661: Markdown link check re-scans every markdown file on every pull request
## Problem

`markdown-link-check.yml` validates the whole repository on every pull request. `scripts/linting/Markdown-Link-Check.ps1` then spawns one `markdown-link-check` Node process **per file**, serially, across roughly 1,150 markdown files. Node cold-start cost dominates before a single link is checked.

Two independent inefficiencies compound:

1. **No changed-files scoping.** The link check is the only lane in `pr-validation.yml` without it. `frontmatter-validation`, `yaml-lint`, `ps-script-analyzer`, `skill-validation`, `pester-tests`, `pytest-tests`, and `eval-validation` all already accept `changed-files-only` and are called with `true`.
2. **Serial, single-file CLI invocations.** Each file is an independent process, so the loop is trivially parallelizable but runs one at a time.

## Evidence

`scripts/linting/Markdown-Link-Check.ps1` invokes the CLI once per file inside `foreach ($file in $filesToCheck)`, with the `plugins/` tree and test fixtures already excluded from discovery.

`.github/workflows/pr-validation.yml` calls the reusable workflow with only `soft-fail: true` and no scoping inputs.

Issue #2650 reports a full local run covering 4,089 links across 1,122 files, which gives a sense of the surface being re-walked on every pull request.

## Suggested approach

**Changed-files scoping.** Add `-ChangedFilesOnly` and `-BaseBranch` to `Markdown-Link-Check.ps1`, using the existing `Get-ChangedFilesFromGit` helper in `scripts/linting/Modules/LintingHelpers.psm1`, and intersect the result with current discovery filters so the `plugins/` and fixture exclusions still apply. Add matching `changed-files-only` and `base-branch` inputs to the reusable workflow, following the `skill-validation.yml` and `asset-docs-validation.yml` pattern, and set `fetch-depth: 0` so `merge-base` resolves.

To avoid losing coverage, `weekly-validation.yml` should run the same reusable workflow with `changed-files-only: false`. It does not call the link check today.

**Bounded parallelism.** Convert the per-file loop to `ForEach-Object -Parallel` with a configurable throttle. Result aggregation, console output, and `Write-CIAnnotation` should stay serial so output ordering and annotations are unchanged.

## Behavior that must be preserved

- An empty changed set reports a clean run rather than failing; an unscoped scan that finds nothing still fails.
- Malformed junit XML still marks a file failed regardless of the CLI exit code.
- The results JSON schema, step summary, and broken-link annotations are unchanged.
- `npm run lint:md-links` still performs a full local scan by default.

## Interaction with #2647

#2647 argues that a green markdown link check carries no information today because the lane is `soft-fail: true` on pull requests. Scoping the lane to changed files narrows what it inspects per pull request, so the two should be considered together: the weekly full sweep is what preserves whole-repository coverage, and #2647's proposal to move relative-link enforcement into a non-soft-fail lane remains the stronger guarantee.

## Further optimizations considered but not proposed here

- Batch multiple files per CLI invocation, removing the remaining process spawns.
- Deduplicate external URLs across files; hosts such as `learn.microsoft.com` repeat across hundreds of files and are re-fetched each time.
- Tune `timeout` (currently `20s`) and add `aliveStatusCodes` for bot-hostile hosts.

## Acceptance

- Pull request validation checks only markdown files changed against the base branch.
- A full-repository sweep runs on a schedule.
- Files are checked concurrently with a configurable, bounded throttle.
- Existing behavior above is preserved and covered by tests.

jkim323:
# Pull Request

## Description

The Markdown Link Check job previously revalidated the entire repository on every pull request, started one `markdown-link-check` Node process per Markdown file, and fetched the same external URL again whenever it appeared in another file.

This branch applies three complementary optimizations:

**Changed-file pull request scoping.** `Markdown-Link-Check.ps1` accepts `-ChangedFilesOnly` and `-BaseBranch`, intersects the shared Git changed-file set with the existing discovery filters, and keeps full local scans as the default. Pull request validation checks changed Markdown files only, while `weekly-validation.yml` runs the full repository sweep.

**Bounded multi-file CLI batches.** Markdown targets are sorted into balanced batches capped by `ThrottleLimit` (default 8). Each worker invokes the pinned CLI once for multiple files, and serial JUnit conversion preserves deterministic per-file results, annotations, and fail-closed report handling.

**Run-scoped external URL deduplication.** Source files are first checked with external requests suppressed. Exact external URLs are then partitioned into bounded synthetic aggregates, checked once through the original configuration, and replayed to every source occurrence. Query and fragment variants remain distinct, ignored URLs remain unfetched, malformed aggregate evidence fails only mapped source files, and one temporary workspace owns all derived files and cleanup.

**Bot-protected host handling.** The link-check configuration now ignores `https://nvd.nist.gov`, which is reachable but rejects automated clients, so full local scans no longer report that host as dead.

### Measured impact

| Measure | Before | After | Reduction |
|---------|--------|-------|-----------|
| Full-sweep Node process starts | 1,122 | At most 16 across source and external stages | 98.6% |
| Full-sweep external checks | 620 file-level occurrences | 408 unique exact URLs | 34.2% |
| Pull request file scope | Entire repository | Changed Markdown files only | Depends on PR; typically over 99% for small changes |

A network-free local benchmark measured approximately 304 ms per isolated CLI startup. Eliminating 1,106 starts saves about 5 minutes 36 seconds per full sweep from startup overhead alone, before accounting for avoided network requests and parallel execution.

### Behavior preservation

* An empty changed set reports a clean run; an unscoped scan finding nothing still fails.
* Missing, malformed, duplicate, unexpected, or incomplete JUnit attribution fails closed.
* Broken-link annotations, result JSON keys and counts, step summaries, and sorted source output remain source-file scoped.
* `npm run lint:md-links` still performs a full local scan by default.
* No dependency, public configuration field, or workflow permission was added.

## Related Issue(s)

Fixes #2661

Fixes #2650

## Type of Change

Select all that apply:

**Code &amp; Documentation:**

* [x] Documentation update

**Infrastructure &amp; Configuration:**

* [x] GitHub Actions workflow
* [x] Linting configuration (markdown, PowerShell, etc.)
* [ ] Security configuration
* [ ] Dependency update

**AI Artifacts:**

**Other:**

## Testing

| Check | Status | Evidence |
|-------|--------|----------|
| `npm run test:ps -- -TestPath &#34;scripts/tests/linting/Markdown-Link-Check.Tests.ps1&#34;` | Passed | 54 selected tests passed, 0 failed; 4 Integration-tagged tests were excluded by the repository runner |
| PSScriptAnalyzer 1.25.0 with repository settings | Passed | Both changed PowerShell files reported 0 findings |
| Editor diagnostics | Passed | No diagnostics in either changed PowerShell file |
| `git diff --check` and exact source-scope inspection | Passed | No whitespace errors; implementation remained within the approved source and test files |
| Targeted NVD link validation | Passed | `https://nvd.nist.gov` was reported as `ignored` with the updated configuration |

A post-fix full scan reported 0 broken links across 4,064 checked links. The exact `npm run validate:local` aggregate remains blocked on Windows because the current default source batches exceed the command-line limit; this is separate from the NVD configuration fix.

Coverage includes changed-file discovery, balanced batches, per-file JUnit attribution, exact URL deduplication, query and fragment variants, ignored URLs, shared dead-link replay, property-level aggregate trust, scoped aggregate failures, empty changed sets, temporary cleanup, and configs without `ignorePatterns`.

### Required Checks

### AI Artifact Contributions

### Required Local Checks

## Security Considerations

No new dependencies were added. Workflow permissions remain `contents: read`, checkout retains `persist-credentials: false`, and full history is fetched only so changed-file merge-base resolution works.

## Additional Notes

Reviewer attention is most useful on:

1. **Coverage timing:** pull requests validate changed Markdown files; the weekly job preserves whole-repository coverage.
2. **Exact URL identity:** query and fragment variants intentionally remain distinct because configuration rules can depend on them.
3. **Conservative failures:** untrusted source batches fail their expected files, while untrusted external aggregates fail only files mapped to URLs in that aggregate.
4. **Remaining follow-up:** timeout and `aliveStatusCodes` tuning for bot-hostile hosts remains separate policy work requiring host-specific evidence.


katriendg (approved):
Great one, thanks for the improvement!

Approving, but leaving one thought and suggestion for you: fall back to a full scan for deletions and renames

The changed-Markdown scope can miss broken links when a PR deletes or renames a target, such as an image, script, template, PDF, or another Markdown file, without modifying the Markdown documents that reference it.

How does this look like:

* Continue scanning only changed Markdown files for ordinary PRs.
* If the diff contains any **deletion** or **rename**, run the full Markdown link scan.

Deletions and renames are relatively uncommon and are the changes most likely to invalidate links from otherwise unchanged documents. This preserves the performance benefit for most PRs without requiring a complex reverse-reference index that could miss encoded, reference-style, or generated links.

jkim323:
> Great one, thanks for the improvement!
> 
> Approving, but leaving one thought and suggestion for you: fall back to a full scan for deletions and renames
> 
> The changed-Markdown scope can miss broken links when a PR deletes or renames a target, such as an image, script, template, PDF, or another Markdown file, without modifying the Markdown documents that reference it.
> 
> How does this look like:
> 
> * Continue scanning only changed Markdown files for ordinary PRs.
> * If the diff contains any **deletion** or **rename**, run the full Markdown link scan.
> 
> Deletions and renames are relatively uncommon and are the changes most likely to invalidate links from otherwise unchanged documents. This preserves the performance benefit for most PRs without requiring a complex reverse-reference index that could miss encoded, reference-style, or generated links.

Thanks @katriendg for the thought! Verified the gap is present. The changed set is built with `--diff-filter=ACMR` and `FileExtensions '*.md'`, so a deleted or renamed `.png`, `.pdf`, `.ps1`, or template never enters scope, while the shared config still resolves relative targets that carry a file extension. A PR can break an unchanged document's links today.

I'd like to propose a new suggestion, because measuring the cost changed my mind about *where* the scoping belongs.

### This PR already runs two passes with very different cost profiles

| Pass | Work | Cost driver |
| --- | --- | --- |
| Source | Injects `^https?://` into `ignorePatterns`, so it resolves **local links only** — zero network | file count |
| External | Checks the **deduplicated** unique URL set exactly once via synthetic files | unique URLs, network, 20s timeout, 429 retries |

Changed-file scoping buys almost nothing on the first and nearly everything on the second. Measured on this branch, a **full-repo source pass is 4.0s** (1060 files, 2949 local links, 8-way parallel, no network). The external pass is the ~408 unique URLs already quoted in the PR description.

### Suggestion: scope by link class rather than by change scope

* Always run the **source pass** repo-wide.
* Apply `-ChangedFilesOnly` to the **external pass** only.

The reasoning is causality: a PR *can* break an unchanged document's internal links by deleting, renaming, or moving a target. It *cannot* break that document's external links — those rot on the internet's schedule, which the weekly full scan already owns.

Compared with escalating on deletion or rename, this:

* covers internal links on **every** PR, not only when the heuristic fires;
* depends on no diff heuristics (`--diff-filter`, rename similarity, file-type coverage);
* has no bimodal worst case — under escalation, any deletion or rename PR pays the full external network scan;
* is less code: two scope variables and a filter, versus a detection probe plus escalation plus tests.

It fully covers the cases you listed, since image, script, template, PDF, and Markdown are all internal targets.

### Two pre-existing issues surfaced while measuring

1. **`pathname://` links fail the checker.** Four errors whose targets exist on disk, in `docs/agents/project-planning/README.md`, `docs/agents/project-planning/arch-diagram-builder.md`, and `docs/customization/skills.md`. All three files already carry these links on `main`, so the weekly full-scan lane is likely red today and the changed-only gate hides it. Needs `^pathname://` in `markdown-link-check.config.json`.

2. **`Split-MarkdownTargetBatch` uses a fixed batch count**, so batch size grows with file count. At full-repo scope on Windows, 4 of 8 batches exceed the 8191-character command-line limit (measured 8195 / 9097 / 9112 / 9761) and fail with `The command line is too long.` It fails closed, so it is a confusing error rather than data loss — but it already affects local `npm run lint:md-links` on Windows, and it blocks any repo-wide pass. Length-budgeted batching fixes it.

What are your thoughts?

katriendg:
Thanks @jkim323 for the new scoping and suggestions, I really like the approaches of scoping to local and external links, and batching.

My recommendation is to implement both before merging:

* repository-wide internal-link validation on every PR;
* changed-Markdown external-link validation;
* length-budgeted batches, including a Windows boundary test.

Please ping if you need a re-review/approval.

jkim323:
Thanks @katriendg for the suggestions. I implemented both recommendations:

- Internal-link validation now runs repository-wide on every PR. The source pass suppresses only http/https requests, so links to local Markdown, images, PDFs, scripts, templates, and other relative targets are still resolved across unchanged documents.
- External-link validation is scoped to changed Markdown files when -ChangedFilesOnly is enabled. External links from unchanged files are reported as skipped, and the weekly workflow retains the full repository external sweep.
- Source-file batching is now length-budgeted in addition to being throttle-balanced. On Windows, the calculation uses the 8,191-character cmd.exe limit, and tests cover batches that exceed the budget plus the single-target overflow case.

.github/workflows/weekly-validation.yml
```diff
@@ -92,6 +92,15 @@ jobs:
     permissions:
       contents: read
 
+  markdown-link-check:
+    name: Markdown Link Check (Full Repository)
+    uses: ./.github/workflows/markdown-link-check.yml
+    permissions:
+      contents: read
+    with:
+      soft-fail: true
```
katriendg, line 101: `soft-fail` is true, while the README.md in this branch is switched to false, I guess one of them needs updating

scripts/linting/README.md
```diff
@@ -235,6 +235,8 @@ Purpose: Detect broken links before deployment.
 ##### Features
 
 * Discovers tracked and untracked, non-ignored Markdown files so local validation does not require staging
+* Restricts the scan to files changed against a base branch with `-ChangedFilesOnly` and `-BaseBranch`
```
katriendg: nit: for internal links it's full and not only changed right?

katriendg (approved):
Left two comments to look at, re-approving as overall all good, thanks.

.github/workflows/weekly-validation.yml
jkim323: Good catch on the mismatch, though I think the resolution is a documentation one rather than a value change.

The `Key Inputs` column in that table lists each reusable workflow's **declared input defaults**, not the values its callers pass. Every other row follows the same convention (`spell-check`, `markdown-lint`, `ps-script-analyzer`, and so on all show `soft-fail` (false), matching their `default: false` declarations). `markdown-link-check.yml` also declares `default: false`, so the `(false)` in this branch is consistent; the previous `(true)` was the outlier, because it had been documenting the PR caller's override instead of the default.

So both values are correct as they stand:

- `markdown-link-check.yml` declares `soft-fail` default `false`.
- `pr-validation.yml` and `weekly-validation.yml` each override it to `true` for their lane.

The actual gap was that the column never said these were defaults, which is exactly what made this look contradictory. I have added a note under the table in 56452be8 that states it explicitly and names both caller overrides.

scripts/linting/README.md
jkim323: You are right, and thanks for catching it -- the bullet was wrong, not just imprecise.

`-ChangedFilesOnly` scopes **external-link fetching** only. Internal links are always validated repository-wide, deliberately, because renaming or deleting a target breaks references in files the change never touched. The script's own comment-based help already says this, and so does the workflow input description; the README bullet was the one place that overstated it.

Fixed in 12c44f65. The same overstatement was present in the `Scope:` line under GitHub Actions Integration, so I corrected that too:

- Internal links: repository-wide in every lane.
- External links: changed files on PR validation, full repository on `weekly-validation.yml`.

I also added a bullet for the unique-URL deduplication, since it was a behavior change in this PR that the feature list did not mention.

rezatnoMsirhC (approved):

PR #2709, opened 2026-08-13
fix(scripts): run ruff format check and lock-bound ruff in lint:py

### LINKED ISSUE #2694: fix(tooling): close the gap between npm run lint:py and the CI Python lint job
## Summary

`npm run lint:py` runs `ruff check` only. The CI job runs `ruff check` **and** `ruff format --check`. A locally green Python change can therefore still fail CI. This actually happened during PR #2599 and cost a build cycle plus two incorrect &#34;validation green&#34; reports.

## Correcting the original framing

This was first recorded as &#34;unpinned `ruff&#62;=0.15` means a formatter release can break CI with no code change.&#34; **That is wrong**, and re-verification on `main` shows why:

- `.github/workflows/python-lint.yml:62` installs with `uv sync --locked`.
- Every skill&#39;s `uv.lock` pins an **exact** ruff version, for example powerpoint `0.15.4`, demo-video `0.15.20`, adr-author `0.15.12`.
- Dependabot already covers the `uv` ecosystem weekly.

CI is deterministic. The lockfile is authoritative. Nobody should spend time &#34;fixing&#34; a determinism problem that does not exist.

## The actual defect

| | `npm run lint:py` | CI Python Lint job |
|---|---|---|
| `ruff check` | ✅ | ✅ |
| `ruff format --check` | ❌ | ✅ |

`scripts/linting/Invoke-PythonLint.ps1` documents this deliberately: it defaults to read-only `ruff check` for CI gating, and applies `ruff format` only under `-Fix`. The intent is reasonable — the default lane should not mutate source. The problem is that the local lane&#39;s name and output imply coverage it does not have.

## Options

1. Add a non-mutating `ruff format --check` to the default `lint:py` path so it matches the CI job. Most direct; keeps the lane read-only.
2. Add a separate `lint:py:format` script and reference it from the contributing docs.
3. Document the divergence only, and leave the lanes as they are.

Option 1 seems right — a local lint lane that does not reproduce its CI counterpart is a trap — but the maintainer should decide, since it may be a deliberate command-taxonomy choice under `.github/instructions/ci-owned-validation.instructions.md`. Read that before changing `package.json`.

## Secondary: ruff version drift

Separate and lower value. Ruff has drifted across 11 skills, so formatting behaviour is not uniform repository-wide even though each skill is individually reproducible:

| Version | Skills |
|---------|--------|
| 0.15.4 | powerpoint |
| 0.15.6 | jira, gitlab |
| 0.15.10 | tts-voiceover |
| 0.15.11 | customer-card-render |
| 0.15.12 | mural, adr-author |
| 0.15.15 | vally-tests |
| 0.15.17 | accessibility |
| 0.15.20 | demo-video, vex |

Also note `project-planning/adr-author` declares `ruff&#62;=0.6` in `pyproject.toml` while every other skill declares `&#62;=0.15`. Worth aligning the declared floor even if the locks stay independent.

Converging these is optional. Do **not** let a version bump silently reformat source as a side effect; if a newer ruff wants formatting changes, surface that as its own decision.

## Local Ruff resolution parity

The local runner currently resolves Ruff from each project's `.venv`, then falls back to a global Ruff on `PATH`. CI instead runs `uv sync --locked` followed by `uv run ruff`, so the local and CI commands can use different Ruff versions and produce different formatter verdicts. This is observable today: some projects use the global fallback, and an unsynchronized environment can cross the Ruff 0.15/0.16 boundary where Markdown code blocks enter formatter scope.

Resolve this as part of this issue rather than a separate follow-up. Reuse the lock-bound `uv` execution pattern already used by the local Python test runner, while retaining a clear and tested fallback when `uv` is unavailable. Account explicitly for the large `scripts/evals/moderation` environment so the fix does not introduce an unexplained 935 MB sync cost into the ordinary lint path.

## Acceptance Criteria

- [ ] A local command reproduces the CI Python lint job&#39;s pass/fail verdict, or the divergence is explicitly documented where contributors will see it.
- [ ] `ci-owned-validation.instructions.md` was consulted before changing any `package.json` script.
- [ ] No source is reformatted as an incidental side effect of this change.
- [ ] `npm run validate:skills` and the dependency-pinning validation pass.
- [ ] Local Ruff execution is lock-bound for projects with `uv.lock`, using the project's locked Ruff version rather than an arbitrary `.venv` or global binary.
- [ ] Behavior when `uv` is unavailable is explicit, documented, and covered by tests.
- [ ] The ordinary lint path handles `scripts/evals/moderation` without an unexplained full-environment sync cost.

Verified against `main` at `278eb128`.

WilliamBerryiii:
# fix(scripts): run ruff format check and lock-bound ruff in lint:py

## Description

`npm run lint:py` ran `ruff check` only, while the CI Python lint job ran `ruff check` **and** `ruff format --check`. A locally green Python change could still fail CI. The local runner also picked whichever ruff it found first — the project `.venv`, otherwise a global install — so local and CI runs could execute different ruff versions and reach different formatter verdicts.

This change made the default local lane execute the same command set as CI and bound ruff selection to each project's committed lockfile.

* `fix(scripts)`: added `ruff format --check .` after `ruff check .` in the default branch of `Invoke-PythonLint.ps1`. Both gates always run, so a lint failure cannot mask a formatting failure, and each result records `checkExitCode` and `formatExitCode` separately with its own console diagnostic.

* `fix(scripts)`: added lock-aware ruff resolution to `PythonLintHelpers.psm1` through three new exported functions. `Get-LockedRuffVersion` reads the ruff version pinned by `uv.lock`, `Get-RuffVersionString` reports a candidate binary's version, and `Resolve-ProjectRuff` accepts only an exact match — evaluating the project `.venv` binary (Linux, then Windows layout) before a global `ruff`, and retaining every rejected candidate as diagnostic evidence.
  * The runner never synchronizes dependencies. A missing or stale binary fails the project before ruff executes and reports the required version plus `uv sync --locked` as the setup action.
  * Projects without a `uv.lock` keep the previous `.venv`-then-global fallback and report an explicit `unlocked-fallback` mode, making no version claim.
  * `-Fix` mode resolves through the same resolver, so checks and fixes cannot run different ruff versions. Its command sequence remains `check . --fix` followed by `format .`.

* `test(scripts)`: extended the two owning Pester suites. `PythonLintHelpers.Tests.ps1` gained coverage for lock parsing (valid, ruff-free, malformed, and version-bleed locks), exact `.venv` matching on both layouts, global fallback after a `.venv` mismatch, global mismatch rejection, and all three unlocked-fallback outcomes. `Invoke-PythonLint.Tests.ps1` gained a call ledger asserting exact ruff arguments and ordering, a non-mutation assertion for default mode, separate lint-only and format-only failure contexts, a resolution-failure context asserting ruff is never invoked, an exact `-Fix` sequence assertion, and serialized per-phase field assertions. The former generic failure context was reclassified as combined failure rather than removed, and the existing missing-ruff, exception, discovery, output-persistence, and real-ruff I001 contexts were left intact.

* `docs(architecture)`: corrected the two `lint:py` rows in `docs/architecture/workflows.md` and added a "Python Lint Parity" subsection naming the three remaining execution differences — CI provisions the environment while the local runner only verifies it, local discovery covers projects the per-PR matrix excludes, and the hosted lane defaults to running only when a PR changes `.py` or `.pyi` files.

* `docs(scripts)`: rewrote the `Invoke-PythonLint.ps1` Features, Parameters, and Usage text in `scripts/linting/README.md` to describe lock-aware verify-only resolution, the default formatter check, per-phase exit codes, and unlocked fallback.

* `fix(devcontainer)`: changed the per-skill provisioning loop in `.devcontainer/scripts/on-create.sh` from `uv sync` to `uv sync --locked`, so a fresh container installs the exact ruff versions the verify-only lane requires. Every `pyproject.toml` under `.github/skills` has a committed `uv.lock`, so no project loses provisioning.

`package.json` was not modified. `.github/instructions/ci-owned-validation.instructions.md` was consulted first, and the existing `lint:py` wrapper already satisfied the requirement without a script-taxonomy change.

## Related Issue(s)

Closes #2694

## Type of Change

Select all that apply:

**Code & Documentation:**

* [x] Documentation update

* [ ] GitHub Actions workflow
* [x] Linting configuration (markdown, PowerShell, etc.)
* [ ] Security configuration
* [x] DevContainer configuration
* [ ] Dependency update

**AI Artifacts:**

>

**Other:**

## Testing

Bounded Pester suites, run through `npm run test:ps`:

| Suite | Result |
|---|---|
| `scripts/tests/linting/PythonLintHelpers.Tests.ps1` | 28 passed, 0 failed, 0 skipped |
| `scripts/tests/linting/Invoke-PythonLint.Tests.ps1` | 35 passed, 0 failed, 0 skipped |

Repository validation:

| Command | Result |
|---|---|
| `npm run lint:py` | Passed across all discovered projects, reporting the locked ruff version per project |
| `npm run lint:ps` | Passed |
| `npm run lint:md` | Passed — 493 files, 0 issues |
| `npm run lint:tables` | Passed — no reformatting required |
| `npm run lint:frontmatter` | Passed — 837 files, 0 errors, 0 warnings |
| `npm run validate:skills` | Passed — 62 skills, 0 errors, 0 warnings |
| `npm run lint:dependency-pinning` | Passed — 679 dependencies, 0 unpinned |
| `shellcheck .devcontainer/scripts/on-create.sh` | Passed — only a pre-existing SC2016 informational finding on an unrelated line |

The verify-only contract was exercised for real during validation. `scripts/evals/moderation` had ruff 0.16.1 in its `.venv` against a 0.16.2 lock, so the first `lint:py` run failed that project before ruff executed and reported `uv sync --locked`. Running that command as a separate setup step brought the environment to the lock, after which the lane passed. The 935 MB moderation environment was never provisioned implicitly by the lint path.

A scoped diff inspection confirmed that exactly the seven intended files changed, with no `.py`, `.pyi`, `uv.lock`, or `pyproject.toml` diff — satisfying the requirement that no source is reformatted as an incidental side effect.

### Required Checks

* [ ] Documentation is updated (if applicable)
* [ ] Files follow existing naming conventions
* [ ] Tests added for new functionality (if applicable)

### AI Artifact Contributions

### Required Local Checks

## Security Considerations

* [ ] Any new dependencies have been reviewed for security issues
* [ ] Security-related scripts follow the principle of least privilege

## Additional Notes

Behavior change for contributors: a project that commits `uv.lock` must now have a matching ruff already installed before `npm run lint:py` will check it. A stale environment fails that project with the required version and `uv sync --locked` as the named setup action, rather than silently linting with a different ruff. Lint never installs or synchronizes dependencies.

The same `--locked` flag now applies to devcontainer skill provisioning. Because `on-create.sh` runs under `set -euo pipefail` and `find -execdir` returns non-zero when an invocation fails, a skill whose committed lock no longer resolves would stop container creation instead of silently relocking. The existing `pytest-tests.yml` and `fuzz-tests.yml` lanes already run `uv sync --locked` against the same projects, so a stale lock fails CI before it can reach a fresh container.

Ruff version convergence across skills was deliberately left out of scope. Lock-bound execution makes the selected version deterministic per project, but the repository still spans several ruff 0.15.x minors. Ruff 0.16 brings Markdown code blocks into formatter scope and at least one existing Markdown file would fail under it, so that belongs to its own change rather than riding along with a parity fix.

✅ - Generated by Copilot


scripts/linting/Modules/PythonLintHelpers.psm1
```diff
@@ -89,6 +90,190 @@ function Resolve-RuffCommand {
...
+            $resolution.resolvedVersion = $candidateVersion
+            return $resolution
+        }
+
+        $reported = if ($candidateVersion) { $candidateVersion } else { 'unknown version' }
+        $resolution.mismatches += "$candidate ($reported)"
+    }
+
+    $found = if ($resolution.mismatches) { "found $($resolution.mismatches -join ', ')" } else { 'found no ruff candidate' }
+    $resolution.reason = "uv.lock requires ruff $lockedVersion but $found. Run 'uv sync --locked' in this project."
```
jkim323, line 276: Hm could we consider using `uv sync --locked` followed by `uv run ruff` here, matching `python-lint.yml` and the sibling`Invoke-PythonTests.ps1` runner, instead of manually parsing `uv.lock` and comparing `ruff --version` strings? `uv sync --locked` would provide the missing fresh-environment bootstrap, while `uv run ruff` would use the project’s lock-selected executable—the same model CI uses. The current resolver recreates only part of uv’s lock/environment-selection behavior and creates a divergent provisioning contract from `test:py` for the same discovered projects. Was the verify-only divergence intentional and tested against every discovered project?

.devcontainer/scripts/on-create.sh
```diff
@@ -145,7 +145,7 @@ main() {
   rm /tmp/uv.tar.gz
 
   echo "Syncing Python environments for skills..."
-  find .github/skills -name pyproject.toml -type f -execdir uv sync \;
+  find .github/skills -name pyproject.toml -type f -execdir uv sync --locked \;
```
jkim323: Does this propagate a failed per-project `uv sync --locked`? With `find -execdir ... \;`, a nonzero command exit only makes that find expression evaluate false; it does not cause find itself to fail, so `set -euo pipefail` will not stop container setup. A stale skill lock can therefore leave its `.venv` uncreated while `on-create.sh` still reports success, and only fail later in `lint:py` as a missing/mismatched Ruff environment. Could we replace this with an explicit loop or wrapper that returns nonzero when any project sync fails, matching the direct moderation sync below?

jkim323 (changes requested):
Thanks for addressing the script issues and tightening the local/CI Python lint behavior. Have two things for you to look at.

.devcontainer/scripts/on-create.sh
katriendg: The `--locked` change is right, but this loop is now load-bearing for the new lint contract and there's a project it doesn't reach.

`Get-PythonSkill` scans the whole repo. Outside `.github/skills` there are two Python projects:

```
./scripts/evals/moderation        [uv.lock / no .venv]
./.github/hooks/shared/telemetry  [uv.lock, ruff 0.15.16]
```

`.github/hooks/shared/telemetry` is provisioned by neither this loop (`.github/skills` only) nor `copilot-setup-steps.yml`. Under the new verify-only contract, a fresh devcontainer or coding-agent runner will fail `npm run lint:py` — and therefore `npm run validate:local` — on that project with `uv.lock requires ruff 0.15.16 … Run 'uv sync --locked'`. CI won't catch it, because `python-lint.yml` syncs each matrix directory itself, and it happens to have a `.venv` in existing containers, which is why local validation passed here.

Two options: extend provisioning to cover non-skill projects, or narrow local discovery to the roots that are actually provisioned.

Separately, `copilot-setup-steps.yml:123` still runs bare `uv sync` while this moved to `--locked`; the repo instructions ask that both environments be evaluated together, so it's worth deciding deliberately whether they should diverge.

docs/architecture/workflows.md
```diff
@@ -427,6 +427,14 @@ Workflows invoke validation through npm scripts defined in `package.json`:
 | `ci:eval:agent:report`          | Runs `ci:eval:agent:matrix` then `ci:eval:agent:dashboard`                                            | CI-owned noninteractive report lane         |
 | `ci:eval:agent:report:dryrun`   | Runs `ci:eval:agent:matrix:dryrun` then `ci:eval:agent:dashboard`                                     | CI-owned noninteractive dry-run report lane |
 
+### Python Lint Parity
+
+`npm run lint:py` runs the same command set as `python-lint.yml`: `ruff check` followed by the non-mutating `ruff format --check`. Execution conditions still differ in three ways:
+
+* Provisioning: the hosted lane runs `uv sync --locked` itself, while the local runner only verifies that a project committing `uv.lock` already provides that exact ruff version. When it does not, the local run fails before ruff executes and reports `uv sync --locked` as the setup action. Projects without a `uv.lock` fall back to the project `.venv` ruff and then a global ruff, with no version guarantee.
+* Project scope: local discovery covers every directory containing a `pyproject.toml`, including projects that `pr-validation.yml` excludes from its per-PR matrix.
```
katriendg: This bullet accurately describes the scope difference, but it understates the consequence now that resolution is strict.

`scripts/evals/moderation` is excluded from the per-PR matrix precisely because of the torch/detoxify stack (~935 MB). Local discovery still picks it up, and it now requires an exactly-locked ruff there. `npm run lint:py` is part of `validate:local`, so the practical effect is that every contributor must provision that heavyweight environment just to run the local aggregate — including contributors who previously had a green run via a global ruff.

Could this bullet say explicitly that the wider local scope is now *fatal* rather than merely broader? And is it worth giving the runner an exclusion that mirrors the CI one (or a `-SkipProject`), so the local aggregate doesn't inherit a cost that CI deliberately declined to pay per-PR?

scripts/linting/Invoke-PythonLint.ps1
```diff
@@ -69,20 +72,27 @@ function Invoke-PythonLint {
...
                 Write-Host "`nRunning ruff --fix and ruff format in $skillPath..." -ForegroundColor Cyan
             } else {
-                Write-Host "`nRunning ruff in $skillPath..." -ForegroundColor Cyan
+                Write-Host "`nRunning ruff check and ruff format --check in $skillPath..." -ForegroundColor Cyan
             }
 
             Push-Location $skillPath
             try {
-                $ruffCmd = Resolve-RuffCommand -SkillPath $skillPath -GlobalRuffAvailable $globalRuffAvailable
+                $resolution = Resolve-ProjectRuff -SkillPath $skillPath -GlobalRuffAvailable $globalRuffAvailable
```
katriendg, line 82: Worth reconciling this with the sibling lane. `scripts/linting/Invoke-PythonTests.ps1` (lines ~89-96) already does lock-aware resolution, but it *provisions* — `uv sync --locked --dev` when `uv.lock` exists, `uv sync --dev` otherwise — and then executes through `uv run`.

So after this change, `npm run test:py` repairs exactly the environment `npm run lint:py` refuses to repair, and a contributor blocked by the lint failure can unblock it by running the test lane. Going through `uv run ruff` here would have given exact CI command parity *and* automatic provisioning for free.

The verify-only choice is defensible — the PR description makes a good case that lint shouldn't implicitly install 935 MB — but right now it's a silent departure from an established in-repo precedent. A short comment here (or a line in the parity section) stating that lint intentionally verifies while the test lane provisions would keep the next person from "fixing" the inconsistency in the wrong direction.

katriendg (commented):
Thanks for taking this on — closing the local/CI gap on `ruff format --check` and binding ruff selection to `uv.lock` is a genuine correctness win, and the per-phase exit codes plus the "Python Lint Parity" section make the remaining differences legible instead of folklore. I pulled the branch and ran both Pester suites in isolation (62 passed, 1 skipped), checked PSScriptAnalyzer against the repo settings (clean), and validated `Get-LockedRuffVersion` against all 16 committed `uv.lock` files — it extracted the right version every time and matched the installed `.venv` ruff wherever one exists. The doc claims about `python-lint.yml` and the per-PR matrix exclusion also check out against the workflows.

I've left three inline comments, all about the blast radius of the new verify-only contract rather than the implementation itself: one project that nothing provisions, the moderation eval becoming a hard blocker for `validate:local`, and an asymmetry with the sibling `Invoke-PythonTests.ps1` lane. None of them are objections to the approach — flagging them so the decisions are explicit rather than implicit.

scripts/linting/Modules/PythonLintHelpers.psm1
WilliamBerryiii: Thanks for calling out the divergence. It is intentional: `lint:py` remains verify-only so the local-safe aggregate does not install project dependencies, while `test:py` provisions before testing. I excluded the heavyweight moderation project from local lint discovery, retained telemetry, added repository-wide coverage proving every eligible project has a lock containing Ruff, and documented the distinction in both linting references.

Validation on `779ef3f5`: 30/30 helper tests, 35/35 lint-runner tests, and `lint:py` passed all 15 eligible locked projects.

.devcontainer/scripts/on-create.sh
WilliamBerryiii: Good catch. I replaced `find -execdir` with a null-delimited `sync_python_environments` loop that records missing locks or failed child syncs, continues checking the remaining projects, and returns nonzero to fail setup.

On Ubuntu at `779ef3f5`, the failure-propagation and missing-lock tests executed and passed; all 5 setup-contract tests passed with zero skips, and ShellCheck passed.

WilliamBerryiii: Addressed both setup gaps. The shared discovery now reaches every lint-eligible Python project, including `.github/hooks/shared/telemetry`, and both the devcontainer and coding-agent setup call the same `uv sync --locked` function. The heavyweight moderation project remains on its separate explicit sync path.

Ubuntu validation at `779ef3f5` passed all 5 setup-contract tests and ShellCheck; local `lint:py` also passed telemetry and all 14 other eligible projects.

docs/architecture/workflows.md
WilliamBerryiii: Agreed on the practical consequence. Local lint discovery now explicitly excludes `scripts/evals/moderation`, matching the intent of the per-PR exclusion, so `validate:local` no longer requires the torch/detoxify environment just to run Ruff. The architecture and linting docs now state that boundary directly.

At `779ef3f5`, `lint:py` discovered and passed 15 eligible projects; moderation was absent and telemetry remained included.

scripts/linting/Invoke-PythonLint.ps1
WilliamBerryiii: Made this asymmetry explicit. The lint runner now states that it intentionally verifies an existing lock-matching Ruff without provisioning, while `Invoke-PythonTests.ps1` provisions before testing. Both documentation surfaces describe the same contract and the setup paths that prepare lint-eligible environments.

Validation at `779ef3f5` includes 35/35 lint-runner tests plus clean Markdown, table, and frontmatter checks for the updated docs.

jkim323 (approved):

agreaves-ms (approved):

PR #2712, opened 2026-08-14
feat(plugins): consolidate HVE Core into one plugin

agreaves-ms:
# feat(plugins): consolidate HVE Core into one plugin

## Description

This PR consolidated HVE Core distribution around one checked-in plugin manifest, one marketplace locator, one extension identity, and one VSIX release path. It removed the retired multi-package generation, package matrix, plugin ZIP, and release-candidate surfaces while preserving the installer, release-channel, provenance, and Marketplace publication contracts that remain part of the one-plugin architecture.

### Plugin Membership and Consumers

* Made **.github/plugin.json** the deterministic authority for 55 agents, 48 commands, 56 rules, 71 skills, and the shared telemetry hook.
* Reduced **.github/plugin/marketplace.json** to one relative `hve-core` locator and added deterministic synchronization and drift checking through *Sync-PluginManifest.ps1*.
* Updated the installer to select components from the live plugin manifest, retain PowerShell/Bash behavior parity, and write package-free installation state.
* Updated extension preparation and packaging to produce one `hve-core` extension without package matrices, package IDs, or package-specific README swapping.
* Validated manifest and prepared contribution paths before projection and staging, including containment and Git pathspec checks.

### Release and Publication

* Unified Stable and PreRelease component membership while retaining their distinct version, source, tag, cadence, and Marketplace channel behavior.
* Replaced the channel-aware version updater with exact, prevalidated JSON updates across the five repository version owners.
* Reconciled releases against one versioned VSIX, complete sidecars, required singleton assets, and explicitly allowed optional assets.
* Retained VSIX provenance, SPDX, Sigstore, in-toto, dependency SBOM, and Stable OpenVEX handling while removing plugin ZIP packaging and evidence.
* Kept Marketplace dependency installation outside the protected environment and required attestation verification before Azure OIDC publication.

### Documentation and Tests

* Updated architecture, contribution, installation, customization, release, security, and role guidance for the one-plugin model.
* Consolidated plugin inventory documentation and removed retired package-specific pages.
* Replaced package-generation and catalog-projection tests with focused manifest, installer, extension, workflow, version, and release-asset coverage.

## Related Issue(s)

None.

## Type of Change

Select all that apply:

**Code & Documentation:**

* [x] Documentation update

* [x] GitHub Actions workflow
* [ ] Security configuration
* [ ] Dependency update

**AI Artifacts:**

>

**Other:**

**User Request:**

> Install a custom selection of HVE Core agents, prompts, instructions, and skills from the current plugin manifest.

**Execution Flow:**

1. Read `.github/plugin.json` as the installable membership authority.
2. Resolve the requested complete or custom component selection.
3. Run collision and containment preflight checks before writes.
4. Copy tracked component files and record package-free schema version 2 installation state.
5. Reuse the recorded components for upgrade detection and replay.

**Output Artifacts:**

* Selected HVE Core component files under the target `.github` artifact directories
* Installation tracking with the selected profile, component paths, versions, hashes, status, and stable maturity
* Report-only collision output when preflight mode is selected

**Success Indicators:**

* Every selected path exists in the current plugin manifest.
* PowerShell and Bash implementations report matching selections and collision outcomes.
* Undeclared, missing, escaping, or colliding paths fail before target writes.

## Testing

Validation was not rerun during PR generation, following the requested use of existing session evidence.

* Full PowerShell suite: 3,525 passed, 0 failed, 7 skipped, and 113 not run
* Extension and workflow suite: 211 passed with zero failures
* Release suite: 79 passed with zero failures
* Installer suite: 141 passed with zero failures and one unrelated platform capability skip
* PSScriptAnalyzer: 251 PowerShell files with zero findings
* Workflow checks: actionlint passed 66 workflows, permissions passed 256 of 256 checks, and dependency pinning passed 654 of 654 checks
* Plugin validation: 55 agents, 48 commands, 56 rules, 71 skills, and the shared telemetry hook passed
* Local aggregate: all locally controlled children passed; five unchanged skill environments lacked local Ruff installations, and the remaining children passed separately
* Markdown links: `npm run lint:md-links` passed
* Security analysis: no sensitive data, new dependency, new privilege scope, or unpinned action was found
* Diff assessment: 125 changed files matched the one-plugin implementation, test, and documentation scope

Manual product testing was not performed.

### Required Checks

### AI Artifact Contributions

### Required Local Checks

* [ ] Local validation aggregate: `npm run validate:local` (all locally controlled children passed; five unchanged Ruff environments were unavailable)
* [ ] Documentation validation (if docs changed): `npm run validate:docs` (not rerun during PR generation)
* [ ] Spell checking: `npm run spell-check` (not rerun during PR generation)
* [x] Link validation: `npm run lint:md-links`

## Security Considerations

* [ ] Any new dependencies have been reviewed for security issues (N/A - no dependencies changed)

## GHCP Artifact Maturity

| File                                                      | Type         | Distribution  | Maturity | Notes      |
|-----------------------------------------------------------|--------------|---------------|----------|------------|
| `.github/instructions/pull-request.instructions.md`       | Instructions | Repository    | N/A      | Local only |
| `.github/skills/installer/hve-core-installer/SKILL.md`    | Skill        | HVE Core      | Stable   | All builds |
| `.github/skills/security/gh-code-scanning/SKILL.md`       | Skill        | HVE Core      | Stable   | All builds |

## Additional Notes

* Release maintainers still need to verify remote branch protection and deployment environment rules before release operations.
* Stable accepts `dependency-diff.md` only when the comparison job produces it; the asset is optional during release reconciliation.


.github/plugin/marketplace.json
```diff
@@ -8,528 +8,10 @@
...
-    {
-      "name": "hve-core-all",
-      "source": {
-        "source": "github",
-        "repo": "microsoft/hve-core",
-        "path": ".github"
-      },
-      "description": "Full bundle of opinionated, rapidly evolving HVE Core agentic SDLC patterns and tools",
+      "source": ".github",
+      "description": "Opinionated, rapidly evolving HVE Core agentic SDLC patterns and tools",
```
rezatnoMsirhC, line 14: Nitpick: These fields duplicate metadata from `.github/plugin.json`, but the new catalog validation only compares the entry name and version. Please either remove repeated fields that the marketplace schema does not require or extend `Test-PluginCatalog` to compare every repeated field with the canonical manifest. Otherwise, description, author, homepage, repository, license, or keywords can drift while `plugin:validate` still passes.

scripts/plugins/Sync-PluginManifest.ps1
```diff
@@ -0,0 +1,620 @@
...
+    $violations = @()
+    $changed = $expectedJson -ne $committedJson
+
+    if ($changed) {
+        if ($Check) {
+            $violations += "$script:PluginManifestFile is out of sync with tracked components"
+            $violations += @(Compare-PluginComponentSet -Committed (ConvertFrom-Json $committedRaw -AsHashtable) -Expected $manifest)
+        }
+        else {
+            Set-Content -LiteralPath $manifestPath -Value $expectedJson -Encoding UTF8 -NoNewline
```
rezatnoMsirhC, line 645: Please validate the marketplace catalog before writing the manifest. The current ordering updates `.github/plugin.json`, then can return a catalog violation, leaving a failed invocation with a partially modified repository. Run `Test-PluginCatalog` first and write only when no violations remain. A regression test should combine manifest drift with an invalid catalog and verify that the manifest remains byte-identical.

docs/getting-started/package-migration.md
```diff
@@ -83,34 +74,32 @@ channel-switch behavior must be confirmed in the installed host.
...
-| `ise-hve-essentials.hve-gitlab` | `ise-hve-essentials.hve-project-planning`                                       |
-
-The `hve-github` row needs both extensions because its skills split across two packages: the backlog workflows moved to project planning, and the `gh-code-scanning` skill moved to security.
+| If you installed                  | Install instead               |
+|-----------------------------------|-------------------------------|
+| `ise-hve-essentials.hve-ado`      | `ise-hve-essentials.hve-core` |
+| `ise-hve-essentials.hve-github`   | `ise-hve-essentials.hve-core` |
+| `ise-hve-essentials.hve-jira`     | `ise-hve-essentials.hve-core` |
+| `ise-hve-essentials.hve-gitlab`   | `ise-hve-essentials.hve-core` |
+| `ise-hve-essentials.hve-core-all` | `ise-hve-essentials.hve-core` |
```
rezatnoMsirhC: Please update this migration inventory to include the extension identities retired by this consolidation. The former `hve-core-all` extension name was `hve-hve-core-all`, as asserted by [`Prepare-Extension.Tests.ps1:49`](https://github.com/microsoft/hve-core/blob/2be87b7ffe311daddd6f5fe4a11773efd9ae03e3/scripts/tests/extension/Prepare-Extension.Tests.ps1#L49) and [`Prepare-Extension.Generation.Tests.ps1:146`](https://github.com/microsoft/hve-core/blob/2be87b7ffe311daddd6f5fe4a11773efd9ae03e3/scripts/tests/extension/Prepare-Extension.Generation.Tests.ps1#L146). This follows from the removed [`Get-ExtensionIdentity` mapping](https://github.com/microsoft/hve-core/blob/2be87b7ffe311daddd6f5fe4a11773efd9ae03e3/scripts/extension/Modules/ExtensionIdentity.psm1#L59), which prefixes every package except `hve-core` with `hve-`. The full Marketplace ID was therefore `ise-hve-essentials.hve-hve-core-all`, not `ise-hve-essentials.hve-core-all`. The previous catalog also produced separate extensions for `coding-standards`, `data-science`, `design-thinking`, `experimental`, `installer`, `project-planning`, `rpi`, and `security`; please map those retired identities to `ise-hve-essentials.hve-core` as well so users can remove stale installations.

rezatnoMsirhC (approved):
I completed a semantic-preservation and local-safe validation pass against the local remediation state built on this PR head.

| Check | Result |
|-------|--------|
| Former resolver versus generated dependency closure | Passed for all 38 formerly expanding seeds with 0 mismatches, missing components, or extra components |
| Plugin synchronizer tests | 65 passed, 0 failed, 0 skipped |
| Complete installer regression suite | 152 passed, 0 failed, 1 existing Windows-only platform skip |
| PowerShell and Bash installer parity | Passed |
| Plugin synchronization and validation | Passed; repeat generation remained byte-identical |
| Skill validation | Passed for 72 skills with 0 errors and 0 warnings |
| Generated documentation drift check | Passed with 237 files unchanged |
| Frontmatter, PowerShell, Markdown, and Bash syntax checks | Passed |
| Stale maturity-contract assertion | Passed; retired maturity output is absent and distribution membership is manifest-grounded |
| Diff hygiene | Passed with no whitespace errors |

Disclaimer: [The Windows directory-junction containment test](https://github.com/microsoft/hve-core/blob/281ac6a5dba97bb6a14ba8c727b6f0833a5fec06/.github/skills/installer/hve-core-installer/tests/component-copy.Tests.ps1#L789-L817) was skipped because the owning Pester workflow and this validation environment run on Linux. The test and containment implementation predate this PR and are unchanged from its merge base, so this is a pre-existing repository-level coverage limitation rather than a regression introduced by these changes. Related POSIX symbolic-link containment tests passed.

These results apply to the local remediation state, so the corresponding semantic fixes still need to be incorporated into the PR.

scripts/plugins/Sync-PluginManifest.ps1
```diff
@@ -0,0 +1,620 @@
...
+        }
+    }
+
+    $violations += @(Test-PluginCatalog -RepoRoot $RepoRoot -Manifest $manifest)
```
WilliamBerryiii: **High - blocking.** `Test-PluginCatalog` is accumulated here unconditionally: this line sits outside both the `if ($changed)` and `if ($Check)` branches above, so **write mode also enforces catalog-to-`package.json` version parity**. Combined with the main block's `exit 1` (line 586), this aborts both promotion lanes.

See the paired comment on `.github/workflows/release-stable.yml:424`.

Two viable fixes, not mutually exclusive:

```powershell
# Scope the gate to verification so a write run cannot fail on state it is about to normalize
if ($Check) {
    $violations += @(Test-PluginCatalog -RepoRoot $RepoRoot -Manifest $manifest)
}
```

and/or swap the invocation order in the two promotion workflows.

scripts/plugins/Sync-PluginManifest.ps1
WilliamBerryiii, line 645: **Low.** The manifest is written to disk here, but `Test-PluginCatalog` does not run until line 573. A catalog violation therefore exits 1 with `.github/plugin.json` already rewritten, leaving a mutated tracked file behind on a failed `npm run plugin:sync`, and a mutated working tree inside the promotion jobs.

Suggest computing violations first and writing only when clean:

```powershell
$violations += @(Test-PluginCatalog -RepoRoot $RepoRoot -Manifest $manifest)

if ($changed) {
    if ($Check) { ... }
    elseif ($violations.Count -eq 0) {
        Set-Content -LiteralPath $manifestPath -Value $expectedJson -Encoding UTF8 -NoNewline
    }
}
```

.github/workflows/release-stable.yml
```diff
@@ -434,83 +398,34 @@ jobs:
...
-          git show "$SOURCE_SHA:.github/plugin/marketplace.json" > "$CANDIDATE_SOURCE_CATALOG"
-
-          # Shared files keep their merged source content; the target version
-          # fields, the restored prior catalog, and the candidate record are
-          # rewritten from target-owned and immutable-source state.
-          pwsh -NoProfile -File "$GITHUB_WORKSPACE/scripts/release/Update-VersionFiles.ps1" \
+          # The selected published source decides component membership, and the
+          # target baseline decides the version. Neither transforms a source
+          # ref, records a release candidate, nor generates a plugin root.
+          pwsh -NoProfile -File "$GITHUB_WORKSPACE/scripts/plugins/Sync-PluginManifest.ps1" \
```
WilliamBerryiii: **High - blocking. This lane fails on every run.**

At this point the tree is version-skewed by construction:

- line 386 - `restore_from_base required CHANGELOG.md .github/plugin/marketplace.json` restores the catalog from `refs/remotes/origin/$BASE_BRANCH` (`release/stable`), so it carries `STABLE_BASELINE`
- lines 391-393 - `git checkout "$SOURCE_SHA" -- package.json ...` sets `package.json` to the promoted PreRelease version
- **line 424 (this line)** - Sync runs in write mode, reads `package.json` via `Get-RepositoryVersion`, compares it to the restored catalog, and exits 1
- line 426 - `Set-RepositoryVersion.ps1` would reconcile them, but is never reached

A stable promotion by definition promotes a version differing from the stable baseline, so the two values cannot be equal.

Swapping the two invocations resolves it - `Set-RepositoryVersion.ps1` already writes `marketplace.json` `$.metadata.version` and `$.plugins[0].version` (asserted by `Set-RepositoryVersion.Tests.ps1:148-149`), so running it first leaves the tree self-consistent before the gate evaluates:

```yaml
          pwsh -NoProfile -File "$GITHUB_WORKSPACE/scripts/release/Set-RepositoryVersion.ps1" \
              -Version "$TARGET_VERSION" \
              -RepoRoot "$PWD"
          pwsh -NoProfile -File "$GITHUB_WORKSPACE/scripts/plugins/Sync-PluginManifest.ps1" \
              -RepoRoot "$PWD"
```

Note this is not covered by the 123 passing checks: PR CI never runs release workflows, and every fixture in `Sync-PluginManifest.Tests.ps1` builds `package.json` and the catalog from the same `$Version` parameter. Worth adding a fixture with deliberately mismatched versions.

.github/workflows/release-prerelease-prepare.yml
```diff
@@ -345,83 +343,34 @@ jobs:
...
-          git show "$SOURCE_SHA:.github/plugin/marketplace.json" > "$CANDIDATE_SOURCE_CATALOG"
-
-          # Shared files keep their merged source content; the target version
-          # fields, the restored prior catalog, and the candidate record are
-          # rewritten from target-owned and immutable-source state.
-          pwsh -NoProfile -File "$GITHUB_WORKSPACE/scripts/release/Update-VersionFiles.ps1" \
+          # The merged source decides component membership, and the target
+          # baseline decides the version. Neither transforms a source ref,
+          # records a release candidate, nor generates a plugin root.
+          pwsh -NoProfile -File "$GITHUB_WORKSPACE/scripts/plugins/Sync-PluginManifest.ps1" \
```
WilliamBerryiii: **High - blocking.** Same ordering defect as `release-stable.yml:424`.

- line ~307 - `merge_ref "$SOURCE_SHA" theirs` brings `package.json` in from `main`
- line 337 - `restore_from_base` restores the catalog from `release/prerelease`, carrying `PRERELEASE_BASELINE`
- **line 369 (this line)** - Sync runs in write mode and compares the two
- line 371 - `Set-RepositoryVersion.ps1`, never reached

Unlike the stable lane this is not guaranteed to fail, but the versions differ whenever `main` is ahead of the prerelease baseline, which is the ordinary condition for a promotion. Apply the identical swap.

.github/workflows/plugin-validation.yml
```diff
@@ -41,26 +41,8 @@ jobs:
       - name: Create logs directory
         run: mkdir -p logs
 
-      - name: Guard tracked plugin output and symbolic links
-        run: npm run lint:plugin-output
-      - name: Validate marketplace manifest
-        run: npm run lint:marketplace
+      - name: Validate plugin manifest and marketplace catalog
+        run: npm run lint:plugin-manifest
```
WilliamBerryiii, line 45: **High - blocking.** Removing the `Guard tracked plugin output and symbolic links` step (with `Assert-NoTrackedPluginOutput.ps1` and the `lint:plugin-output` script) deletes the repository's **only** check for git index mode `120000`. Verified: zero remaining matches for `120000` anywhere under `scripts/` or `.github/`.

This lands in the same change that makes `.github` the verbatim distributed plugin root, so the blast radius grows exactly when the guard disappears:

1. `Get-TrackedPluginFile` lists a tracked symlink (`git ls-files` reports it as an ordinary path) and `Get-PluginComponentSet` classifies it into `plugin.json`
2. `Test-PluginComponentCoverage` calls `Test-Path` (line 435), which **follows** the link, so containment passes
3. `Copy-PreparedArtifacts` stages it into the VSIX, dereferencing it into a published Marketplace artifact
4. For the CLI plugin channel the `.github` tree is consumed directly, so a link such as `-> ../../.git/hooks/pre-commit` turns a plugin install into an arbitrary local file write

Suggest restoring the mode-`120000` scan as a required step here and in `validate:local`, and rejecting links defensively inside `Get-TrackedPluginFile` before classification.

docs/getting-started/package-migration.md
```diff
@@ -83,34 +74,32 @@ channel-switch behavior must be confirmed in the installed host.
...
-| `ise-hve-essentials.hve-github` | `ise-hve-essentials.hve-project-planning` and `ise-hve-essentials.hve-security` |
-| `ise-hve-essentials.hve-jira`   | `ise-hve-essentials.hve-project-planning`                                       |
-| `ise-hve-essentials.hve-gitlab` | `ise-hve-essentials.hve-project-planning`                                       |
-
-The `hve-github` row needs both extensions because its skills split across two packages: the backlog workflows moved to project planning, and the `gh-code-scanning` skill moved to security.
+| If you installed                  | Install instead               |
+|-----------------------------------|-------------------------------|
+| `ise-hve-essentials.hve-ado`      | `ise-hve-essentials.hve-core` |
+| `ise-hve-essentials.hve-github`   | `ise-hve-essentials.hve-core` |
+| `ise-hve-essentials.hve-jira`     | `ise-hve-essentials.hve-core` |
```
WilliamBerryiii: **High - blocking.** This table lists 5 retired identities, but the PR retires 13. Verified by diffing catalog plugin names between the merge-base and head: `coding-standards`, `data-science`, `design-thinking`, `experimental`, `installer`, `project-planning`, `rpi`, and `security` are removed and have **no row here**.

That matters more than usual because this page states the Marketplace offers no deprecation or tombstone signal, so an already-installed retired extension keeps surfacing commands that no longer resolve. Anyone who installed one of those eight has no actionable instruction. The prose above ("Remove any retired domain, utility, or hve-core-all plugin registration") names no identity.

Suggest adding the eight missing rows plus the matching CLI rows:

```
copilot plugin uninstall coding-standards@hve-core
copilot plugin uninstall data-science@hve-core
...
```

.github/plugin/marketplace.json
```diff
@@ -8,528 +8,10 @@
     "name": "Microsoft"
```
WilliamBerryiii, line 8: **High.** This reduction retires nine installable package names (`coding-standards`, `data-science`, `design-thinking`, `experimental`, `hve-core-all`, `installer`, `project-planning`, `rpi`, `security`) plus the `hve-core-all` `starter` profile.

That is a breaking consumer change, but the PR description checks "Bug fix" and "New feature", leaves "Breaking change" unchecked, and leaves "Changes are backwards compatible (if applicable)" unchecked with no note - so the description contradicts itself. `docs/getting-started/package-migration.md` is in this diff and tells users to uninstall retired extensions, which is the definition of a consumer-visible break, yet the description never references it.

Suggest checking **Breaking change** and adding a Breaking Changes section enumerating the retired package names, the `starter` profile, the installer `-PackageName` parameter, the `selection.package` tracking field, the four removed npm commands, and the withdrawn plugin ZIP / `plugin-release-evidence.json` release assets - each with its migration path.

Separately: the entry `source` changes from the object form to the bare relative string `.github`. Worth confirming the Copilot CLI marketplace resolver treats a relative source identically for consumers who added the marketplace by GitHub repository.

package.json
```diff
@@ -19,8 +19,7 @@
     "lint:md-links": "pwsh -File scripts/linting/Markdown-Link-Check.ps1",
     "lint:frontmatter": "pwsh -NoProfile -File ./scripts/linting/Validate-MarkdownFrontmatter.ps1 -WarningsAsErrors -EnableSchemaValidation",
     "lint:adr-consistency": "pwsh -NoProfile -File scripts/linting/Validate-AdrConsistency.ps1 -Paths docs/planning/adrs",
```
WilliamBerryiii, line 21: **Medium.** With `lint:marketplace` and `scripts/linting/schemas/marketplace-manifest.schema.json` both deleted, `plugin:validate` reduces to `lint:plugin-manifest && lint:hooks`.

`Test-PluginCatalog` does not validate what the deleted schema required: `metadata.description`, `owner.name`, the `^[a-z0-9-]+$` plugin-name grammar, the `^https://` shape of `homepage`, or `license`/`description` presence. It also returns early on the first source violation, so a malformed `source` masks every component-coverage result in the same run.

This lands in the same PR that changes the catalog `source` from an object to a bare relative string, so schema enforcement is lost exactly when the schema shape changes - a structurally invalid catalog can now reach a release tag with validation green. Suggest retaining a JSON-schema check for the single-locator shape and wiring it back into `plugin:validate` and `validate:local`.

.github/workflows/release-stable-publish.yml
```diff
@@ -408,81 +376,25 @@ jobs:
...
-              -CandidateAction Apply \
-              -CandidateSourceCommit "$SOURCE_COMMIT" \
-              -CandidateSourceCatalog "$CANDIDATE_SOURCE_CATALOG" \
-              -BaselineTag "$BASELINE_TAG"
+              -RepoRoot "$RELEASE_REPO"
 
           cd "$RELEASE_REPO"
           git config user.name 'github-actions[bot]'
           git config user.email '41898282+github-actions[bot]@users.noreply.github.com'
           git add --all
```
WilliamBerryiii, line 392: **Medium.** With `plugin-package.yml`, `Assert-PluginReleaseEvidence.ps1`, and both `upload-plugin-packages` jobs deleted, the plugin channel loses build provenance, the Syft SPDX SBOM, the `.sigstore.json` bundle, the `.intoto.jsonl` envelope, and the repository-level `plugin-release-evidence.json`. `verify-provenance` no longer downloads `-p '*.zip'`, and the release notes no longer publish `gh attestation verify`.

CLI plugin consumers now install from a git ref with no signed artifact, SBOM, or attestation to verify, so tampering at rest or in transit has no cryptographic detection path. The VSIX channel keeps its full chain, so only this channel regresses.

Suggest publishing and attesting a deterministic tree-digest evidence document for the released `.github` root, bound to version, tag, and source commit, and verifying it alongside the VSIX. Recommending a standalone supply-chain review to score the resulting SLSA/Sigstore posture.

scripts/extension/Prepare-Extension.ps1
```diff
@@ -29,126 +27,127 @@ param(
     [string]$Channel = 'Stable',
```
WilliamBerryiii: **Medium.** Verified: `$Channel` appears exactly twice in this file - this parameter default, and a `Write-Host` at line 493. `Invoke-PrepareExtension` no longer accepts it, and the PR's own `Prepare-Extension channel parity` test asserts both channel values produce byte-identical `package.json` and `README.md`.

Yet `extension-package.yml:88` still passes it and `scripts/extension/README.md` still documents it as a live parameter, so two commands advertise different behaviour while producing identical output. `PSReviewUnusedParameter` is excluded in `PSScriptAnalyzer.psd1`, so nothing flags it.

Either retire the parameter along with its callers and the two `extension:prepare*` npm scripts, or keep it and make the contract explicit:

```powershell
.PARAMETER Channel
    Reporting label only. Component membership and generated output are identical
    on Stable and PreRelease, so this value never changes what is written.
```

docs/plugins/hve-core.md
```diff
@@ -1,140 +1,62 @@
...
-| **vally-tests**           | experimental | Authors Vally conformance tests for prompts, instructions, agents, and skills, including refusals for jailbreak, prompt-injection, harmful-elicitation, TOS, CoC, and PII-extraction stimuli                                                                                                     |
-
-### Hooks
-
-| Name          | Maturity     | Description                                                                |
-|---------------|--------------|----------------------------------------------------------------------------|
-| **telemetry** | experimental | Records Copilot session lifecycle events to local telemetry for reporting. |
-
-<!-- END AUTO-GENERATED ARTIFACTS -->
+HVE Core is the single plugin and extension identity for all distributable HVE Core content.
```
WilliamBerryiii, line 14: **Medium.** This page replacement drops the `> [!CAUTION]` adoption-risk block (verified present at the merge-base, absent at head) warning that HVE Core is a rapidly evolving framework best treated as patterns and learning rather than a production dependency.

This is not documentation-only: `Prepare-Extension.ps1:275` reads this file via `Get-PluginDocumentBody` and substitutes it into `{{BODY}}` (line 280), and the same change hard-codes `{{MATURITY_NOTICE}}` to an empty string. So the **published Marketplace listing loses both its maturity notice and its adoption-risk caveat in the same change**, with no replacement added to `extension/templates/README.template.md`. The PR description does not mention removing this disclosure.

docs/plugins/hve-core.md
```diff
@@ -1,140 +1,62 @@
...
+
+Install the plugin from a registered marketplace:
+
+```bash
+copilot plugin install hve-core@hve-core
+```
+
+Install the extension as `ise-hve-essentials.hve-core`. For a repository-owned subset, use `hve-core-installer` to choose all manifest components or a custom selection. The installer records `selection.profile` and `selection.components` in `.hve-tracking.json` and does not copy hooks.
+
+The full path inventory remains machine-readable in `.github/plugin.json`. Agent, prompt, instruction, and skill reference pages are available under `docs/reference/`.
```
WilliamBerryiii, line 37: **Medium.** These counts replace the `<!-- BEGIN AUTO-GENERATED ARTIFACTS -->` region (verified present at the merge-base, absent at head), but nothing regenerates or validates them: `plugin:validate` checks manifest membership and hooks, and `docs:generate:check` owns `docs/reference/`.

Every future artifact addition or removal must now be mirrored here by hand or this page - and the Marketplace README it feeds - goes stale. The drift risk is already realised inside this PR: the description says 56 rules while this table and the manifest both say 57.

Suggest either emitting the table from `Sync-PluginManifest.ps1` into a generated region and gating it in CI, or dropping the Count column in favour of qualitative wording that cannot drift.

docs/customization/packages.md
```diff
@@ -12,70 +12,51 @@ keywords:
...
-Hooks are per-plugin declarations. `hve-core` and `hve-core-all` each include the telemetry hook. VS Code has no declarative hook contribution point, so extension users configure hook locations manually. Hooks are not copied during selective installation.
+The plugin manifest includes the telemetry hook. VS Code has no declarative hook contribution point, so extension users configure its location manually. Hooks are not copied during selective installation.
 
 ## Validation and Package Staging
 
-Ordinary validation reads canonical `.github` sources and never creates a
-repository-root `plugins/` tree. Extension preparation writes
-`extension/package*.json` and `extension/README*.md`. Generated ZIP and VSIX
-paths are host-specific package layouts, not catalog membership vocabulary.
+Ordinary validation reads canonical `.github` sources and never creates a repository-root `plugins/` tree. Extension preparation refreshes the single `extension/package.json` and `extension/README.md`. Packaging creates one VSIX.
```
WilliamBerryiii, line 43: **Medium.** This line still asserts that ordinary validation "never creates a repository-root `plugins/` tree", and `.github/copilot-instructions.md:157` still states such a directory "is forbidden as validation or package output" - but `Assert-NoTrackedPluginOutput.ps1` and the `lint:plugin-output` script are deleted in this PR, so nothing enforces either statement.

The practical risk is low now that `Generate-Plugins.ps1` no longer exists to produce such a tree, but the docs currently describe a guardrail that does not exist. Either restate this as a factual description of what the build does, or restore a minimal guard (which would also address the tracked-symlink gap flagged on `plugin-validation.yml`).

.github/skills/installer/hve-core-installer/references/component-installation.md
```diff
@@ -226,19 +184,18 @@ The copy writes `.hve-tracking.json` at the target root using schema version 2:
...
   },
   "files": {
     ".github/agents/hve-core/rpi-agent.agent.md": {
       "component": "agents/hve-core/rpi-agent.md",
       "kind": "agent",
       "maturity": "stable",
-      "version": "3.3.106",
+      "version": "3.2.2",
       "sha256": "<hash>",
       "status": "managed"
```
WilliamBerryiii, line 200: **Medium.** Two consumer-visible installer behaviours change here with no migration note in the PR description.

1. `.hve-tracking.json` stays at schema version 2 but `selection.package` is dropped and `upgrade-detection` no longer emits `INSTALLED_PACKAGE`. Under the new replay contract, an upgrade **stops** and asks the user to reselect if any recorded component is no longer declared in `.github/plugin.json` - and the new tests explicitly assert `skills/security/owasp-docker` is rejected. Existing consumers can hit a hard stop rather than an upgrade.
2. Per-component maturity is gone: both `component-copy.ps1` and `component-copy.sh` hardcode `Maturity = 'stable'`, and the confirmation prompt drops the "N component(s) are labeled preview or experimental" warning. Components previously disclosed as experimental are now recorded and displayed as stable.

The tests correctly encode the new behaviour; the gap is that the PR describes this only as "write package-free installation state". Suggest documenting both in the description and in `package-migration.md`, and considering whether a maturity map belongs in `plugin.json` rather than defaulting every component to `stable`.

.github/workflows/extension-package.yml
```diff
@@ -14,36 +14,29 @@ on:
...
     runs-on: ubuntu-latest
     permissions:
       contents: read
-    outputs:
-      matrix: ${{ steps.discover.outputs.matrix }}
     steps:
       - name: Checkout code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: ${{ inputs.source-ref }}
```
WilliamBerryiii, line 58: **Low.** Checkout selects code from `inputs.source-ref` here, before the `Validate packaging inputs` gate at line 40, and the ref is never shape-checked against `^[0-9a-f]{40}$` or an exact tag grammar.

To be fair to the gate, it is stronger than it first looks: lines 56-62 read `extension/templates/package.template.json` **from the checked-out ref** and fail unless its version equals `inputs.version`, which catches most wrong refs. The residual gap is that the check runs after checkout has already selected the code, and binds on version equality rather than ref identity - so a ref carrying the expected version but not the intended commit still passes.

`extension-provenance.yml` validates a 40-hex ref *before* checkout, and the deleted `plugin-package.yml` required a full SHA plus tag resolution; that ordering rigor was not carried forward. Trigger is `workflow_call`, so callers are in-repo today - hence Low.

scripts/release/Assert-ReleaseAssetSet.ps1
```diff
@@ -420,29 +279,20 @@ function Assert-ReleaseAssetSet {
...
-        [Parameter(Mandatory = $true)]
-        [ValidatePattern('^\d+\.\d+\.\d+$')]
-        [string]$Version,
+        [Parameter(Mandatory = $false)]
+        [string]$OptionalAssetPath = '',
 
         [Parameter(Mandatory = $true)]
         [ValidateNotNullOrEmpty()]
-        [string]$ReleaseTag,
+        [string]$Version,
```
WilliamBerryiii, line 295: **Low.** This function-level `$Version` keeps only `[ValidateNotNullOrEmpty()]`, while the script-level parameter at line 52 retains `[ValidatePattern('^\d+\.\d+\.\d+$')]`.

`$Version` is the sole input to `Get-ReleaseExpectedVsixName`, so any caller that dot-sources this script and invokes the function directly (as the Pester suite does, and as any future workflow reuse would) can pass a malformed version and shift the entire expected-asset set - a release with the wrong VSIX would reconcile as complete. Defense-in-depth rather than an exploitable path today.

Also worth switching both patterns from `$` to `\z`, since `$` permits a trailing newline in .NET regex. `Set-RepositoryVersion.ps1` already does this correctly.

scripts/release/Assert-ReleaseAssetSet.ps1
```diff
@@ -364,6 +213,20 @@ function Test-ReleaseAssetSet {
         }
     }
 
+    $allowed = [System.Collections.Generic.HashSet[string]]::new(
+        [string[]]@($RequiredAsset + $OptionalAsset), [System.StringComparer]::Ordinal)
+    foreach ($kind in $primaryKinds) {
+        foreach ($primary in ([string[]]@($kind.Expected) + [string[]]@($kind.Actual))) {
+            [void]$allowed.Add($primary)
```
WilliamBerryiii, line 224: **Low.** The new unexpected-asset allowlist is correct for releases produced by this lane, but it will report every plugin ZIP, per-ZIP sidecar, and `plugin-release-evidence.json` on historical `v<version>` / `prerelease-v<version>` tags as `unexpected asset`.

Live risk is low because the workflows only invoke this against the release currently being published. But the PR states historical tags and assets remain immutable supported records, without noting that the current reconciliation tooling no longer accepts their shape. Worth recording that scope limit in the script `.NOTES` and in the description.

scripts/release/Set-RepositoryVersion.ps1
```diff
@@ -0,0 +1,381 @@
...
+        Path     = 'package.json'
+        Location = @(, @('version'))
+        Assert   = @('$.version')
+    }
+    [ordered]@{
+        Path     = 'package-lock.json'
+        Location = @(
+            , @('version')
+            , @('packages', '', 'version')
+        )
```
WilliamBerryiii, line 64: **Low (nit).** `Assert` here and in the `.github/plugin/marketplace.json` entry below are written with one fewer space than the aligned `Path` / `Location` / `Assert` columns in the sibling entries, inside a literal that reads as a table.

docs/docusaurus/src/data/marketplaceCounts.ts
```diff
@@ -2,14 +2,15 @@
...
 
 import * as fs from 'fs';
+import * as path from 'path';
 
 import type { PackageCardData } from './packageCards';
-import { resolvePackageMaturity } from './packageCards';
+import { labelRegistry } from './labelRegistry';
 
 const componentFields = ['agents', 'commands', 'rules', 'skills', 'hooks'];
 const errorPrefix = '[marketplacePackages]';
```
WilliamBerryiii: **Low.** `countMarketplaceComponents` was deliberately renamed to `countPluginComponents`, but the surrounding vocabulary was not carried along: this `errorPrefix` still reads `[marketplacePackages]`, so runtime messages surface as `[marketplacePackages] hve-core: plugin manifest not found`. The module filename and `__tests__/packageCards.test.ts` also keep marketplace-era names while now exercising only manifest behaviour.

A half-applied rename is the state most likely to mislead the next reader about which file owns what. Suggest completing it in this PR.

docs/getting-started/install.md
```diff
@@ -58,30 +46,24 @@ The installer can copy agents, prompts, instructions, and complete skill directo
...
-   CATALOG --> FOCUSED["hve-core<br/>(focused)"]
-   CATALOG --> FULL["hve-core-all<br/>(full bundle)"]
-   CATALOG --> DOMAIN["domain and utility packages"]
-   FOCUSED --> EXT["package-specific VSIX"]
-   FOCUSED --> PLUGIN["package-specific plugin root"]
-   FULL --> EXT
-   FULL --> PLUGIN
-   DOMAIN --> EXT
-   DOMAIN --> PLUGIN
+   REPO["microsoft/hve-core<br/>(canonical source)"] --> MANIFEST[".github/plugin.json<br/>(complete membership)"]
```
WilliamBerryiii, line 57: **Low (two issues on this diagram).**

1. **Correctness:** `CATALOG` has no inbound edge, so `marketplace.json` renders as a second disconnected root - contradicting the prose on this page stating both files live in the repository. The previous diagram had a single connected root, so this is a regression from the rewrite. Adding `REPO --> CATALOG` fixes it.

2. **Accessibility (WCAG 2.2 SC 1.1.1, Level A):** Docusaurus renders this as an inline SVG, and Mermaid emits `<title>`/`<desc>` only when the diagram declares `accTitle`/`accDescr`. Without them the graphic has no accessible name. `docs/agents/backlog/README.md` already uses this pattern.

```
graph LR
   accTitle: HVE Core distribution relationships
   accDescr: The canonical microsoft/hve-core source produces .github/plugin.json, which declares complete membership for both the hve-core plugin and the hve-core VSIX. marketplace.json holds one relative locator to the .github plugin root.
   REPO["microsoft/hve-core<br/>(canonical source)"] --> MANIFEST[".github/plugin.json<br/>(complete membership)"]
   REPO --> CATALOG["marketplace.json<br/>(one relative locator)"]
   CATALOG --> ROOT[".github<br/>(plugin root)"]
```

docs/architecture/workflows.md
```diff
@@ -303,9 +297,8 @@ validated exact channel tag.
 
 ```mermaid
 flowchart LR
-    PRE[Validate prerelease-v tag] --> PRE_MATRIX[Resolve PreRelease package matrix]
-    PRE_MATRIX --> GENERIC[Generic Marketplace publisher]
-    STABLE[Validate v tag and discover Stable matrix] --> GENERIC
+    PRE[Validate prerelease-v tag and catalog] --> GENERIC[Generic Marketplace publisher]
```
WilliamBerryiii, line 306: **Low (WCAG 2.2 SC 1.1.1, Level A).** This rewritten flowchart declares no `accTitle`/`accDescr`, so the Docusaurus-rendered SVG exposes no accessible name and the publication sequence is unavailable to assistive technology.

Low rather than higher because the Channel Tags table and the paragraph below restate the same download/verify/publish sequence in text.

```
flowchart LR
    accTitle: Marketplace publication flow for both release channels
    accDescr: Both the PreRelease and Stable tag validation lanes feed one generic Marketplace publisher, which downloads the VSIX release asset, verifies its lane-specific attestation, and publishes through Azure OIDC and vsce.
```

Only 3 of roughly 60 documentation diagrams currently declare these, so a follow-up issue to backfill the rest may be worth opening separately.

scripts/plugins/README.md
```diff
@@ -1,199 +1,64 @@
...
 ## Scripts
 
-| Script                           | npm Command                | Description                                     |
-|----------------------------------|----------------------------|-------------------------------------------------|
-| Generate-Plugins.ps1             | `npm run plugin:generate`  | Materialize plugin packages in external staging |
-| Validate-Marketplace.ps1         | `npm run lint:marketplace` | Validate marketplace.json plugin manifest       |
-| Assert-PluginReleaseEvidence.ps1 | `npm run plugin:evidence`  | Record or verify canonical release evidence     |
-| Modules/PluginHelpers.psm1       | (library)                  | Plugin materialization and packaging helpers    |
+| Script                  | npm Command           | Description                                                |
+|-------------------------|-----------------------|------------------------------------------------------------|
```
WilliamBerryiii: **Low.** This table has a single row for `npm run plugin:sync`, but the page body directs readers to `npm run plugin:validate` in three places (Source to Manifest Pipeline step 3, Synchronizing After Artifact Changes, and Locator Validation), and never lists `lint:plugin-manifest` at all. A reader scanning the table cannot discover either command, and the one row conflates writing membership with validating the locator - two different modes of the same script.

```markdown
| Script                          | npm Command                    | Description                                       |
|---------------------------------|--------------------------------|---------------------------------------------------|
| Sync-PluginManifest.ps1         | `npm run plugin:sync`          | Write deterministic membership into `plugin.json` |
| Sync-PluginManifest.ps1 -Check  | `npm run lint:plugin-manifest` | Fail on manifest or marketplace locator drift     |
| (aggregate)                     | `npm run plugin:validate`      | Manifest check plus hook manifest validation      |
```

scripts/tests/release/Set-RepositoryVersion.Tests.ps1
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        It 'Leaves an earlier nested package version untouched' {
+            $package = Get-Content -LiteralPath (Join-Path $script:Root 'package.json') -Raw | ConvertFrom-Json
+            $package.metadata.version | Should -BeExactly '0.1.0'
+        }
+
+        It 'Leaves <Path> unchanged' -ForEach @(
+            @{ Path = '.release-please-manifest.json' }
+            @{ Path = '.github/plugin/release-candidate.json' }
```
WilliamBerryiii: **Low.** This case asserts that the version writer leaves `.github/plugin/release-candidate.json` byte-identical - but **this PR deletes that file**, and no production code references it at head (only test files do).

The test therefore proves a guarantee with no subject. Suggest dropping this `-ForEach` entry and the matching fixture at lines 100-105; the `.release-please-manifest.json` half remains meaningful and should stay.

scripts/tests/extension/Prepare-Extension.Tests.ps1
```diff
@@ -3,177 +3,173 @@
...
+
+        return [ordered]@{
+            name        = 'hve-core'
+            description = 'Manifest fixture description'
+            version     = '9.9.9'
+            agents      = @('agents/labs/gamma.agent.md', 'agents/core/alpha.agent.md')
+            commands    = @('prompts/core/build.prompt.md')
+            rules       = @('instructions/core/style.instructions.md')
+            skills      = @('skills/labs/probe', 'skills/core/toolkit')
+            hooks       = 'hooks/core/session.json'
```
WilliamBerryiii, line 62: **Low.** Asserting `Should -Not -Match '(?i)marketplace'` against the raw source text of the script under test is brittle: any incidental occurrence in a comment or help string fails the suite even when behaviour is unchanged. A future comment such as "no longer read from the marketplace catalog" would break this.

The behavioural intent is already covered by the adjacent fixture case - `New-PluginFixtureRepo` deletes `$fixture.CatalogPath` precisely so a surviving read fails loudly. Suggest narrowing to an AST/string-literal assertion or relying on the fixture case. Same pattern appears in `Package-Extension.Tests.ps1` and both installer suites.

.github/workflows/release-stable.yml
```diff
@@ -434,83 +398,34 @@ jobs:
...
-              -SkipManifest \
-              -SkipPluginGenerate \
-              -CandidateAction Record \
-              -CandidateVersion "$VERSION" \
-              -CandidateSourceCommit "$SOURCE_SHA" \
-              -CandidateSourceCatalog "$CANDIDATE_SOURCE_CATALOG" \
-              -BaselineTag "$BASELINE_TAG"
+              -RepoRoot "$PWD"
 
           # The captured changelog cutoff and the exact release intent are
```
WilliamBerryiii, line 432: **High (release-candidate removal).** Anchored here; the line this concerns is `git add --all` at line 449 of this step.

That command stages whatever the preparation tree happens to contain, and the gates that previously anchored this to an immutable source commit are all deleted in this PR: `.github/plugin/release-candidate.json`, the `Verify managed release head reproduces its retained candidate` job in `pr-validation.yml`, and the `Verify retained candidate record` steps in the two publish lanes.

The net effect is that published component membership is derived from a mutable preparation tree and pushed by a credentialed job with nothing to reconcile against. A release-branch head that moved after the triggering event, a partially applied promotion, or any file left behind by a prior step becomes authoritative membership instead of failing the run.

Suggested direction: stage only the files this job owns, and re-derive in `-Check` mode before the push so a tree that does not reproduce the derived manifest fails closed.

```yaml
          git add -- package.json package-lock.json \
            extension/templates/package.template.json \
            .github/plugin.json .github/plugin/marketplace.json
```

.github/instructions/pull-request.instructions.md
```diff
@@ -114,39 +114,18 @@ Priority rules:
...
 
-1. Import `scripts/lib/Modules/MarketplaceHelpers.psm1` and load `.github/plugin/marketplace.json` with `Get-MarketplaceCatalog`.
-2. Build `Get-MarketplaceSourcePolicyIndex`, which includes active membership and `componentMaturity` tombstones.
-3. Normalize `SKILL.md` paths to their skill directory before lookup; use the canonical repository path for every other artifact.
-4. Call `Get-MarketplaceSourceMaturity` for each path. Omit undeclared paths from package maturity claims.
-5. When one source appears in multiple packages, the helper returns the most restrictive value in this order: `removed`, `deprecated`, `experimental`, `preview`, `stable`.
+1. Agents, prompts, and instructions are distributable when they are tracked beneath a package subdirectory and match their canonical suffix.
+2. Skills are distributable when a tracked `.github/skills/<package>/<skill>/SKILL.md` exists and its top-level license has no noncommercial qualifier.
+3. Repository-root artifacts without a package segment are repository-specific.
+4. Do not infer per-artifact maturity. Stable and PreRelease ship the same manifest membership.
```
WilliamBerryiii, line 125: **High - blocking.** This rule says *"Do not infer per-artifact maturity."* But the deletion in this hunk stopped short: the `## GHCP Maturity Output` heading was removed while its **entire body survives** below - the `For deprecated files:` caution template (line 130), the directive *"Always include the maturity summary table when any GHCP files are detected"* (line 138), the four-column `GHCP Artifact Maturity` table (lines 141-152), and the `### GHCP Maturity Acknowledgment` block (line 154).

So this file now tells the agent not to infer maturity, then fifteen lines later requires it to *always* emit a per-artifact maturity table. The `x-hve.componentMaturity` map that was the only possible data source is deleted by this same PR, so the agent has a mandatory output with no source and must either fabricate values or silently skip a rule marked "always".

This is an always-on instruction file for its `applyTo` glob, so the contradiction is live on every PR-description run.

Suggest deleting lines 130-156 and replacing them with a membership summary the new model can actually produce:

```markdown
| File | Type | Membership change |
|------|------|-------------------|
| `new-feature.prompt.md` | Prompt | Added |
| `legacy.agent.md` | Agent | Removed |
```

extension/templates/README.template.md
```diff
@@ -32,14 +32,14 @@ Review channel ownership, lifecycle disclosure, and selective clone adoption:
...
 ## Pre-release Channel
 
-HVE Core publishes the same artifacts on two release channels:
+HVE Core publishes the same complete component set on two release channels:
 
-| Channel     | Description                                         | Maturity Levels                     |
-|-------------|-----------------------------------------------------|-------------------------------------|
-| Stable      | Promoted release cadence                            | `stable`, `preview`, `experimental` |
-| Pre-release | Early access to changes ahead of the next promotion | `stable`, `preview`, `experimental` |
+| Channel     | Description                                         | Membership        |
```
WilliamBerryiii, line 35: **Low.** Anchored on this table; the comment covers two leftovers from the maturity removal.

1. Line 5 still carries `{{MATURITY_NOTICE}}`, which `New-ExtensionReadme` now replaces with the empty string unconditionally (`Prepare-Extension.ps1:279`) - the preview and experimental branches that populated it were deleted. The token survives only because a `(\r?\n){3,}` collapse hides the blank line it leaves, so it is a permanently dead substitution that invites a future author to reintroduce maturity rendering nothing else supports.
2. Line 29 still directs the reader to *"Review channel ownership, lifecycle disclosure, and selective clone adoption"*, pointing at a concept this PR removes.

Since this template renders the published Marketplace README, suggest dropping the token and its `-replace`, and rewording line 29 to something like "Review channel ownership, release assurance, and selective clone adoption". (The token also lingers in `ExtensionTestFixtures.psm1:246`.)

The Membership column itself is a good change - it says exactly what the channels now guarantee.

scripts/plugins/Sync-PluginManifest.ps1
```diff
@@ -0,0 +1,620 @@
...
+    <#
+    .SYNOPSIS
+        Reads the repository version from the root package manifest.
+
+    .PARAMETER RepoRoot
+        Root directory of the repository.
+
+    .OUTPUTS
+        [string] Semantic version string.
+    #>
```
WilliamBerryiii, line 288: **Medium.** Membership moves from an explicitly enumerated, human-reviewed recipe to automatic derivation, and the only exclusion is a license heuristic that fails open in several ways:

- `Get-SkillLicense` accepts only a zero-indent `license:` key in the first frontmatter block within the first 200 lines. A skill that declares no license, indents the key, expresses it under `metadata:`, or uses `SEE LICENSE IN <file>` returns an empty string - and `Test-NoncommercialLicense` returns `$false` for empty input, so it ships.
- The token regex matches `CC-BY-NC-4.0` but not `BUSL-1.1`, `Elastic-2.0`, `PolyForm-Shield-1.0.0`, `Prosperity-3.0.0`, `CC-BY-ND-4.0`, or `Proprietary`.
- Agents, prompts, and instructions have **no license or eligibility gate at all**.

The retired `x-hve.componentMaturity` map allowed any artifact to be excluded with an explicit `removed` tombstone - that is how `skills/security/owasp-docker` was kept out of the VSIX. Its replacement classifies by path shape only, so an author can no longer express "tracked but not distributable" for any artifact kind.

Because this sync now runs unattended inside the release preparation jobs and commits its result, the default outcome is silent inclusion. Suggest a fail-closed SPDX allowlist plus a `distributable: false` frontmatter opt-out honored across all four artifact kinds, with unclassified skills surfaced as `-Check` violations so inclusion is an explicit decision.

scripts/plugins/Sync-PluginManifest.ps1
```diff
@@ -0,0 +1,620 @@
...
+    [CmdletBinding()]
+    [OutputType([string[]])]
+    param(
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string]$RepoRoot,
+
+        [Parameter(Mandatory = $true)]
+        [System.Collections.IDictionary]$Manifest
```
WilliamBerryiii, line 533: **Low (naming).** `Test-PluginCatalog` and `Test-PluginComponentCoverage` are both `[OutputType([string[]])]` and return violation messages, with an empty array meaning success. The `Test-` verb signals a predicate, and this same file uses it correctly for `Test-NoncommercialLicense` (`[OutputType([bool])]`), as do the two new path guards added by this PR - so one new file now carries two opposite meanings of the same verb.

It matters at the call site: `if (Test-PluginCatalog ...)` reads as "the catalog is valid" but is true precisely when it is **invalid**, because a non-empty string array is truthy. Suggest `Get-PluginCatalogViolation` / `Get-PluginComponentCoverageViolation`.

scripts/plugins/Sync-PluginManifest.ps1
```diff
@@ -0,0 +1,620 @@
...
+    [OutputType([string[]])]
+    param(
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string]$RepoRoot,
+
+        [Parameter(Mandatory = $true)]
+        [ValidateNotNullOrEmpty()]
+        [string]$SourcePath,
+
```
WilliamBerryiii, line 483: **Low.** `$changed` is computed from a full-text comparison of the serialized manifest, but `Compare-PluginComponentSet` inspects only `agents`, `commands`, `rules`, and `skills`.

When drift is confined to `version`, `description`, `license`, `keywords`, `hooks`, or key ordering, `-Check` emits only the generic *"is out of sync with tracked components"* with no further detail - misleading when no component actually changed. The suite confirms this: the `Detects a stale version` case asserts only `Should -Match 'out of sync'`.

Suggest adding scalar-field drift reporting alongside the component comparison so the operator is told what actually differs.

.github/workflows/release-stable.yml
```diff
@@ -176,40 +177,6 @@ jobs:
             exit 1
           fi
```
WilliamBerryiii, line 178: **Medium.** This removal deletes the only gate in the stable lane that observed the **content** being promoted.

What went with it: the `plugin-release-evidence.json` download from the promoted pre-release tag, and the assertions that it matched `.schema`, `.sourceCommit`, `.version`, `.locator.repo`, `.locator.ref == "prerelease-v" + version`, positive package and file counts, and 64-hex sha256 digests. The `Verify cross-channel package-set consistency` step, which independently proved the Stable and PreRelease sets matched, is deleted in the same change.

What remains proves only tag shape, odd minor, resolvable commit, published non-draft state, manifest version match, and branch ancestry - all properties of the *tag*, none of the *content*. A pre-release whose published artifacts were incomplete, replaced, or produced by a diverging re-run now promotes to stable undetected, and the two channels can drift apart silently.

If the evidence document is genuinely retired, consider replacing it with a signed tree-digest anchor and re-asserting that binding here before promotion proceeds.

scripts/release/Assert-ReleaseAssetSet.ps1
```diff
@@ -494,12 +339,10 @@ function Assert-ReleaseAssetSet {
 if ($MyInvocation.InvocationName -ne '.') {
     try {
         [void](Assert-ReleaseAssetSet -AssetNamePath $AssetNamePath `
```
WilliamBerryiii, line 345: **Low (nit).** `-OptionalAssetPath` is indented 12 spaces while its four sibling arguments in this backtick-continued call use 16. The `Test-ReleaseAssetSet` call just above uses the correct alignment, so the file now contradicts itself and the argument reads as though it belongs to a different expression.

.github/skills/installer/hve-core-installer/scripts/component-copy.ps1
```diff
@@ -124,69 +118,52 @@ $schemaVersion = 2
...
-    $catalogRoot = $segments[0]
+    $segments = $ManifestPath -split '/', 2
+    if ($segments.Count -lt 2) { return $ManifestPath }
+    $manifestRoot = $segments[0]
     $relative = $segments[1]
     foreach ($field in $fieldMap.Keys) {
         $descriptor = $fieldMap[$field]
-        if (-not [string]::Equals($descriptor.CatalogRoot, $catalogRoot, [System.StringComparison]::Ordinal)) { continue }
+        if (-not [string]::Equals($descriptor.ManifestRoot, $manifestRoot, [System.StringComparison]::Ordinal)) { continue }
         if ($descriptor.SourceSuffix) {
```
WilliamBerryiii, line 155: **Low.** This function is rewritten by this PR (the parameter is renamed `-CatalogPath` to `-ManifestPath` and the body retargeted to the plugin manifest), so it is touched code - but it remains the only function in the file declared with a bare `param()` and no block comment-based help, `[CmdletBinding()]`, or `[OutputType()]`.

`powershell.instructions.md` lists `PSProvideCommentHelp` and `PSUseOutputTypeCorrectly` as enforced, and both are enabled in `scripts/linting/PSScriptAnalyzer.psd1` with `ExportedOnly = $false` / `BlockComment = $true` / `Placement = 'before'`. They emit at Information severity, so the repo's Error/Warning filter keeps CI green while the convention is unmet. The explanatory comment above documents intent but satisfies neither rule.

scripts/README.md
```diff
@@ -347,7 +349,6 @@ Key rules:
 * [Extension Packaging Documentation](extension/README.md)
 * [Library Utilities Documentation](lib/README.md)
 * [Linting Scripts Documentation](linting/README.md)
```
WilliamBerryiii, line 351: **Low.** This list drops `* [Plugin Generation Documentation](plugins/README.md)` while `scripts/plugins/README.md` is retained and rewritten in this same PR as "Plugin Manifest Scripts" - making it the only script subdirectory README without an entry here.

Separately, the Release Scripts intro still reads "contains the repository version-update helper used by release workflows" (singular) directly above a table this PR expands to four scripts covering version normalization, promotion-version resolution, release-asset reconciliation, and provenance verification.

WilliamBerryiii (changes requested):
Thanks for this — the one-plugin consolidation is a genuinely large simplification and it lands cleanly. Every retired subsystem has a named replacement with no dangling callers, the deleted Pester suites are replaced rather than dropped, the new tests actively assert that package identity stays gone, and the 921-line `Update-VersionFiles.ps1` giving way to a 381-line structural `Set-RepositoryVersion.ps1` is a real win on the release path. The added path-containment guards and the genuine fix to the Bash installer's traversal check are nice touches, and the ~45-page documentation sweep is unusually thorough.

Requesting changes on six items before merge.

**Blocking — the Stable release lane cannot complete.** `Sync-PluginManifest.ps1` enforces catalog-to-`package.json` version parity even in write mode (the `Test-PluginCatalog` accumulation at line 573 sits outside both the `if ($changed)` and `if ($Check)` branches, and the main block exits 1 at 586). In `release-stable.yml`, `marketplace.json` is restored from `release/stable` at line 386 while `package.json` is checked out from `$SOURCE_SHA` at lines 391-393 — then Sync runs at 424, before `Set-RepositoryVersion.ps1` at 426. Since a stable promotion by definition promotes a version that differs from the stable baseline, this aborts every time. `release-prerelease-prepare.yml` (337 / 369 / 371) hits the same gate whenever `main` is ahead of the prerelease baseline, which is the ordinary case. PR CI never runs release workflows, and every fixture in `Sync-PluginManifest.Tests.ps1` builds both versions from the same `$Version` parameter, so the 123 green checks don't cover it. Swapping the two invocations resolves it — `Set-RepositoryVersion` already writes both catalog version fields.

**Blocking — two supply-chain guards removed as the distributed surface widens.** `Assert-NoTrackedPluginOutput.ps1` (the only check rejecting git index mode `120000`) is deleted in the same change that makes `.github` the verbatim distributed plugin root, so a tracked symlink now flows into `plugin.json`, past a `Test-Path` check that follows links, and into both the VSIX and every consumer's workspace with no CI detection. Separately, `release-candidate.json` and its byte-for-byte reproduction gates are removed from `pr-validation.yml`, `release-prerelease.yml`, and `release-stable-publish.yml`, leaving published membership derived from a mutable preparation tree and pushed by a credentialed job via `git add --all` with no immutable anchor.

**Blocking — the change is under-declared.** The catalog drops nine installable package names (`coding-standards`, `data-science`, `design-thinking`, `experimental`, `hve-core-all`, `installer`, `project-planning`, `rpi`, `security`) plus the `starter` profile, removes the installer's `-PackageName` parameter and the `selection.package` tracking field, drops four npm commands, and withdraws release assets `SECURITY.md` told consumers to verify — but "Breaking change" and "Changes are backwards compatible" are both unchecked. And `docs/getting-started/package-migration.md`, the page consumers are pointed to, lists only 5 of the 13 retired identities, so users of the eight domain extensions have no actionable row on a page that itself notes the Marketplace emits no deprecation signal.

**Blocking — an always-on instruction file now contradicts itself.** `pull-request.instructions.md` deleted the `## GHCP Maturity Output` heading but kept its entire body, so three lines after "Do not infer per-artifact maturity" it still orders a maturity table — and the `componentMaturity` source that fed it was deleted by this same PR.

Also worth resolving: the undisclosed gh-aw v0.85.4 → v0.86.2 regeneration bundled into the diff (and the stated "125 changed files" vs. the actual 137); three unchecked Required Local Checks against ~40 changed docs files; "56 rules" where the manifest and the new plugin page both say 57; and the adoption-risk `[!CAUTION]` block dropped from the page that now feeds the published Marketplace README.

Full findings — 43 items across functional, standards, security, PR, readiness, and accessibility perspectives, each with file:line and a suggested fix — are in the attached review. Happy to walk through any of them.

---

### Findings that could not be anchored to the diff

These are real but sit outside any changed hunk, so GitHub cannot host them as inline comments:

- **`scripts/tests/extension/ExtensionTestFixtures.psm1:113,126`** *(Low)* — these fixtures still construct `componentMaturity` data. The file is not touched by this PR, but the PR removes the key from `marketplace.json` and every production reader, so the fixtures are left silently encoding a retired catalog contract. They pass precisely because nothing reads the key.
- **`docs/contributing/release-process.md`** *(intended anchor line 318)* — see the full review artifact; this line is outside a changed hunk.
- **`.github/workflows/extension-marketplace-publish.yml`** *(intended anchor line 60)* — see the full review artifact; this line is outside a changed hunk.

### Not defects, but worth a decision

- **PR description accuracy.** The body states 56 rules; `.github/plugin.json` and the new `docs/plugins/hve-core.md` inventory both carry **57**. It also states "125 changed files" against an actual 137 — the gap is the bundled gh-aw regeneration.
- **Validation evidence.** Three of four Required Local Checks are unchecked (`validate:local`, `validate:docs`, `spell-check`) against roughly 40 changed documentation files, and the reported pass counts were captured before the two most recent merges from `main`.
- **Mergeability.** `mergeStateStatus` is BLOCKED with 123 checks passing and `reviewDecision: REVIEW_REQUIRED`. Worth confirming the two skipped checks are not themselves required, since a required check reported as skipped still blocks.
- **No linked issue or ADR.** This retires a distribution architecture, published identities, a release asset set, and a persisted consumer tracking schema. The repository maintains an ADR convention under `docs/planning/adrs/`; right now the reasoning survives only in this PR description.

_Review generated with AI assistance against `b2b0f0c7`, then machine-verified: every High and Medium finding was re-checked against the worktree, which corrected two claims and surfaced two more. Findings are suggestions for a human reviewer to weigh, not approval or sign-off._

docs/getting-started/package-migration.md
rezatnoMsirhC: Correction: I conflated the newly generated manifest identity with the extension identity that was actually published. The Marketplace listing is `ise-hve-essentials.hve-core-all`, while `ise-hve-essentials.hve-hve-core-all` does not exist. The cited tests describe generator behavior introduced after the published extension, so they do not establish a retired identity that users need to remove. I am retracting this finding.

jkim323:
I may have identified a potential upstream Copilot CLI issue: on Windows, the installer may race a running host’s skill watcher. When VS Code’s embedded Copilot host observes `skills/<name>/SKILL.md` in the CLI staging directory, it can hold a directory handle that prevents the installer from renaming the staging directory into its final location, resulting in `Access is denied. (os error 5)`. The same locking behavior can also prevent `copilot plugin uninstall`.

This is not yet confirmed as a general CLI issue and may depend on host version, timing, or local environment. However, because this PR consolidates HVE Core into a skill-heavy plugin, what are your thoughts on adding a brief known-issue note next to the CLI update instructions advising Windows users to close VS Code and active Copilot sessions before installing, updating, or uninstalling the plugin.

scripts/release/Assert-ReleaseAssetSet.ps1
agreaves-ms: I kept current reconciliation scoped to releases produced by the one-plugin lanes. Historical multi-plugin releases remain inspectable records but are not replayed by the current recovery path.

WilliamBerryiii (approved):

PR #2720, opened 2026-08-14
refactor(agents)!: consolidate supply-chain reviewers and align data-science-engineering vocabulary

### LINKED ISSUE #2733: fix(agents): SSSC Reviewer diff mode consumes a changed-files list no workflow step produces
## Summary

PR #2720 merged the deleted `Supply Chain Reviewer` assessment pipeline into `.github/agents/security/sssc-reviewer.agent.md`, but the diff-mode input-resolution stage was not carried over. Step 3.1 instructs the reviewer to pass a "changed files list" to `Supply Chain Skill Assessor`, and Step 4.3 reasons about diff-scoped findings, yet no step in the Required Workflow ever produces that list. Step 1.3 says only "Resolve the target scope for the selected mode."

`diff` is one of the reviewer's three operating modes and it is now undefined at its entry point.

## Why this matters

The deleted agent owned this stage explicitly, and it carried a rule specific to supply-chain review: after resolving the changed files and excluding binaries, it **deliberately retained** CI/CD workflow files, dependency manifests, lockfiles, SBOM documents, and signing or provenance configuration in scope, because those files carry the primary supply-chain evidence in a diff. It kept the filtered list for assessment and the unfiltered list for the report appendix.

That retention rule is what makes a diff-scoped supply-chain review meaningful. Without it, a reviewer scoped to a diff can silently skip exactly the files the review exists to examine. The failure is quiet in both directions: the mode does not error, and `docs/agents/security/sssc-reviewer.md` still advertises that diff mode "resolves the changed files ... and keeps supply-chain-relevant configuration ... in scope."

A compounding packaging gap: `.github/plugin/marketplace.json` does not list `skills/shared/pr-reference` in the `project-planning` package, so the skill this fix depends on is absent from one of the three packages that ship the reviewer. That same package also gained `agents/security/subagents/cve-analyzer.agent.md` without `skills/security/vex`, `instructions/security/vex-generation.instructions.md`, or `instructions/security/vex-standards.instructions.md` — the reference set the reviewer's VEX Assessment Capability names as canonical. `hve-core-all` and `security` both carry the full set.

## Acceptance criteria

- [ ] The Setup step in `sssc-reviewer.agent.md` resolves a changed-files list for `diff` mode, so the input consumed by Steps 3 and 4 has a defined producer.
- [ ] The supply-chain scope-retention rule is restored in wording: binaries and images are excluded, and CI/CD workflows, dependency manifests, lockfiles, SBOM documents, and signing or provenance configuration are explicitly retained.
- [ ] The filtered list is used for assessment and the unfiltered list remains available for the report's changed-files appendix.
- [ ] A fallback is stated for when `pr-reference` is unavailable, consistent with the degrade-and-continue guardrail the other three stages already follow, including the Limitations record.
- [ ] `skills/shared/pr-reference` is present in every package that ships the reviewer, or the fallback is documented as the supported path for packages that omit it.
- [ ] The `project-planning` package either gains `skills/security/vex` and the two VEX instruction files with matching `componentMaturity` entries, or drops `cve-analyzer`.
- [ ] `docs/agents/security/sssc-reviewer.md` and the agent agree on one story for how a user supplies or produces the changed-files list. Quick Start step 3 currently requires a pull request reference while Operating Modes and Prerequisites no longer name the mechanism that produces one.
- [ ] `npm run plugin:validate` passes after the package membership changes.
- [ ] Running the reviewer in `diff` mode against a change that touches only a lockfile and a CI workflow produces a diff-scoped report rather than falling back to a full audit.

## Suggested direction

Restore the stage in Setup rather than inside Step 3, so both `diff` and `plan` resolve their inputs in the same place:

````markdown
3. Resolve the target scope for the selected mode.
   * For `diff`, generate a PR reference using the `pr-reference` skill and resolve the
     changed files list. Exclude binary and image files. Retain supply-chain-relevant
     configuration in scope (CI/CD workflow files, dependency manifests, lockfiles, SBOM
     documents, and signing or provenance configuration), since these carry the primary
     supply-chain evidence. Keep the filtered list for assessment and retain the unfiltered
     list for the report's changed-files appendix. When `pr-reference` is unavailable,
     resolve the changed files directly from the repository and note the reduced rigor in
     Limitations.
   * For `plan`, resolve the plan document from the explicit path or the available context.
     If no plan document can be resolved, ask the user for the path and wait.
````

## Related

* Introduced by #2720.
* Origin of the deleted agent: #2070, #2073.
* Reviewer documentation added in #2198.

## Notes

Found during an AI-assisted code review of #2720. Findings were reviewed by a human before this issue was filed.

### PR DESCRIPTION
**Pull request by WilliamBerryiii**
# Summary

Consolidates an overlapping corner of the agent catalog and aligns the data-science package vocabulary. Three related changes travel together; they share no files and are separable by commit.

1. **Retire the duplicate supply-chain reviewer.** `Supply Chain Reviewer` and `SSSC Reviewer` both assessed the single `supply-chain-security` skill across the same audit, diff, and plan modes. `SSSC Reviewer` is the load-bearing one — two prompts, the `vex-draft` workflow, an instructions `applyTo` glob, the `vex` skill, a `vex-detect` user message, the documented security model, and a PowerShell eval fixture all bind to it. `Supply Chain Reviewer` had none of those bindings.
2. **Align the data-science vocabulary.** Package becomes `data-science-engineering`, the coach becomes the **Data Science and Engineering Coach**, and five skills drop the `ds-` prefix.
3. **Promote privacy, Responsible AI, and telemetry to cross-cutting concerns** in the coach.

`Dependency Reviewer` was evaluated and deliberately left standalone.

## Why the duplicate wasn&#39;t just deleted

`Supply Chain Reviewer` was the *only* agent that actually dispatched the assessment pipeline. `SSSC Reviewer` declared four subagents in frontmatter but assessed inline and never called them. A plain delete would have discarded adversarial verification of FAIL/PARTIAL findings and orphaned `supply-chain-skill-assessor.agent.md`, whose sole dispatcher was the file being removed.

So the pipeline was **ported first** (`Codebase Profiler` → `Supply Chain Skill Assessor` → `Finding Deep Verifier`), then the duplicate was removed. `Report Generator` was deliberately excluded: its report roots are hard-coded to `.copilot-tracking/security`, so adopting it would have reintroduced the exact report-root collision this consolidation removes. `SSSC Reviewer` continues authoring its own nine-section report into `.copilot-tracking/sssc-reviews/`.

## Things reviewers should know

* **`Supply Chain Reviewer` shared `Security Reviewer`&#39;s report namespace** — identical directory, filename pattern, and sequence counter. Its removal resolves that collision as a side effect.
* **The cross-cutting change is a defect fix, not a feature.** `marketplace.json` already shipped `privacy-standards`, `rai-standards`, `rai-planner`, and `telemetry-foundations` in this package, and the job registry already named three of them. But the coach&#39;s stop rules said to *&#34;stop rather than cross a seven-skill boundary&#34;* — contradicting the registry it is told to load. `rai-standards` was routed nowhere at all.
* **`project-planning` shipped `SSSC Reviewer` without the subagents it now dispatches.** Marketplace closure keys on `handoffs`, not the `agents` list, so `plugin:validate` would not have caught this. The package now carries the three subagents plus `security-reviewer-formats`.
* **The published package identity changed.** Already-installed users will see `data-science-engineering` / *HVE Core - Data Science and Engineering*.

## Deliberately preserved

| Identifier | Why |
|---|---|
| `DS_CATALOG_V1`, its schema `$id`, and its schema/template filenames | Cross-package document contract; `hve-core/architecture-diagrams` reads it, and users may have authored catalogs with it |
| Feasibility interchange schema identifier | Same reasoning |
| `data-workstream-session-v1` state schema version | Durable identifier inside live user session files |
| `.copilot-tracking/ds/` session root | Pairs with the existing `.copilot-tracking/dt/` convention; renaming would silently orphan in-flight user sessions |

## Rename history

Every rename is a committed `git mv` separate from its content rewrite. Because this repo squashes PRs, branch commits do not reach `main` and only final-tree similarity determines rename detection — so verification is a merge-base check rather than a branch-only `git log --follow`:

```text
git diff -M --find-renames --summary $(git merge-base HEAD origin/main)...HEAD
# 76 renames detected
```

## Validation

| Lane | Result |
|---|---|
| `lint:plugin-output`, `lint:tables`, `lint:md`, `lint:ps`, `lint:yaml`, `lint:json` | Pass |
| `lint:frontmatter` (897 files) | Pass, 0 errors / 0 warnings |
| `plugin:validate` (10 plugins) | Pass |
| `lint:hooks`, `lint:version-consistency` (627 pinned actions) | Pass |
| `lint:py` (16 skills), `validate:skills` (72 skills) | Pass |
| `lint:ai-artifacts`, `lint:models` | Pass |
| `lint:asset-docs` | Pass, 0 errors (215 pre-existing repo-wide stub warnings) |
| `docs:generate:check` (235 pages) | Zero drift |
| `lint:links`, `lint:md-links` | Not run locally; long-running link checks left to CI |

Two issues surfaced during validation and were fixed: an initial package `version` bump to `4.0.0` was rejected because package versions are lockstep with root `package.json`, and table formatting drifted across ten files (`format:tables` applied).

## Generated output

`docs/reference/` was regenerated with `docs:generate`; the `docs/plugins/` marker region belongs to a different generator (`plugin:generate`) and was refreshed with an external staging root. The generator preserved eight renamed reference pages as orphans because they carried authored tails — those tails were carried forward before the orphans were removed, so no hand-written &#34;When to use it&#34; content was lost.

## Review notes

* The three changes are separable by commit if you prefer to review them independently.
* An independent `hve-builder` review gate ran against the coach rewrite and returned **Revise**; all six findings were applied, including one real inconsistency where the RAI trigger differed between the agent and the registry.

## Related issues

Closes #2733. `SSSC Reviewer` diff mode now resolves its changed-files list from a defined producer, including a degraded path that stays in `diff` mode when `pr-reference` cannot be activated, the reviewer documentation matches that contract, and an isolated diff-mode stimulus locks the behavior.

🤖 Crafted with precision by ✨Copilot following brilliant human instruction, then carefully refined by our team of discerning human reviewers.


.github/plugin/marketplace.json
jkim323, line 1: **High — project-planning installs VEX-capable agents without their canonical VEX reference set.** 

This package now includes `sssc-reviewer` and `cve-analyzer`, but its rules omit `vex-generation.instructions.md` and `vex-standards.instructions.md`, and its skills omit `skills/security/vex`. The reviewer explicitly declares all three as required for VEX review; the analyzer cannot perform its required setup without at least the skill and generation instructions. 

**Suggested fix:** Please either add the complete VEX closure (including maturity entries), or remove `cve-analyzer` and the VEX capability from this package. Also add `skills/shared/pr-reference` when restoring the diff-mode setup contract.

.github/agents/security/sssc-reviewer.agent.md
```diff
@@ -85,29 +100,43 @@ Each report must also include a dedicated evidence inventory section that record
...
+5. If the request includes a subdirectory focus, restrict the audit review to that scope, pass the focus to the profiler, and note the boundary explicitly.
+6. When `Codebase Profiler` is unavailable, profile the scope directly from the repository and note the reduced rigor in the report's Limitations section. Never end a review because a subagent could not be dispatched.
 
 ### 3. Assess Supply-Chain Posture
 
-1. Evaluate the relevant posture areas, such as dependency hygiene, provenance, signing, SBOM generation, build isolation, release integrity, and repository controls.
-2. Prefer evidence from the repository itself, such as workflow files, dependency manifests, signing configuration, release automation, build outputs, and release artifacts.
-3. Classify findings as PASS, PARTIAL, or FAIL when the evidence supports a clear judgment. If evidence is insufficient, mark the item as NEEDS_REVIEW.
+1. Run `Supply Chain Skill Assessor` once per applicable skill, passing the skill name and the codebase profile. For `diff`, also pass the changed files list. For `plan`, also pass the plan document content.
```
jkim323: **High — diff mode has no producer for `changed files list`.** 

Step 3.1 requires passing this list to `Supply Chain Skill Assessor`, but setup only says “resolve the target scope” and never resolves a PR reference, enumerates changed files, or applies the supply-chain retention policy. The deleted `Supply Chain Reviewer` explicitly used `pr-reference`, filtered binaries/images, preserved CI/CD/manifests/lockfiles/SBOM/provenance files, and retained an unfiltered appendix list. 

**Suggested fix**: Please restore that setup stage and pass the resulting filtered list to the profiler/assessor while retaining the unfiltered list for reporting.

.github/agents/security/sssc-reviewer.agent.md
```diff
@@ -85,29 +100,43 @@ Each report must also include a dedicated evidence inventory section that record
 
 ### 2. Profile the Scope
 
-1. Profile the repository or plan document to identify the relevant technology stack, release surfaces, package managers, CI/CD flow, and supply-chain risk surfaces.
-2. Use the `supply-chain-security` skill as the primary reference source for posture concepts, standards links, and remediation guidance.
-3. If the request includes a subdirectory focus, restrict the audit review to that scope and note the boundary explicitly.
+1. Run `Codebase Profiler` and capture its profile output, which identifies the technology stack, release surfaces, package managers, CI/CD flow, and supply-chain risk surfaces. Pass the full profile text verbatim to the downstream subagents.
+2. Intersect the profiler's applicable-skill list with the Available Skills list. When no skill remains, state that and assess against `supply-chain-security` directly rather than ending the review.
+3. When a single target skill is supplied, skip profiling, validate the skill against the Available Skills list, and build a minimal profile stub instead.
```
jkim323: **High — SSSC Reviewer exposes a target-skill fast path that its documented input contract cannot receive.** 

Step 2.3 branches when a single `target skill` is supplied, but Inputs does not declare targetSkill (or an equivalent target-skill parameter), so callers have no supported way to invoke the profiling bypass. The branch also requires a Minimal Profile Stub, while Format Specifications no longer references `security-reviewer-formats/references/completion-formats.md`, where that required format is defined. 

**Suggested fix:** Please either restore the optional target-skill input and the Completion Formats reference, or remove Step 2.3 if this optimization is intentionally out of scope for the consolidated reviewer.

.github/agents/data-science-engineering/data-science-engineering-coach.agent.md
```diff
@@ -1,36 +1,41 @@
 ---
-name: Data Workstream Coach
-description: "Coach a persistent data-science and data-engineering workstream through explicit jobs, durable state, routed skill authority, and safe customer-artifact writes."
+name: Data Science and Engineering Coach
+description: "Coach a persistent data science and data engineering workstream through explicit jobs, durable state, routed skill authority, and safe customer-artifact writes."
```
jkim323, line 3: **High — the agent rename leaves five lifecycle-guide picker instructions pointing to a non-existent agent.** 

The agent is now named Data Science and Engineering Coach, but Discovery, Implementation, and Review still tell readers to select Data Workstream Coach. These are hand-authored lifecycle guides, so `npm run docs:generate` will not repair them; users following these workflows will search the agent picker for the old name and fail to find it. 

**Suggested fix**: Please replace both the display-name and “Select … agent” references at `discovery.md:37`, `implementation.md:35,122`, and `review.md:35,102`.

.github/skills/hve-core/architecture-diagrams/SKILL.md
```diff
@@ -16,7 +16,7 @@ metadata:
 
 Turn infrastructure source files or a declared `DS_CATALOG_V1` data model into a readable architecture diagram for reviews, ADRs, and design discussions. Preserve the caller's selected output format and the source's authority boundaries.
 
-Infrastructure inputs include Terraform, Bicep, ARM templates, shell scripts, Kubernetes manifests, and Docker or Compose files. Catalog input uses declared entities and relationships from `ds-catalog`, the durable data-catalog skill. It does not infer a data model from SQL or ORM files.
+Infrastructure inputs include Terraform, Bicep, ARM templates, shell scripts, Kubernetes manifests, and Docker or Compose files. Catalog input uses declared entities and relationships from `data-catalog`, the durable data-catalog skill. It does not infer a data model from SQL or ORM files.
```
jkim323, line 19: **High — renamed skill and package identifiers remain in preserved reference-page prose.** 

This change correctly updates the source skills to `data-catalog` and `dataops`, but the hand-authored tails in `docs/reference/skills/hve-core/architecture-diagrams.md:29` and `docs/reference/skills/project-planning/experiment-design.md:32,37` still direct readers to `ds-catalog`, `ds-dataops`, and the removed `data-science package`. Those sections sit outside generated regions, so regeneration will preserve the stale names. 

**Suggested fix:** Please update the published reference pages to `data-catalog`, `dataops`, and `data-science-engineering` so the documentation remains consistent with the renamed skills and package.

.github/skills/data-science-engineering/data-science-engineering-foundation/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
-name: data-workstream-foundation
-description: "State, resume, reconstruction, job-lifecycle, transition, and flow-state mechanics for the Data Workstream Coach. Loaded by the coach; not a user entry point."
+name: data-science-engineering-foundation
```
jkim323, line 2: **High — the renamed foundation skill still introduces itself with the retired Data Workstream name.** 

This change correctly renames the frontmatter to `data-science-engineering-foundation` and updates its description for the Data Science and Engineering Coach, but the H1 still reads `Data Workstream Foundation`. That makes the artifact’s visible identity disagree with its registered skill name and leaves the deprecated vocabulary as the first heading readers see. 

**Suggested fix:** Please rename the H1 to `Data Science and Engineering Foundation` and sweep the sibling foundation references for the same remaining “Data Workstream” terminology.

.github/instructions/shared/disclaimer-language.instructions.md
```diff
@@ -62,7 +62,7 @@ Authoring contract (parsed by scripts/linting/Validate-PlannerArtifacts.ps1):
 > [!CAUTION]
 > **Disclaimer:** This agent is an assistive coaching tool only. It does not conduct user research, observe stakeholders, or speak for the people whose problems you are designing for, and it does not replace primary research, direct stakeholder contact, design review, or product and strategy decision authority. Personas, problem statements, journey maps, empathy maps, concept tests, and other Design Thinking artifacts produced with this tool are scaffolding for your own research and synthesis — not substitutes for real stakeholder voice or observed behavior. Validate all AI-generated assumptions, personas, themes, and insights against actual stakeholders before treating any Design Thinking artifact as a basis for product, design, or strategy commitments. Outputs from this tool do not constitute validated research findings or design approval.
 
-## Data-Science Coaching
+## Data Science and Engineering Coaching
```
jkim323, line 71: This rename changes the identifier this file's own authoring contract derives. `Import-DisclaimerSource` in Validate-PlannerArtifacts.ps1 takes the first whitespace-delimited heading word as the slug, so `Data-Science Coaching` derived `data-science-planner` / `data-science-full-disclaimer`, while `Data Science and Engineering Coaching` derives generic `data-planner` / `data-full-disclaimer`.

CI stays green because footer-with-review.yml consumes only `rai-full-disclaimer` and `sssc-full-disclaimer`, but the identifier no longer aligns with the `data-science-engineering` package and can silently collide with any future H2 disclaimer section starting with "Data."

That collision mode is already live: `## Privacy Planning` and `## Privacy Review` both derive `privacy`, and the hashtable assignment overwrites without error — so the Privacy Planning text is currently unreachable from parsed config, and anything referencing `privacy-full-disclaimer` would validate against Privacy Review prose.

Smallest fix here is a stable package-aligned first token:

```markdown
## Data-Science-Engineering Coaching
 ```

A duplicate-key guard (or explicit ids, as `footer-with-review.yml` already uses for footers) is worth doing, but it has to resolve the Privacy collision in the same change - I suggest to consider filing that separately rather than grow this PR.

**Review by jkim323 — CHANGES_REQUESTED** [review]
Left a couple things for you to look at. Happy to review back - thanks for working on this consolidation effort!

File: .github/agents/security/sssc-reviewer.agent.md [hunk 7]
```diff
@@ -85,29 +100,43 @@ Each report must also include a dedicated evidence inventory section that record
...
+6. When `Codebase Profiler` is unavailable, profile the scope directly from the repository and note the reduced rigor in the report's Limitations section. Never end a review because a subagent could not be dispatched.
 
 ### 3. Assess Supply-Chain Posture
 
-1. Evaluate the relevant posture areas, such as dependency hygiene, provenance, signing, SBOM generation, build isolation, release integrity, and repository controls.
-2. Prefer evidence from the repository itself, such as workflow files, dependency manifests, signing configuration, release automation, build outputs, and release artifacts.
-3. Classify findings as PASS, PARTIAL, or FAIL when the evidence supports a clear judgment. If evidence is insufficient, mark the item as NEEDS_REVIEW.
+1. Run `Supply Chain Skill Assessor` once per applicable skill, passing the skill name and the codebase profile. For `diff`, also pass the changed files list. For `plan`, also pass the plan document content.
+2. Collect the structured findings from each successful assessment. Exclude any skill that fails after the retry protocol and record the reason.
```
katriendg: **High — this step cites a "retry protocol" the consolidated agent never defines.**

`retry protocol` appears exactly once in this file and is defined nowhere in it. It came from the deleted `Supply Chain Reviewer`'s `## Required Protocol`:

> 5. After each subagent invocation, handle clarifying questions before proceeding.
> 6. If a subagent response is incomplete or malformed, retry once. If it still fails, exclude that skill from subsequent steps and record the reason.

That section was not carried over — `SSSC Reviewer` has no `Required Protocol` at all. So the absorbed pipeline has a defined fallback for a subagent that is *unavailable* (2.6, 3.5, 4.7, plus the new Guardrail bullet) but no defined behavior for one that returns *malformed or incomplete* output, and it points the reader at a procedure they cannot resolve. Clarifying-question handling was lost the same way.

Inlining the rule is the smaller fix:

```suggestion
2. Collect the structured findings from each successful assessment. Exclude any skill that returns malformed or incomplete output twice and record the reason.
```

The fuller fix is restoring a `## Required Protocol` section carrying retry-once-then-exclude *and* clarifying-question handling, since the latter has no replacement anywhere in the file.

.github/agents/security/sssc-reviewer.agent.md
```diff
@@ -85,29 +100,43 @@ Each report must also include a dedicated evidence inventory section that record
...
 
-1. Write the report to the resolved path in the `sssc-reviews` directory.
-2. Include the mode, scope, findings, evidence, remediation guidance, limitations, and recommended follow-up actions.
-3. End with a concise completion summary that lists the report path and the highest-priority next steps.
-4. Follow hve-core Markdown, writing-style, and licensing-posture conventions for generated reports. Paraphrase standards guidance and cite or reference the canonical skill rather than reproducing large standards tables or extended source text.
+1. Author the report directly. Do not delegate to `Report Generator`.
+2. Translate the subagent vocabulary into this reviewer's report contract before writing:
+   * `PASS`, `PARTIAL`, and `FAIL` carry through unchanged.
+   * `NOT_ASSESSED` becomes `NEEDS_REVIEW`.
+   * Each verified finding records its verification verdict (`CONFIRMED`, `DOWNGRADED`, or `UNCHANGED`), its verified status, and its verified severity alongside the finding in the Findings section.
```
katriendg: **Medium — this verdict list admits a value that cannot occur and omits one that can.**

`finding-deep-verifier.agent.md:36` declares `Verdict values: CONFIRMED, DISPROVED, DOWNGRADED`. A **verified** finding can therefore never be `UNCHANGED`, and `DISPROVED` is missing here even though the very next bullet handles disproved findings.

`UNCHANGED` is the *pass-through* verdict — `finding-formats.md:27`: "UNCHANGED items (PASS and NOT_ASSESSED) use the Finding Serialization Format with an added `- **Verdict:** UNCHANGED` field." But step 4.2 was trimmed from the deleted agent's "Pass through PASS and NOT_ASSESSED findings unchanged **with verdict `UNCHANGED`**" down to just "Pass PASS and NOT_ASSESSED findings through unchanged."

Net: nothing in the file now assigns `UNCHANGED` to anything that can hold it, and the one place it appears is a place it cannot occur. `NOT_VERIFIED` (line 126) also has no row here, so the report contract never says how it renders.

```suggestion
   * Each verified finding records its verification verdict (`CONFIRMED`, `DOWNGRADED`, or `DISPROVED`), its verified status, and its verified severity alongside the finding in the Findings section.
   * `NOT_VERIFIED` findings render with their original status and severity and are labelled unverified in the Findings section.
```

Line 121 needs the companion change to give `UNCHANGED` its home back. It is outside this comment's range, so not applicable as a suggestion here:

```markdown
2. Pass PASS and NOT_ASSESSED findings through unchanged with verdict `UNCHANGED`.
```

.github/agents/security/sssc-reviewer.agent.md
```diff
@@ -85,29 +100,43 @@ Each report must also include a dedicated evidence inventory section that record
...
-3. Classify findings as PASS, PARTIAL, or FAIL when the evidence supports a clear judgment. If evidence is insufficient, mark the item as NEEDS_REVIEW.
+1. Run `Supply Chain Skill Assessor` once per applicable skill, passing the skill name and the codebase profile. For `diff`, also pass the changed files list. For `plan`, also pass the plan document content.
+2. Collect the structured findings from each successful assessment. Exclude any skill that fails after the retry protocol and record the reason.
+3. The assessor evaluates the relevant posture areas, such as dependency hygiene, provenance, signing, SBOM generation, build isolation, release integrity, and repository controls, preferring evidence from the repository itself.
 4. Record severity and priority separately for each finding. Severity describes the practical impact or risk level. Priority describes the order in which remediation should be handled when a recommendation is made.
+5. When `Supply Chain Skill Assessor` is unavailable, perform the same assessment inline against the `supply-chain-security` skill, using the same statuses and severities, and note the reduced rigor in Limitations.
```
katriendg: **Medium — this fallback depends on a status set the PR deleted three lines up.**

The bullet says the inline fallback assesses "using the same statuses and severities", but this file's only status-classification rule was removed in the same change:

> ~~3. Classify findings as PASS, PARTIAL, or FAIL when the evidence supports a clear judgment. If evidence is insufficient, mark the item as NEEDS_REVIEW.~~

The status set is now owned solely by `Supply Chain Skill Assessor` — the subagent this fallback exists to replace. `## Format Specifications` declares `references/severity-definitions.md` for severities but declares no status source. It stays recoverable from the Step 5.2 translation table, so this is indirect rather than broken.

```suggestion
5. When `Supply Chain Skill Assessor` is unavailable, perform the same assessment inline against the `supply-chain-security` skill, using the assessor status set PASS, PARTIAL, FAIL, and NOT_ASSESSED and the severities in `references/severity-definitions.md`, and note the reduced rigor in Limitations.
```

.github/agents/security/sssc-reviewer.agent.md
```diff
@@ -157,6 +195,8 @@ Use the following compact skeleton when validating or iterating on the report co
 
 ## Findings
 
+### Disproved Findings
```
katriendg, line 211: **Low — the consolidation removed the only artifact emitting the canonical SSSC disclaimer verbatim.**

`disclaimer-language.instructions.md` scopes `**/.copilot-tracking/sssc-reviews/**` in its `applyTo`, and the deleted `Supply Chain Reviewer` routed it correctly:

> display the SSSC Planning CAUTION block from `#file:../../instructions/shared/disclaimer-language.instructions.md` **verbatim**

`SSSC Reviewer` has no `#file:` reference to that instruction anywhere. Output Contract item 1 (line 80) asks only for "a professional-review disclaimer near the top", and this skeleton hard-codes a substitute at lines 183-184 — `> [!IMPORTANT]` with paraphrased prose rather than the canonical `> [!CAUTION]` block. After this PR that divergent copy is the only disclaimer the reviewer emits.

Rated Low because the skeleton text predates this PR — but the PR removed the compliant emitter, and the skeleton was edited here (`### Disproved Findings` added), so it is in scope.

The fix lands at lines 183-184, outside this comment's range, so no applicable suggestion. Replace:

```markdown
> [!IMPORTANT]
> This review is an assistive assessment for human review only. It is not a substitute for qualified human validation.
```

with a pointer, and change Output Contract item 1 to read "the SSSC Planning CAUTION block from `#file:../../instructions/shared/disclaimer-language.instructions.md` verbatim".

.github/agents/data-science-engineering/data-science-engineering-coach.agent.md
```diff
@@ -95,21 +100,47 @@ lifecycle class and output. Do not begin work until the user confirms one.
...
+  and PII-safe telemetry conventions.
+
+A cross-cutting concern always surfaces an observation and offers concrete
+choices. It never blocks a durable write, never edits an artifact silently, and
+never decides for the user. The durable-write scan gate remains the only
+condition that blocks a write.
+
+When a concern triggers, name it, say why it applies, name its owning skill,
+offer the contribution, and let the user choose. When the owning skill is
+unavailable, say so and continue without improvising its authority.
```
katriendg: **Medium — the coach reproduces the rules it just said the registry owns.**

Line 121 names the registry as owner — "The foundation job registry **owns** their trigger conditions, owning skills, and contributions; read its cross-cutting concerns section before offering or performing job work" — and then these two paragraphs reproduce that registry section instead of deferring to it. Lines 135-138 are word-for-word identical to `job-registry.md`, and the four owning-skill descriptions above are restated from its `## Skill authority boundaries` table.

The duplicate has already drifted: the registry reads "When the owning skill is unavailable **in the installed collection**, say so"; the copy at line 140 dropped "in the installed collection" — the qualifier that makes the rule actionable.

This also conflicts with the coach's own success criterion ("this agent does not copy its schemas or rule tables") and the foundation's ("Each orchestration rule has one owner in this package").

Deleting both duplicated paragraphs leaves the pointer sentence and the four skill names carrying the section:

```suggestion
```

evals/agent-behavior/README.md
```diff
@@ -248,41 +248,40 @@ For an entirely new agent, also re-run [Build-AgentInventory.ps1](../../scripts/
...
+| sssc-reviewer                  | code-reviewer    | light     | [.github/agents/security/sssc-reviewer.agent.md](../../.github/agents/security/sssc-reviewer.agent.md)                                                                   |
+| system-architecture-reviewer   | research-writer  | light     | [.github/agents/project-planning/system-architecture-reviewer.agent.md](../../.github/agents/project-planning/system-architecture-reviewer.agent.md)                     |
+| ux-ui-designer                 | research-writer  | light     | [.github/agents/project-planning/ux-ui-designer.agent.md](../../.github/agents/project-planning/ux-ui-designer.agent.md)                                                 |
+
+The maintained stimulus inventory totals 57 agents: 35 parent agents plus 22 enrolled subagents whose stimulus partials exist in [stimuli/](stimuli/). Subagents without a matching stimulus partial remain excluded from the matrix run set and are documented separately in the inventory generator and related eval research. [AGENTS.yml](AGENTS.yml) remains generator-owned and is refreshed in the generation phase.
```
katriendg: **Medium — this count was decremented but line 14 was not, so the file contradicts itself.**

This line now reads "totals **57** agents: **35** parent agents plus 22 enrolled subagents", while line 14 still reads "the maintained stimulus inventory contains **58** enrolled agents, including 22 subagents." Same file, same fact, two values. Line 14 sits outside the diff, which is why it was easy to miss — and also why there is no applicable suggestion here.

Line 14 needs:

```markdown
The suite is organized around four behavioral classes (research-writer, code-reviewer, workitem-manager, planner-coach). Every parent agent belongs to exactly one class, and class membership selects the stimulus shape and grader template used in [stimuli/](stimuli/). The parent-agent table below is the authoritative class assignment; the maintained stimulus inventory contains 57 enrolled agents, including 22 subagents.
```

Out of scope for this PR, but worth a follow-up: both figures were already off by one beforehand. At the merge base `AGENTS.yml` carried 57 slugs and `stimuli/` held 57 files while the README claimed 58; at HEAD they are 56 and 56 while it claims 57. The decrement here is correct relative to the removed agent — the baseline it decrements from is what is wrong.

docs/reference/agents/data-science-engineering/data-science-engineering-coach.md
```diff
@@ -1,29 +1,29 @@
...
 <!-- END AUTO-GENERATED: metadata -->
 
 ## What it does
 
 <!-- BEGIN AUTO-GENERATED: overview -->
-Coach a persistent data-science and data-engineering workstream through explicit jobs, durable state, routed skill authority, and safe customer-artifact writes.
+Coach a persistent data science and data engineering workstream through explicit jobs, durable state, routed skill authority, and safe customer-artifact writes.
 <!-- END AUTO-GENERATED: overview -->
 
 ## When to use it
```
katriendg, line 29: **Medium — the renamed package's own reference pages still send readers to the retired names.**

Anchoring here because everything below this heading is a preserved authored tail that GitHub's diff does not show — the rename only touched the generated regions above, so none of the stale text below is visible in this PR view and none of it is suggestable.

Same class as the `architecture-diagrams` / `experiment-design` comment, but it also lands on the `data-science-engineering` pages themselves. This page is the sharpest case: line 43 of this file tells the reader

> 1. Select **Data Workstream Coach** from the agent picker.

which is a name that no longer exists in the picker — on the renamed agent's own reference page, three lines after the metadata table correctly says `Data Science and Engineering Coach`.

Because these sit in preserved tails, regeneration will not repair them. `docs:generate:check` reports zero drift across 236 pages, so CI stays green on every one:

| File | Lines | Stale token |
|---|---|---|
| this file | 31, 43 | Data Workstream Coach |
| `docs/reference/skills/data-science-engineering/data-catalog.md` | 31, 35, 41 | `ds-catalog`, `ds-dataops`, Data Workstream Coach |
| `docs/reference/skills/data-science-engineering/analysis-authoring.md` | 33, 35 | `ds-catalog`, `ds-dataops`, `ds-evaluation-design` |
| `docs/reference/skills/data-science-engineering/data-science-engineering-foundation.md` | 31, 36, 37 | Data Workstream Coach, `ds-catalog`, `ds-dataops` |
| `docs/reference/skills/data-science-engineering/ml-experimentation.md` | 41, 42 | `ds-dataops` |
| `docs/reference/skills/data-science-engineering/feasibility.md` | 31 | `ds-feasibility` |
| `docs/reference/skills/data-science-engineering/evaluation-design.md` | 35 | Data Workstream Coach |

Suggested sweep for this file's line 43:

```markdown
1. Select **Data Science and Engineering Coach** from the agent picker.
```

katriendg (approved):
Went through this one in depth — agents, skills, docs, evals, and the generated surfaces — and ran the local validation lanes against the branch.

**Approving.** Both consolidations are architecturally right. Folding the supply-chain pipeline into `SSSC Reviewer` removes a genuinely duplicated orchestrator and keeps report authorship local instead of routing through `Report Generator`'s incompatible template. Hoisting privacy, RAI, and telemetry into trigger-based cross-cutting concerns is a better model than the per-job table it replaces. Nothing below asks you to rethink either one — they are absorption-hygiene and rename-sweep gaps, all local text edits. Approving rather than blocking because you reliably close findings before merging.

I left seven inline comments — one High, five Medium, one Low.

Nice piece of work — the `git mv`-separate-from-rewrite discipline is what made 112 files reviewable.

*Reviewed with AI assistance; every finding was independently verified against the diff and the referenced files before filing. Treat it as input to human review, not a substitute for it.*

.github/plugin/marketplace.json
WilliamBerryiii: Thanks for catching this. The concern was valid at the reviewed commit, but it is now resolved by main rather than by this branch.

Main's single-plugin consolidation (#2712, and the later move to a repository-root manifest) removed per-package membership entirely. There is no `project-planning` package recipe left to amend. The flat manifest already carries the complete closure you asked for: the `vex` skill, both VEX instruction files, `cve-analyzer`, `sssc-reviewer`, and `skills/shared/pr-reference`.

So this remediation deliberately makes no manifest membership change. `npm run plugin:sync` and `npm run plugin:validate` both report the manifest and catalog in sync on the current head.

.github/agents/security/sssc-reviewer.agent.md
WilliamBerryiii: Fixed. This was a genuine port gap, and the setup contract is now restored.

Diff mode uses supplied PR-reference evidence or activates the `pr-reference` skill to generate it, then derives an unfiltered changed-files inventory. A separate filtered assessment list excludes binary and image files, and the supply-chain retention rule is applied after exclusions so CI/CD workflows, dependency manifests, lockfiles, SBOM documents, and signing or provenance configuration all remain assessable. The filtered list is what reaches `Codebase Profiler` and `Supply Chain Skill Assessor`; the unfiltered list is reserved for the report's artifact inventory and evidence appendix.

Verified with a deterministic source-contract checklist covering the producer, binary/image exclusion, each retention class, both downstream handoffs, and the reporting reservation. A contained behavior run also exercised both the supplied-evidence and generated-evidence paths.

.github/agents/security/sssc-reviewer.agent.md
WilliamBerryiii: Fixed by taking your first option and keeping the capability.

Inputs now declares the optional target skill, the workflow validates it against the Available Skills list before bypassing profiling, and Format Specifications names Completion Formats (`references/completion-formats.md`) as the owner of the Minimal Profile Stub. An unrecognized target stops and reports the supported skills rather than proceeding with an unvalidated bypass.

Verified by the same deterministic contract checklist and by a contained behavior scenario that supplies a valid target skill and confirms the profiler bypass plus stub emission.

.github/agents/data-science-engineering/data-science-engineering-coach.agent.md
WilliamBerryiii: Fixed at all five cited sites.

`discovery.md`, `implementation.md`, and `review.md` now name **Data Science and Engineering Coach** in both the tool tables and the "Select ... agent" instructions, so the picker text matches the registered agent name. `ms.date` was refreshed on each page and the tables were reformatted through `npm run format:tables`.

Markdown lint, frontmatter validation, and the table format check all pass on the current head.

.github/skills/hve-core/architecture-diagrams/SKILL.md
WilliamBerryiii: Fixed, and the sweep found more than the cited sites.

`architecture-diagrams.md` now points to `data-catalog`, and `experiment-design.md` now points to `dataops` and describes the `data-science-engineering` collection. An exhaustive sweep of the authored tails also caught `analysis-authoring.md`, `data-catalog.md`, `evaluation-design.md`, `feasibility.md`, and `ml-experimentation.md`, which still referenced `ds-catalog`, `ds-dataops`, `ds-feasibility`, `ds-evaluation-design`, and the old coach name.

One narrow correction to the comment: at `experiment-design.md:32` the line is the minimum-sufficient scoping sentence, not stale terminology, so it is intentionally unchanged. The stale identifiers were on the adjacent lines.

A repository-wide search for the retired vocabulary now returns zero active matches, while the preserved protocol identifiers (`DS_CATALOG_V1`, the `ds-catalog-v1` schema and template filenames, and `data-workstream-session-v1`) are unchanged by design.

.github/skills/data-science-engineering/data-science-engineering-foundation/SKILL.md
WilliamBerryiii: Fixed, including the sibling sweep you suggested.

The root H1 is now `# Data Science and Engineering Foundation`. The sibling references were also updated: `job-registry.md`, `lifecycle-classes.md`, `flow-state.md`, `transition-protocol.md`, and `session-state.md` now use Data Science and Engineering in their frontmatter titles, descriptions, and H1s.

The `schema_version: data-workstream-session-v1` value inside `session-state.md` is deliberately unchanged, since it is a persisted session-state contract rather than display terminology. Skill structure validation and frontmatter validation both pass.

.github/instructions/shared/disclaimer-language.instructions.md
WilliamBerryiii: Great analysis, and thank you for tracing the derivation precisely. Your read of the parser was correct: the rename did change the derived identifier from `data-science-*` to generic `data-*`.

Rather than encode machine identity in display text, the fix decouples the two. The readable `## Data Science and Engineering Coaching` heading stays, and the section now carries an explicit marker:

```markdown
<!-- disclaimer-id: data-science-engineering -->
```

`Import-DisclaimerSource` now accepts exactly one canonical `<!-- disclaimer-id: lower-kebab -->` marker placed immediately before the section's CAUTION block. Explicit metadata takes precedence, and the existing first-token derivation applies only when no marker is present, so every current section (including the configured `rai-full-disclaimer` and `sssc-full-disclaimer` references) keeps its present identifier. Empty, non-kebab, misplaced, repeated, and noncanonical-whitespace markers are all rejected deterministically rather than silently falling back.

Focused Pester coverage was added for explicit-ID derivation, unchanged RAI/SSSC fallback, and each rejection case; the suite passes 50 tests.

On the Privacy collision: agreed, and taken as you suggested. `Privacy Planning` and `Privacy Review` still both derive `privacy`, and resolving that (with either explicit IDs or a duplicate-key guard) is tracked as separate follow-up work rather than expanded into this PR.

jkim323 (approved):

.github/agents/security/sssc-reviewer.agent.md
WilliamBerryiii: Good catch — this was a real dangling reference, and you're right that clarifying-question handling was lost in the same deletion. Took the fuller fix rather than the inline one.

Restored a `## Required Protocol` section to `sssc-reviewer.agent.md` carrying both halves of what the deleted agent owned:

1. Ordered workflow execution for the resolved mode.
2. Clarifying questions resolved before progression, explicitly called out as **a resolution step, not a retry**, so it doesn't consume the retry budget.
3. Retry-once on an incomplete or non-conforming response, then log, exclude, and note the exclusion in Limitations.

Step 3.2 now reads "...fails after the retry protocol **in Required Protocol**..." so the pointer resolves.

Two adjustments beyond your suggestion, both surfaced by behavior-simulating the result:

- Rule 3 carries a `Finding Deep Verifier` carve-out. Applied unmodified, "exclude that skill's findings" would have dropped the verifier's PASS and NOT_ASSESSED pass-throughs, contradicting step 4.2 and the guardrail that a report is always produced. It now excludes only FAIL and PARTIAL there.
- The malformed predicate reads "does not match its contract in Format Specifications **or in the subagent's own response-format section**". Format Specifications carries no response contract for `Supply Chain Skill Assessor` or `Codebase Profiler` — those live in the subagents' own files — so the predicate was undecidable at the cited location.

Also disambiguated the two fallback triggers: steps 3.5 and 4.7 now say "cannot be dispatched" rather than "is unavailable", so the dispatch-failure branch no longer overlaps the twice-malformed branch.

Fixed in `6fbd73ec`.

.github/agents/security/sssc-reviewer.agent.md
WilliamBerryiii: Agreed on the diagnosis — `UNCHANGED` had no home and `NOT_VERIFIED` had no rendering rule. Both fixed. I deviated on `DISPROVED`, and want to be explicit about why.

Applied:

- Step 4.2 → "Pass PASS and NOT_ASSESSED findings through unchanged **with verification verdict `UNCHANGED`**." `UNCHANGED` now lands where `finding-formats.md:27` says it belongs.
- Added the `NOT_VERIFIED` rendering bullet you asked for: renders with its assessed status and severity, labelled unverified, accounted for in Limitations.
- Added an `UNCHANGED` bullet: records the verdict without a separate verified status or severity, since the assessed values carry through.

Deviation — the verified bullet reads `CONFIRMED` or `DOWNGRADED`, not all three:

Adding `DISPROVED` there makes every disproved finding match **two** bullets with conflicting placement. The verified bullet says "alongside the finding **in the Findings section**"; the next bullet routes `DISPROVED` to "a distinct subsection of Findings". Resolving that requires the reader to treat one bullet's parenthetical as an implicit exclusion list. I scoped the bullet instead so each finding matches exactly one rendering rule, and kept `DISPROVED` handled solely by its dedicated bullet.

Scoping it that way did leave a gap you'd have hit later, so the `DISPROVED` bullet now names its field set explicitly: assessed status and severity, the `DISPROVED` verdict, and the verifier's justification.

One correction to the shared-vocabulary claim while in here. Format Specifications now attributes each vocabulary to what actually defines it: `finding-formats.md` owns the statuses and `UNCHANGED`; `Finding Deep Verifier` enumerates `CONFIRMED`, `DOWNGRADED`, and `DISPROVED`; and `NOT_VERIFIED` is fenced as reviewer-local, since it appears nowhere in the shared contract.

Fixed in `6fbd73ec`. The `sssc-reviewer` behavior gate moved 73.3% → 86.7% against its 70% threshold, with the stimulus and grader intents unchanged.

.github/agents/security/sssc-reviewer.agent.md
WilliamBerryiii: Right — the fallback was leaning on a status set the same change deleted, and the only remaining owner was the subagent the fallback exists to replace. Applied essentially as suggested.

Step 3.5 now reads:

> When `Supply Chain Skill Assessor` **cannot be dispatched**, perform the same assessment inline against the `supply-chain-security` skill, using the `PASS`, `PARTIAL`, `FAIL`, and `NOT_ASSESSED` statuses and the severity levels defined by the `security-reviewer-formats` references named in Format Specifications, and note the reduced rigor in Limitations.

Two deltas from your text:

- Severities point at Format Specifications rather than naming `references/severity-definitions.md` inline, so the file keeps one place that resolves reference paths.
- "cannot be dispatched" instead of "is unavailable". Behavior-simulating this surfaced an ambiguity: a twice-malformed assessor could plausibly claim either this inline-fallback branch or the new exclude branch in Required Protocol. The wording now matches step 2.6 and separates dispatch failure from a bad response.

Your point about the missing status source is addressed directly too — Format Specifications previously declared a severity source and no status source. It now names `references/finding-formats.md` as the authority for `PASS`, `PARTIAL`, `FAIL`, and `NOT_ASSESSED`, so the set is resolvable without reverse-engineering it from the Step 5.2 translation table.

Fixed in `6fbd73ec`.

.github/agents/security/sssc-reviewer.agent.md
WilliamBerryiii: Confirmed and fixed. The consolidation did leave the paraphrase as the only disclaimer the reviewer emitted, and that copy had drifted from the canonical block.

Applied both halves:

- Output Contract item 1 now reads "...and the SSSC Planning CAUTION block from `#file:../../instructions/shared/disclaimer-language.instructions.md` reproduced verbatim near the top under a distinct **Professional Review Disclaimer** heading." Same relative-path form the sibling agents in this directory already use.
- The `> [!IMPORTANT]` paraphrase is deleted. The Report Skeleton now shows a `## Professional Review Disclaimer` heading with a substitution placeholder, so placement is specified without creating a second copy that can drift again.

Kept it a placeholder rather than inlining the canonical text for exactly the reason this comment exists — an inlined copy is what drifted in the first place.

One related cleanup: SSSC Review Artifact Safeguards had been restating the disclaimer rule in full. It now points at Output Contract item 1, so the agent carries exactly one authoritative statement and one `#file:` reference.

Worth noting the block is **SSSC Planning**, not an SSSC Review variant — `disclaimer-language.instructions.md` has no review-specific section, and the planning text explicitly covers "review findings" and "review reports".

Fixed in `6fbd73ec`.

.github/agents/data-science-engineering/data-science-engineering-coach.agent.md
WilliamBerryiii: Applied as suggested — both duplicated paragraphs deleted.

The drift you spotted is the strongest argument here: the copy had already lost "in the installed collection", which is the qualifier that makes the availability rule actionable. That is the duplication failing in exactly the way the coach's own success criterion ("this agent does not copy its schemas or rule tables") exists to prevent.

What remains in `## Cross-cutting concerns`:

- The pointer sentence naming the registry as owner of trigger conditions, owning skills, and contributions.
- The four owning-skill bullets (`privacy-standards`, `rai-standards`, `rai-planner`, `telemetry-foundations`).
- The closing "Produce the confirmed job's durable output directly using its owning skill" paragraph.

Verified nothing downstream referenced the deleted sentences, and the `## Durable-write safety` section defines the write-blocking gate independently, so the "never blocks a durable write" rule is still enforced — it just lives at its one owner in `job-registry.md` now.

Fixed in `6fbd73ec`.

For transparency on the eval side: `data-science-engineering-coach` sits below its behavior threshold at 30.3%, but it was already at 23.6% before this change, so the deletion improved it rather than causing the shortfall. It is an advisory spec and a pre-existing failure; tracking it separately rather than in this PR.

evals/agent-behavior/README.md
WilliamBerryiii: Fixed both statements — and took your follow-up note rather than the suggestion, because you were right that the baseline is what's wrong.

Recounted against `AGENTS.yml` at the current head: **56 slug entries, 56 paths, 26 under `subagents/`, 30 parents.**

So the suggested "57 enrolled agents, including 22 subagents" would have replaced one wrong number with another. Both lines now read:

- Line 14 → "the maintained stimulus inventory contains **56** enrolled agents, including **26** subagents."
- Line 284 → "totals **56** agents: **30** parent agents plus **26** enrolled subagents."

The parent/subagent split was the larger error. The file claimed 35 parents and 22 subagents; the inventory has 30 and 26. That gap is big enough that it was never a simple off-by-one from the removed agent.

Since these are now derived from the generator-owned inventory rather than maintained by hand, they should stay correct as `AGENTS.yml` changes.

Fixed in `6fbd73ec`.

docs/reference/agents/data-science-engineering/data-science-engineering-coach.md
WilliamBerryiii: Verified the full table you built, and it's now clean — but the state had moved since you wrote it. Fifteen of the sixteen stale tokens were already corrected in earlier work on this branch; one had survived.

The survivor was `docs/reference/skills/data-science-engineering/data-catalog.md:31`:

> Use `ds-catalog` when an engagement needs a durable inventory of business entities...

now `data-catalog`. Fixed in `6fbd73ec`.

Swept every file in your table for `Data Workstream Coach`, `ds-catalog`, `ds-dataops`, `ds-evaluation-design`, `ds-feasibility`, and `ds-analysis-authoring`. All seven report clean, including the "Select **Data Workstream Coach** from the agent picker" line you flagged as the sharpest case.

The remaining `ds-catalog` occurrences in the repo are deliberate and left alone — the `DS_CATALOG_V1` contract name, the `ds-catalog-v1.schema.json` asset, and the `ds-catalog-v1.md` template. Those are versioned identifiers rather than skill names, so renaming them would break the schema `$id` and the validator path.

Your point about preserved tails being invisible to both the diff and `docs:generate:check` is the useful lesson here — 238 pages report zero drift while a preserved tail can still send a reader to a name that no longer exists in the picker. Worth a lint rule that greps preserved tails for retired artifact names; I'll raise that separately rather than widen this PR.

agreaves-ms (approved):

PR #2775, opened 2026-08-24
fix(skills): harden Mural CLI credential redaction paths

### LINKED ISSUE #2756: [Security] Unredacted MuralError text printed to stderr by mural CLI main()
## Summary

The top-level `MuralError` handler in the mural CLI `main()` writes the exception text to stderr without passing it through `_redact`. Several `MuralAPIError` messages are constructed directly from raw OAuth token-endpoint and asset-upload response bodies, so a token-refresh failure can print credential material in clear text.

This is the same failure class as #2460, but at a different sink. It is **not** covered by CodeQL alert 388 (already `fixed`), and it is not currently detected by any scanner.

## Evidence

The unredacted sink, in [`scripts/mural/__init__.py`](https://github.com/microsoft/hve-core/blob/main/.github/skills/experimental/mural/scripts/mural/__init__.py#L1249-L1253):

```python
except MuralError as exc:
    print(f"error: {exc}", file=sys.stderr)      # <-- not redacted
    return 1
except Exception as exc:  # noqa: BLE001
    print(f"internal error: {_redact(repr(exc))}", file=sys.stderr)   # <-- redacted
```

The broad `Exception` fallback redacts. The narrower `MuralError` handler — the one that actually carries API response bodies — does not.

A second `MuralError` handler earlier in the same file has the same defect:

```python
except MuralError as exc:
    print(str(exc), file=sys.stderr)      # <-- L1195, not redacted
    return EXIT_FAILURE
```

Both sites must be fixed. Repairing only L1249 leaves the vulnerability reachable through L1195.

Tainted messages reach that handler from [`scripts/mural/_transport.py`](https://github.com/microsoft/hve-core/blob/main/.github/skills/experimental/mural/scripts/mural/_transport.py):

| Line | Construction | Carries |
|---|---|---|
| L270 | `MuralAPIError(status, "TOKEN_INVALID_JSON", text)` | Raw token-endpoint response body |
| L315-L317 | `MuralAPIError(exc.code, "REFRESH_FAILED", text)` | Raw refresh-endpoint error body |
| L319 | `MuralAPIError(status, "REFRESH_FAILED", json.dumps(data))` | Full parsed refresh payload |
| L603 | `MuralAPIError(status, "ASSET_UPLOAD_FAILED", payload)` | Azure Blob response (SAS query strings) |

The clearest demonstration is the refresh path, where the **same string** is handled two different ways:

```python
except urllib.error.HTTPError as exc:
    text = _read_response_body(exc).decode("utf-8", errors="replace")
    _emit(f"refresh failed: HTTP {exc.code} {text}", level=logging.ERROR)   # redacted via _emit
    raise MuralAPIError(exc.code, "REFRESH_FAILED", text or "refresh failed") from exc
```

`_emit` routes through `_redact`. The re-raise does not, and `main()` then prints it verbatim. So the defense holds on the logging path and is bypassed on the exception path.

## Why this matters

[`mural-log-hygiene.instructions.md`](https://github.com/microsoft/hve-core/blob/main/.github/instructions/experimental/mural/mural-log-hygiene.instructions.md) states that `_redact` "only protects log output that is actually routed through it. Bare `LOGGER.*` and `print(*)` sites bypass it." This is exactly such a site, on a path that handles OAuth refresh tokens and client secrets. The code contradicts its own documented contract.

## Why scanners missed it

`py/clear-text-logging-sensitive-data` traced taint through a local dict in `_cli_auth.py` (alert 388) but does not follow taint across module boundaries through exception construction and re-raise into a top-level handler. Detection here requires either manual review or a custom query.

## Proposed fix

1. Wrap the output of **both** `MuralError` handlers (L1195 and L1249) in `_redact`, matching the adjacent `Exception` handler.
2. Audit the other `print(json.dumps(envelope), file=sys.stderr)` handlers in `main()` and route them through a redacting helper for uniformity.
3. Prefer not embedding raw response bodies in exception messages at all in `_transport.py`; carry a status/code plus a redacted excerpt.
4. Add a regression test alongside `tests/test_redaction.py` asserting that a `MuralAPIError` carrying token-shaped material is masked when surfaced by `main()`.

## Acceptance criteria

- [ ] `MuralError` output is redacted at both handler sites (L1195 and L1249)
- [ ] Remaining stderr JSON envelopes in `main()` use a redacting path
- [ ] `_transport.py` no longer embeds raw token/asset response bodies in exception messages
- [ ] Regression test covers the `main()` exception sink
- [ ] Reviewed and validated by a qualified human reviewer

## Related

- #2460 — CodeQL alert 388 (`_cli_auth.py`), already `fixed` upstream; residual hardening in that file handled separately

jkim323:
# Pull Request

## Description

The mural CLI leaked unredacted secret material on multiple output channels. `main()` printed raw `MuralError` text to stderr, the four structured error envelopes were serialized without redaction, `_emit_json` wrote unredacted payloads to stdout, and `_transport.py` embedded raw response bodies into error messages. OAuth responses, refresh-token failures, and asset-upload failures could therefore surface bearer tokens, client secrets, and SAS query strings in terminal output and CI logs.

This PR turns redaction into a structural barrier at the emit boundary rather than a text filter applied after serialization, then extends that barrier to every channel a later security sweep found still uncovered.

### Core redaction barrier

* All three `main()` stderr handlers pass error text through `_redact` before printing.
* New `_emit_json_error` redacts and emits error envelopes to stderr; the four envelopes (`human_authored_widget_protected`, `tag_merge_conflict`, `AREA_CAPACITY_EXCEEDED`, `bulk_atomic_abort`) route through it.
* New `_redact_payload` redacts the payload *structure* before serialization. Regex-scrubbing serialized JSON was structurally unsafe: the form-value pattern consumed closing quotes and produced output that no longer parsed, and escaped quotes in nested strings defeated the JSON pattern. It walks `str`, `dict`, `list`, `tuple`, `set`, and `frozenset`, and is key-aware, so a mapping key matching `_REDACT_KEYS` is masked regardless of its value type or quoting. That is strictly stronger than the previous string-level behavior.
* `_emit_json` (stdout) redacts through the same barrier.
* New `_error_excerpt` in `_transport.py` redacts and then truncates response bodies at `MAX_ERROR_EXCERPT_CHARS` (512). Ordering is deliberate: redacting first means truncation cannot sever a key name from its value and defeat a substitution.

### Follow-on hardening

A security sweep of this branch (OWASP Top 10, MCSB, Secure by Design, OWASP MCP, OWASP Agentic) found five coverage gaps in the barrier above. All five are closed here.

* **Azure SAS signatures now mask regardless of host casing.** The SAS pattern lacked `re.IGNORECASE` while `_validate_asset_url` normalizes with `parsed.hostname.lower()`, so a mixed-case host such as `acct.Blob.Core.Windows.Net` passed validation and then defeated redaction, putting a full `sig=` into a debug log line. A standalone `sig=` pattern was added as well, because an upstream error body can quote SAS parameters without the URL and truncation can strip the host prefix.
* **`_build_api_error` now shares the excerpt bound.** `_error_excerpt` covered five of six exception-construction sites. The sixth carries every non-2xx API response and was bounded only by `MURAL_MAX_BODY_BYTES` (16 MiB default), four orders of magnitude above the 512-character cap.
* **The record output channel now redacts.** `_emit_records` and `_emit_record` printed `_format_output(...)` directly while `_emit_json` redacted, which left the primary stdout data path as the only channel without a barrier.
* **Logout transparency restored.** A CodeQL Autofix commit on this branch replaced `removed.append(key)` with `removed.append("***")` in `_logout_remove_credentials`. The masked values are the env-style identifiers in `_KNOWN_CREDENTIAL_KEYS`, not credential values, so `mural auth logout` reported `(keys: ***, ***, ***)` and no secret was ever protected. The change also left the equivalent `errors[key]` path uncut, so it did not achieve its own stated goal. Reverted, and code scanning alert 723 is dismissed as a false positive: the expression it flags is the redaction sink itself, which CodeQL does not model as a sanitizer.
* **Threat models corrected.** Both `SECURITY.md` and `docs/security/security-model.md` described an "embedded stdio MCP server" that does not exist. The package registers no `mcp` subcommand, implements no JSON-RPC framing, and exposes no tool dispatch; the only MCP artifact is a `MCPInvalidParamsError` validation exception documented as retained for CLI helper compatibility. The repo-wide model additionally carried a full "Mural Skill MCP Server" entry asserting a `python -m mural mcp` deployment and scope re-checks "at dispatch". Both now describe the real surface: an argparse CLI whose stdout and stderr are captured into agent context through a terminal tool. The genuine hardening items from that entry were relocated rather than dropped.

### Test isolation fix

`test_autoload_credentials_returns_none_when_file_absent` never pinned a credential backend, so `resolve_backend` fell through to `auto` and read the developer's real keyring. It failed on any machine holding real Mural credentials, and its assertion form (`assert "MURAL_CLIENT_ID" not in env`) made pytest render the whole mapping into the failure message, printing live secret values. It now pins `MURAL_CREDENTIAL_BACKEND=file`, reuses the existing `_isolate_credential_env` helper, and compares key sets so a failure reports names only.

Out of scope: the residual `print(*)` sites in `_cli_auth.py`, `_commands.py`, and `_operations.py`, which #2756 explicitly defers to #2460. The security sweep also recorded lower-severity residuals (mapping-key redaction, argparse writing raw argv before the guarded block, SAS scope and expiry validation) that are not addressed here.

## Related Issue(s)

Fixes #2756

## Type of Change

**Code & Documentation:**

* [x] Documentation update

**AI Artifacts:**

**Other:**

## Testing

Run from `.github/skills/experimental/mural`:

* `uv run ruff check .` - all checks passed
* `uv run ruff format --check .` - 44 files already formatted
* `uv run pytest tests -q` - **860 passed, 143 skipped, 0 failed**

The suite has no failures. The one previously failing test is the environment-dependent credential test described above, fixed in this PR.

Repository gates: `npm run lint:py`, `npm run lint:frontmatter`, `npm run lint:tables`, and `npm run validate:skills` all clean (74 skills, 0 errors, 0 warnings).

44 test functions were added across four files:

* `tests/test_redaction.py` (29) - `main()` error sinks and source-level pins, `_redact_payload` container coverage, sensitive mapping keys, `_emit_json` stdout channel, SAS host casing and bare-signature masking, record-channel redaction
* `tests/test_mural_transport.py` (11) - `_error_excerpt` unit behavior, `_build_api_error` bounding and masking, end-to-end transport coverage
* `tests/test_logout_transparency.py` (4) - `removed_keys` content and rendered summary, which no test previously pinned
* `tests/test_credential_storage.py` - the isolation fix described above

Each security fix was checked for vacuity by temporarily reverting the source change and confirming the new tests fail. That check earned its keep: it caught that two of the four logout-transparency tests exercise only the renderer and pass against the regressed code, so the coverage claim is scoped to the two that do fail.

### Required Checks

## Security Considerations

## Additional Notes

Two points worth a reviewer's attention.

**Alert 723 is dismissed, not fixed.** CodeQL flags `print(json.dumps(_redact_payload(payload), indent=2))` in `_emit_json`, which is the redaction sink introduced by this PR. `_redact_payload` is not modeled as a sanitizer, so the safest path in the module is the one most likely to be flagged. Reverting the Autofix restores the flagged dataflow, which is why the dismissal is a prerequisite rather than a cleanup step. Modeling the sanitizer properly is tracked as follow-up work.

**Record-channel redaction can over-mask user content.** Routing `_emit_records` through `_redact_payload` means a Mural sticky note whose text legitimately reads `code=ABC123` now renders as `code=***` in list output. That is a real behavior change on the primary data path, accepted because the alternative leaves the busiest stdout channel as the only one without a barrier, and because `_emit_json` already accepts the same trade-off for envelopes.


.github/skills/experimental/mural/scripts/mural/_output.py
```diff
@@ -171,21 +265,28 @@ def _apply_widget_text_coalesce(payload: Any) -> Any:
...
+    as :func:`_emit_json`. Without it the primary stdout data path would be the
+    one channel in the package that emits unredacted upstream text.
+    """
     _apply_widget_text_coalesce(records)
     fields = _read_fields(args)
     fmt = (
         "json" if _state._CLI_FORCE_JSON else (getattr(args, "format", None) or "json")
     )
-    print(_format_output(records, fields, fmt))
+    print(_format_output(_redact_payload(records), fields, fmt))
```
WilliamBerryiii: This applies the diagnostic redactor to arbitrary Mural-authored records. A legitimate note such as `use code=ALPHA` now returns as `use code=***`, and sensitive-looking field names are replaced regardless of whether they hold credentials. Because this is the primary successful-output path, can we preserve user content and redact only fields known to carry transport credentials or validated SAS URLs? Please add fidelity tests for JSON and table output with credential-shaped but legitimate text.

.github/skills/experimental/mural/scripts/mural/_transport.py
```diff
@@ -90,6 +95,24 @@ def _redact(text: str) -> str:
...
 
 
+def _error_excerpt(text: str) -> str:
+    """Return a redacted, length-bounded excerpt of an upstream response body.
+
+    Exception messages built from token-endpoint or asset-upload responses are
+    surfaced to the operator, so the raw body must never travel inside them.
+    Redaction runs before truncation: cutting first could sever a key name from
+    its value and defeat the substitution patterns. Truncation then runs as
+    defense-in-depth, bounding both unrecognised secret shapes and hostile
```
WilliamBerryiii: Bounding the body helps, but `_redact` is still a regex allow-list rather than a fail-closed barrier. For example, malformed `{"refresh_token":"SECRET`, case-varied `Access_Token`, and escaped key names survive `_error_excerpt`, then the top-level handler applies the same patterns. At token and SAS boundaries, could we avoid carrying opaque bodies and emit only allow-listed status/error metadata? This is a potential disclosure path, not evidence that an actual secret has leaked.

.github/skills/experimental/mural/scripts/mural/__init__.py
```diff
@@ -1240,14 +1243,14 @@ def main(argv: list[str] | None = None) -> int:
...
+        _emit_json_error(envelope)
         return EXIT_AREA_CAPACITY
     except MuralBulkAtomicAbort as exc:
         envelope = {"error": "bulk_atomic_abort", "aborted": True, **exc.summary}
-        print(json.dumps(envelope), file=sys.stderr)
+        _emit_json_error(envelope)
         return EXIT_TEMPFAIL
     except MuralError as exc:
-        print(f"error: {exc}", file=sys.stderr)
+        print(f"error: {_redact(str(exc))}", file=sys.stderr)
```
WilliamBerryiii, line 1261: `MuralAPIError.__str__` includes trusted diagnostics as `code=REFRESH_FAILED`, so redacting the complete string changes that to `code=***`. Could this handler preserve `status` and `exc.code` structurally and redact only the untrusted message/request-id fields? An end-to-end test should keep `REFRESH_FAILED` visible while masking a secret in the message.

.github/skills/experimental/mural/SECURITY.md
```diff
@@ -23,17 +23,17 @@ This document records the STRIDE threat model for the Mural skill (the `mural` p
...
+The Mural skill is a local Python CLI dispatched through `argparse`. It authenticates to Mural with OAuth 2.0 Authorization Code + PKCE, caches access and refresh tokens in the OS keyring (or a `0600` file fallback), and makes authenticated HTTPS calls to the Mural REST API. Its highest-risk behaviors are at-rest credential storage on the operator workstation and the browser-mediated OAuth login flow; both are mitigated in code, with residual gaps tracked in the gap register. The skill runs no public listener (the loopback receiver is single-shot and bound to `127.0.0.1`) and treats all Mural-authored content returned through the CLI as untrusted. An agent caller reaches the skill by invoking the CLI through a terminal tool, so both stdout and stderr are captured into agent context; that is why every output sink routes through the redaction barrier described in B4 rather than only the operator-facing ones.
```
WilliamBerryiii: “Every output sink” is broader than this PR establishes. The PR explicitly defers residual direct-print paths, and opaque response bodies are still protected only when their representation matches known patterns. Please name the covered channels and avoid an absolute guarantee until a complete sink inventory and fail-closed body policy support it.

.github/skills/experimental/mural/SECURITY.md
```diff
@@ -417,6 +418,7 @@ The following gaps are known limitations of the current implementation. They are
...
 | G-INF-4 | Mural payloads returned through CLI output may include PII or confidential workshop content (A4). The skill does not classify or redact this content; downstream handling is the caller's responsibility.                                                                                                                                                                                                                                                                                               | InfoDisc-Med    | By design; documented in B4 Information Disclosure.                                                                                                                                                                                                                                                                                                                                                                                                                                      |
```
WilliamBerryiii, line 421: This now conflicts with `_emit_record` and `_emit_records`, which apply credential-pattern and sensitive-key redaction to complete Mural payloads. Please distinguish “does not classify or comprehensively redact PII/confidential content” from the new credential-pattern masking, and document the acknowledged risk of altering legitimate user-authored content.

.github/skills/experimental/mural/tests/test_redaction.py
```diff
@@ -276,3 +325,417 @@ def test_logger_format_string_redacts_exception_repr_at_runtime(
...
+    assert SECRET_VALUE not in captured.err
+    assert json.loads(captured.err)["detail"] == "code=***"
+
+
+# ---------------------------------------------------------------------------
+# `main()` sink call-site contracts (defense-in-depth)
+# ---------------------------------------------------------------------------
+
+
+def test_main_error_handlers_wrap_exception_text_in_redact(
```
WilliamBerryiii: These exact source-string checks are brittle: harmless formatting can fail them, while an unsafe equivalent spelling can pass. The nearby `_drive_main` seam can verify every structured handler behaviorally; transport response tests can verify each construction path without coupling the suite to source text.

WilliamBerryiii (changes requested):
The issue identifies a real reachable propagation path: raw upstream response text could be carried in `MuralAPIError` and printed by an unredacted typed-error handler. For credential exposure to occur through this path, however, the Mural or Azure API, or an intermediary handling that traffic, would first have to include credential material in its response body. The reviewed evidence does not show that this happened in production, so it supports potential exposure rather than the statement that production secrets actually leaked. The description should also acknowledge that `_emit_json` already performed post-serialization regex redaction; this PR replaces that with safer structural redaction. Please also change the PR title from `fix(skills): redact secrets on every mural CLI output channel` to the evidence-bounded `fix(skills): harden Mural CLI credential redaction paths`, or equivalent wording that does not imply confirmed secret exposure or complete coverage of every output channel.

I am requesting changes for two implementation concerns. Applying the diagnostic redactor to all record output alters legitimate Mural-authored content, and regex-only excerpts still allow malformed, case-varied, or encoded credential representations to reach captured stderr. The inline comments also cover loss of stable API error codes, overbroad “every output sink” documentation, the G-INF-4 contradiction, and brittle source-text tests.

The focused changed-area suite passes (167 passed, 18 skipped), and the PR makes several solid improvements: typed error sinks are covered, structured JSON remains parseable, SAS matching is stronger, and credential tests are isolated from the real keyring. Please narrow the output behavior, make credential-boundary diagnostics fail closed, and revise the claims to match the demonstrated evidence.

.github/skills/experimental/mural/scripts/mural/_output.py
jkim323: Addressed. Record output no longer applies the diagnostic redactor to arbitrary Mural-authored content.

`_emit_record` and `_emit_records` now preserve credential-shaped text and sensitive-looking field names in both JSON and table output. Masking is limited to complete Azure Blob URLs that pass the existing validator and contain a parsed sig query parameter.

I replaced the previous broad-masking tests with:

- JSON fidelity coverage for `code=ALPHA`, `access_token=placeholder`, and a `client_secret` field
- Equivalent table-output fidelity coverage
- A regression proving validated Azure Blob SAS URLs remain masked

.github/skills/experimental/mural/scripts/mural/_transport.py
jkim323: Addressed. Token and Azure SAS failure paths no longer carry opaque response bodies or raw Content-Type values into exceptions or stderr.

The token parser, refresh flow, authorization-code exchange, and SAS upload paths now expose only:

- HTTP status
- Stable internal error code
- Static diagnostic message

Ordinary Mural API diagnostics still use the existing bounded `_error_excerpt` path. I added behavioral coverage for malformed JSON, case-varied and escaped keys, hostile Content-Type values, refresh stderr, returned error statuses, and `HTTPError` paths.

.github/skills/experimental/mural/SECURITY.md
jkim323: Addressed. The security model no longer claims that every output sink routes through one redaction barrier.

The Executive Summary and B4 now name the protections established by the current implementation:

- `_emit` diagnostics and selected transport URL logs use `_redact`
- `_emit_json` and `_emit_json_error` structurally redact machine-readable envelopes
- top-level typed, scope, structured, and unexpected error handlers redact or use the structured helper
- record output preserves Mural-authored content and masks only validated Azure Blob SAS URLs containing sig
- ordinary Mural API errors use bounded _error_excerpt diagnostics
- token parsing, refresh, authorization exchange, and SAS upload failures discard opaque bodies and retain only status, stable code, and static messages

The model now explicitly states that remaining direct print and `LOGGER` calls are outside a package-wide guarantee and require sink-specific review. I also updated the G-INF-5 audit row to reflect the current fail-closed body policy. G-INF-4 was left unchanged because it is covered by the separate review thread.

.github/skills/experimental/mural/tests/test_redaction.py
jkim323: Addressed. I removed the three exact source-string tests and replaced their intent with behavioral coverage.

The existing `_drive_main` seam now exercises all four structured handlers:

- `MuralHumanAuthoredProtected`
- `MuralTagMergeConflict`
- `MuralAreaCapacityExceeded`
- `MuralBulkAtomicAbort`

Each case verifies the deterministic exit code, empty stdout, parseable stderr JSON, expected error identifier, exact envelope keys, and credential-marker redaction.

Existing behavioral tests continue to cover the credential-autoload, auth-scope, and typed `MuralError` handlers. The transport and OAuth suites remain the authority for ordinary API, token parser, refresh, authorization exchange, and SAS construction paths. Unrelated logger source-contract tests were left unchanged.

.github/skills/experimental/mural/SECURITY.md
jkim323: Addressed in 74451ba9. The security model no longer claims that every output sink routes through a single redaction barrier.

The Executive Summary and B4 now describe the protections by channel:

- Diagnostic and selected transport output uses _redact
- Machine-readable envelopes use structural redaction
- Record output preserves Mural-authored content and narrowly masks validated Azure Blob SAS URLs
- Ordinary Mural API errors use bounded excerpts
- Token and SAS failures discard opaque response bodies and retain only status, stable code, and static messages

The model also explicitly states that the remaining direct print and LOGGER paths are outside any package-wide guarantee and require sink-specific review. G-INF-5 was updated to reflect the fail-closed token and SAS policy.

.github/skills/experimental/mural/scripts/mural/__init__.py
jkim323: Addressed in b3936183. I added a dedicated MuralAPIError handler before the generic MuralError handler.

The handler now preserves status and exc.code structurally, so diagnostics such as `code=REFRESH_FAILED` remain visible. Only `exc.message` and `exc.request_id` are passed through _redact.

The end-to-end regression verifies:

- HTTP status and `REFRESH_FAILED` remain visible
- Credential-shaped content in the message and request ID is masked
- None and empty optional fields are omitted with correct punctuation
- Output remains stderr-only with exit code 1

WilliamBerryiii (approved):

agreaves-ms (approved):

PR #2718, opened 2026-08-14
feat(skills)!: harden provider authentication and port the redaction triad to jira and gitlab

WilliamBerryiii:
# Pull Request

## Description

This PR hardens the authentication boundaries for the Mural, GitLab, and Jira skills, and ports the Mural redaction architecture into GitLab and Jira.

The GitLab skill moves to public-client OAuth as its default. Authorization Code with PKCE runs over a fixed loopback callback, with a device-authorization flow for browserless environments. PAT authentication is still available, but only through an explicit `GITLAB_AUTH_MODE=legacy-token`; mixed OAuth and PAT configuration fails closed. OAuth profiles are bound to the issuer and client ID and persist with owner-only permissions, symlink-resistant atomic writes, and both in-process and cross-process locking.

The Jira skill gains scoped Cloud token routing through the fixed Atlassian resource origin, using a validated single-segment Cloud ID. Data Center PAT, unscoped Cloud, and scoped Cloud configurations are mutually isolated.

Both skills received the redaction triad that the Mural skill already carried: a pinned `_REDACT_KEYS` tuple, a `_REDACT_PATTERNS` table covering JSON, escaped-JSON, bare form, percent-encoded, and Azure SAS shapes, and a single `_redact` entry point. Every write to stdout and stderr is funneled through emit helpers that redact first, and typed API errors render only a status, method, query-stripped resource, redacted summary, and request ID.

Mural is constrained to the canonical production API destination, permitting a loopback development destination only when explicitly enabled. Its token, API, and SAS-upload requests now use separate no-redirect openers with finite timeouts.

* feat(skills): routed the GitLab CLI through public-client OAuth with PKCE, added `auth login`, `auth device-login`, `auth status`, and `auth logout` commands, and moved the resolved credential off the mutable `gitlab_token` module global onto a frozen `AuthContext` whose token field carries `repr=False`

* feat(skills): added `_gitlab_oauth.py` and `_gitlab_credentials.py`, covering PKCE generation, the bounded loopback callback that validates exact host, path, state, and deadline, device-authorization polling that honors `authorization_pending` and `slow_down`, refresh rotation, and the locked, owner-only profile store

* feat(skills): added scoped Jira Cloud token routing with Cloud ID validation, and isolated the three Jira authentication modes so mixed configuration fails closed

[...]

PR #2779, opened 2026-08-25
refactor(skills)!: remove the GitLab die() helper in favor of GitLabError

### LINKED ISSUE #1555: [security] Architectural: introduce central output sink + LOGGER + typed error class to jira/gitlab skills
## Background

The Mural skill's recent OAuth security hardening landed three coupled
abstractions that together form the redaction contract:

- **`LOGGER`** (a module-level `logging.getLogger(name)`) — single observable
  log surface.
- **`_emit()` / `_emit_debug_traceback()`** — single output sink that runs
  every formatted string through `_redact()` before reaching `LOGGER` or
  `stderr`.
- **`MuralAPIError`** — typed error class with a controlled `__str__` (status,
  code, message, request_id) that never carries raw header dicts or response
  bodies into exception chains.

References:

- `.github/skills/mural/mural/scripts/mural.py` L298 (`LOGGER`), L312
  (`MuralAPIError`), L1540 (`_emit`), L1547 (`_emit_debug_traceback`)
- `.github/skills/mural/mural/SECURITY.md` Bucket B2 / B4 Information
  Disclosure rows

**Without these three abstractions, redaction is unenforceable** — a
contributor can always reach `print(..., file=sys.stderr)` directly, and a
freshly-raised exception will carry the raw upstream body into the traceback.
This is precisely the structural pattern that allowed **G-INF-1**
(`client_secret` not in the keyset) to remain undetected for as long as it
did, until the central sink and source-contract tests were introduced.

### `.github/skills/jira/jira/scripts/jira.py` (472 LOC)

| Site | Gap |
|------|-----|
| L37 `class ScriptError(Exception)` | Generic; no `status`/`code`/`request_id`; no controlled `__str__` |
| L97 `def request(...)` | No `LOGGER`, no debug logging, no central sink |
| L120, L124-128, L132-134 | Raw URL + raw upstream body embedded in exception text |
| L143-159 `_extract_error_message` | Returns upstream body verbatim |
| L204 `_print_selected_fields`, L218 `_print_result` | Raw `print()` of upstream JSON |
| L323 `handle_comment` reads `sys.stdin.read()` bare | Raw payload could land in error traceback |

### `.github/skills/gitlab/gitlab/scripts/gitlab.py` (439 LOC)

| Site | Gap |
|------|-----|
| L34 `gitlab_url`, `gitlab_token`, `api_url` module globals | Mutable global secret state |
| L40 `die()` helper | Raw `print(..., file=sys.stderr)` + `SystemExit`; no error class at all |
| L122 `def request(...)` | No `LOGGER`, no central sink |
| L154-159, L161, L168, L172 | Raw upstream body / URL into stderr or stdout |
| **L387-394 `cmd_job_log` second `urlopen` site** | Bypasses `request()` entirely; raw `print(error.read().decode(), file=sys.stderr)` |
| L218 `print_fields()` | Raw `print()` of arbitrary upstream field values |
| L427-437 `main()` | Bare `print(..., file=sys.stderr)` for KeyboardInterrupt + BrokenPipeError |

## Why this is an architectural prerequisite, not a refactor

Adding `_REDACT_KEYS` / `_REDACT_PATTERNS` / `_redact()` to either skill
**without** first introducing `LOGGER` + `_emit()` + a typed error class
results in a redaction net with holes wherever a `print()` or a raw exception
chain still exists. Specifically:

1. The redaction call must live somewhere — `_emit()` is that somewhere.
   Otherwise, every site that wants to log must remember to call `_redact()`,
   which is exactly the failure mode source-contract tests exist to prevent.
2. Raised exceptions are formatted by Python at the top of the stack — if the
   error message contains an unfiltered upstream body, the traceback printed
   on `--no-debug` paths leaks it. A typed error class with a controlled
   `__str__` is the choke point.
3. GitLab's `cmd_job_log` second `urlopen` site (L387-394) is the cleanest
   evidence: even after redaction is added to `request()`, this site would
   continue to leak. It must be refactored to route through `request()` (or
   through a shared `_http_get_raw()` helper that itself uses `_emit()` for
   error paths).

This issue **must merge before** the Jira and GitLab redaction-port issues
land.

### Common to both skills

1. `LOGGER = logging.getLogger("jira")` / `logging.getLogger("gitlab")`.
2. `_emit(level, message, *args)` — formats, redacts, routes to `LOGGER.log`
   and (for user-facing errors) to `sys.stderr`.
3. `_emit_debug_traceback(exc)` — gated on `JIRA_DEBUG` / `GITLAB_DEBUG`;
   calls `traceback.format_exception` and pipes the result through
   `_redact()` (depends on the redaction-port issues for the actual
   `_redact` function, but the call site lands here).
4. Typed error class with controlled `__str__`:
   - `JiraAPIError(ScriptError)` — fields: `status`, `code`, `message`,
     `request_id`. Replaces raw `ScriptError(f"HTTP ... details ...")` at
     L124-128 / L132-134.
   - `GitLabAPIError(Exception)` — same field set. Replaces every `die()`
     call site (gitlab has no error class at all today).
5. Refactor every existing `print(...)` / `print(..., file=sys.stderr)` /
   `die(...)` to route through `_emit()` or `raise <error class>(...)`.

### GitLab-specific

6. **Refactor `cmd_job_log` L387-394**: collapse the second `urlopen` site
   into the central `request()` helper. Add a source-contract test asserting
   exactly one `urlopen` call site in the file.
7. Refactor `main()` L427-437 (`KeyboardInterrupt`, `BrokenPipeError`) to
   route through `_emit()`.
8. Move `gitlab_token` off the module global into a small immutable config
   object mirroring Jira's `JiraClient` dataclass.

## Acceptance criteria

- [ ] Both skills define `LOGGER`, `_emit`, `_emit_debug_traceback`, and a
      typed API error class.
- [ ] Source-contract test: zero `print(..., file=sys.stderr)` outside
      `_emit()`.
- [ ] Source-contract test: zero `LOGGER.exception(` calls (mirror
      `.github/skills/mural/mural/tests/test_redaction.py`
      `test_logger_no_bare_exception_calls`).
- [ ] GitLab: source-contract test asserts exactly one `urlopen` call site.
- [ ] GitLab: no `die()` helper remaining; all error paths use the
      `GitLabError` hierarchy, and API transport failures use
      `GitLabAPIError`.
- [ ] GitLab: `gitlab_token` is no longer a module-level mutable global.
- [ ] Jira: `JiraClient.__repr__` does not contain the raw `auth_header`
      value.
- [ ] Both skills: typed error class `__str__` excludes raw upstream body
      (negative test).

## Dependencies

- **Blocks** the Jira redaction-port issue.
- **Blocks** the GitLab redaction-port issue.
- **Blocks** the per-skill CI gating issue.

## References

- `.github/skills/mural/mural/scripts/mural.py` L298, L312, L1540, L1547
- `.github/skills/mural/mural/tests/test_redaction.py` source-contract section
- `docs/security/security-model.md` `OA-1`..`OA-17`

jkim323:
# Pull Request

## Description

Removes the `die()` helper from the GitLab skill so `GitLabError` becomes the module's single failure mechanism, and adds a source contract so the boundary cannot drift again.

`GitLabError` is the module-wide typed failure base. `GitLabAPIError(GitLabError)` is the API-specific subclass used for transport failures; non-transport validation, configuration, and command errors remain `GitLabError`.

This closes the last open acceptance criterion of #1555. The other seven criteria — `LOGGER`, the `_emit` output sink, `_emit_debug_traceback`, typed API error classes, the source-contract tests, and the `gitlab_token` global removal — already landed in #2718 for both the Jira and GitLab skills. **No Jira changes are needed.**

### Why this is not cosmetic

`die()` was already redaction-safe: it routed through `_emit()`, so the raw-stderr leak described in the issue was fixed by #2718. The remaining problem was an internal contract violation. The `GitLabError` docstring stated that `die()` survives only for "the argument-parsing and command-dispatch layer, where no value is promised" — yet `die()` was being called from four helpers that *do* promise return values:

| Helper                      | Returns          |
|-----------------------------|------------------|
| `_auth_headers`             | `dict[str, str]` |
| `_required_oauth_client_id` | `str`            |
| `project`                   | `str`            |
| `_read_capped`              | `bytes`          |

That is exactly the "silent `None` return" hazard the docstring warns against. The documented contract and the code disagreed, and nothing enforced the split.

### What changed

* All 61 `die()` call sites now `raise GitLabError(message, exit_code)`, preserving each original message and exit-code constant. Sites inside `except` blocks chain with `from exc`.
* `die()` is deleted; the now-unused `NoReturn` import is dropped.
* `GitLabError`'s docstring is rewritten to describe the single-mechanism contract.
* 42 `SystemExit` test references are migrated across 7 files.
* New source contract: no `raise SystemExit`, no `sys.exit(...)`, and no `die` definition outside the `__main__` guard — with one negative case per predicate.

### Behavior

Exit codes are unchanged. `main()` already caught `GitLabError`, emitted once via `_emit()`, and returned `exc.exit_code`, so no new error handling was introduced.

Two deliberate consequences, both improvements:

1. Library helpers now raise *without* emitting. Emission happens once, at the `main()` boundary, removing a double-emission path.
2. `SystemExit` is a `BaseException` and previously bypassed `main()` entirely. These failures now flow through `main()`, so `GITLAB_DEBUG=1` shows a redacted traceback for argument-validation failures that previously printed only a one-line error.

## Related Issue(s)

Closes #1555

## Type of Change

**Code & Documentation:**

**Other:**

`die()` was a public module attribute, so its removal is technically breaking for any external caller importing `gitlab.die`. The CLI surface, exit codes, and output format are unchanged.

### Test migration

The migration was not a uniform substitution. Sites fell into three kinds:

| Kind             | Migration                                                                                                                                 |
|------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
| Direct-call      | `pytest.raises(gitlab.GitLabError)`, `.code` to `.exit_code`, and stderr assertions moved to `str(exc_info.value)` because converted helpers no longer emit |
| `main()` boundary | `pytest.raises` removed entirely; assert `gitlab.main() == <constant>` plus emitted stderr                                                 |
| Already typed    | Untouched                                                                                                                                 |

No assertion was weakened to a bare `pytest.raises(Exception)`. Every migrated site that checked an exit code still checks the same constant.

`TestDie` was removed. Its exact `error: boom` single-line assertion was **not** already covered elsewhere — `test_emit_writes_exactly_one_stderr_line` exercises `_emit` directly, and no test previously reached `main()`'s handler because `SystemExit` bypassed it. A replacement assertion was added to a converted `main()` test:

```python
assert capsys.readouterr().err == "error: --fields is not valid with auth commands\n"
```

### Validation status

| Command                   | Result                                          |
|---------------------------|-------------------------------------------------|
| `npm run lint:py`         | Pass                                            |
| `npm run validate:skills` | Pass — 74 skills, 0 errors, 0 warnings          |
| `npm run test:py`         | **Blocked — environmental, not a code failure** |

`npm run test:py` could not execute locally. `uv sync` fails with `os error 32` inside the OneDrive-synced workspace and `os error 396` outside it, PyPI is unreachable with a TLS `HandshakeFailure`, and the existing skill virtualenvs hold a partially installed `pytest`. This predates the change and affects unrelated skills identically.

Because the suite could not run, acceptance criteria were verified directly against system Python: `die` is absent from the module, the source contract rejects all three synthetic violations and accepts the guarded `sys.exit(main())`, and no `SystemExit` reference remains outside the contract's own test fixtures.

**Reviewers: please confirm the GitLab suite passes in CI before merging.** That is the one gate not satisfied locally.

### Required Checks

* [x] Documentation is updated (if applicable) — n/a, no doc-visible behavior change
* [x] Changes are backwards compatible (if applicable) — see Type of Change

### Required Local Checks

* [x] `npm run validate:local`
* [x] `npm run spell-check`
* [x] `npm run lint:md-links`

## Security Considerations

No new dependencies. No change to `_redact`, `_REDACT_KEYS`, `_REDACT_PATTERNS`, the emit helpers, `AuthContext`, the OAuth modules, or the audit sink. The existing print-ownership, `LOGGER.exception`, and two-owner credentialed-egress contracts pass unmodified.

## Additional Notes

The `path:line` references in #1555 are stale — the skills moved from `.github/skills/{jira,gitlab}/{jira,gitlab}/` to `.github/skills/project-planning/{jira,gitlab}/` and both files roughly tripled in size. Worth correcting the issue body on close.

The PR validation failure on `Asset Docs Validation` was due to the shared PowerShell module installer assuming `PSGallery` was already registered on the runner. The installer now self-registers `PSGallery` on demand, and the fix is covered by focused tests in `scripts/tests/security/Install-PSModules.Tests.ps1`.
Coverage follow-up for review thread 3850925767: the complete production-only Ubuntu run now passes 288 tests with 87.70% line coverage (1,062/1,211), 78.86% branch coverage (317/402), and 85.49% combined coverage. The terminal report publishes missing lines and partial branches for each runtime module: `_gitlab_credentials.py` 84%, `_gitlab_oauth.py` 79%, and `gitlab.py` 88%. The GitLab project now measures branches across `scripts`, reports missing lines, and enforces an 80% combined floor through the existing pytest CI lane. A focused regression also verifies typed `GitLabError` exit-code propagation, single redacted emission, and debug-traceback delegation at `main()`.


.github/skills/project-planning/gitlab/tests/test_gitlab_helpers.py
```diff
@@ -83,15 +83,59 @@ def _network_invocations(source: str, module: str) -> list[tuple[str, str, str]]
...
+    to the ``__main__`` guard alone, so a ``raise SystemExit`` or ``sys.exit``
+    anywhere else, or any ``die`` definition, reintroduces the second mechanism
+    this contract exists to prevent.
+    """
+    tree = ast.parse(source)
+    guarded: set[int] = set()
+    for node in ast.walk(tree):
+        if _is_main_guard(node):
+            for child in ast.walk(node):
+                guarded.add(id(child))
```
katriendg: The source contract currently treats any comparison with `__name__` and `"__main__"` as the main guard, and `ast.walk(node)` exempts the entire `if`, including `orelse`. The review probe confirmed that both `if __name__ != "__main__": sys.exit(2)` and `sys.exit(2)` in a canonical guard's `else` branch return no violations.

Suggested fix:

```python
def _is_main_guard(node: ast.AST) -> bool:
    if not isinstance(node, ast.If) or not isinstance(node.test, ast.Compare):
        return False
    return (
        isinstance(node.test.left, ast.Name)
        and node.test.left.id == "__name__"
        and len(node.test.ops) == 1
        and isinstance(node.test.ops[0], ast.Eq)
        and len(node.test.comparators) == 1
        and isinstance(node.test.comparators[0], ast.Constant)
        and node.test.comparators[0].value == "__main__"
    )

for node in ast.walk(tree):
    if _is_main_guard(node):
        for statement in node.body:
            for child in ast.walk(statement):
                guarded.add(id(child))
```

Please also add negative cases for `!=` and for an exit in `orelse` so the contract proves both boundaries.

.github/skills/project-planning/gitlab/scripts/gitlab.py
```diff
@@ -1329,15 +1342,15 @@ def cmd_auth_login(args: list[str]) -> None:
             audit_outcome=_oauth_audit_outcome,
         )
     except oauth.OAuthError as exc:
-        die(str(exc), EXIT_FAILURE)
+        raise GitLabError(str(exc), EXIT_FAILURE) from exc
```
katriendg: `from exc` preserves provider-controlled `OAuthError` text as a raw cause. `main()` redacts its own formatted traceback, but direct command-helper consumers or external exception logging can render the cause before that sink. A runtime probe with a token-shaped OAuth error confirmed that the raw chain contains the secret while `_redact()` removes it.

Suggested fix at both this authorization-code path and the device-login path on line 1370:

```python
except oauth.OAuthError as exc:
    raise GitLabError(_redact(str(exc)), EXIT_FAILURE) from None
```

Please add a regression test that formats the complete exception traceback from direct helper use and asserts token-shaped provider text is absent. Applying the same treatment to both OAuth login wrappers keeps the public helper boundary consistent.

.github/skills/project-planning/gitlab/scripts/gitlab.py
```diff
@@ -196,12 +196,16 @@ class AuthContext:
 class GitLabError(Exception):
     """Base CLI failure carrying an exit code and a redacted string form.
 
-    Library-level helpers raise this instead of calling :func:`die`. A helper
-    that promises a return value must end every path in an explicit ``return``
-    or ``raise``; relying on ``die`` never returning makes the contract
-    unverifiable by static analysis and turns a future change to ``die`` into a
-    silent ``None`` return. :func:`die` stays for the argument-parsing and
-    command-dispatch layer, where no value is promised.
+    This is the only failure mechanism in the module. Every error path raises
```
katriendg, line 199: This implementation consistently establishes `GitLabError` as the typed failure base, with `GitLabAPIError` as the API-specific subclass. Issue #1555's acceptance criterion instead literally says that all error paths use `GitLabAPIError`, so closing the issue would leave the recorded requirement inconsistent with the accepted design.

Suggested fix: before merge, update #1555's criterion to state that all error paths use the `GitLabError` hierarchy and that API transport failures use `GitLabAPIError`. Mirror that clarification in the PR description. If the narrower `GitLabAPIError` requirement was intentional, align the implementation instead.

.github/skills/project-planning/gitlab/tests/test_gitlab_helpers.py
```diff
@@ -288,10 +332,31 @@ def test_source_contract_detects_non_owner_invocations(
         self, source: str, expected: str
     ) -> None:
         invocations = _network_invocations(source, "unexpected.py")
-
         assert invocations
         assert invocations[0][1] == expected
 
+    def test_gitlab_error_is_the_only_failure_mechanism(self) -> None:
```
katriendg, line 429: The suite passes 283 tests, but the PR evidence does not report the measured line or branch coverage for this GitLab skill. The apparent coverage level of roughly 70% is too low for a security-sensitive CLI boundary that owns OAuth, credential handling, redaction, audit fail-closed behavior, and transport validation. Please run the skill suite with coverage, publish the missing-line report, and add tests for the highest-risk uncovered branches. If the current value is near 70%, please raise it materially (at least above 80%, or to the repository-defined threshold if higher) and enforce the agreed floor in CI so it cannot regress. Branch coverage for the newly changed error and exception paths would be especially useful.

katriendg (approved):
Thank you. The failure-mechanism migration is consistent, the GitLab suite and repository checks pass, and the breaking API removal is disclosed. I found three medium follow-ups: tighten the AST contract so non-equality and `else` exits cannot bypass it; prevent provider-controlled OAuth causes from escaping redaction for direct helper consumers; and reconcile issue #1555's `GitLabAPIError` wording with the `GitLabError` base type used here before closing the issue.

Approving, as comments need to be closed before merge, and I know you do that!

.github/skills/project-planning/gitlab/tests/test_gitlab_helpers.py
jkim323: Thanks for catching both gaps. Addressed in 230401a4:

- _is_main_guard now requires an ast.Eq comparison.
- Only statements in the canonical guard body are exempted; orelse remains subject to the source contract.
- Added regression cases for __name__ != "__main__" and sys.exit in the else branch.
- Validation passed: 83 helper tests, Ruff lint/format, and git diff --check.

.github/skills/project-planning/gitlab/scripts/gitlab.py
jkim323: Thanks for catching this. Addressed in 442dcea9:

- Both OAuth login wrappers now redact the provider-controlled message before translating it to GitLabError.
- Both translations use from None, preventing the raw OAuthError cause from appearing in complete tracebacks.
- Added parameterized regression coverage that formats the full traceback for direct authorization-code and device-login helper calls and confirms the provider secret is absent.

.github/skills/project-planning/gitlab/tests/test_gitlab_helpers.py
jkim323: Measured and gated this on the runtime code rather than the test files. The complete Ubuntu run now passes 288/288 tests at 87.70% line coverage (1,062/1,211), 78.86% branch coverage (317/402), and 85.49% combined coverage. The term-missing report publishes uncovered lines and partial branches per module (`_gitlab_credentials.py` 84%, `_gitlab_oauth.py` 79%, `gitlab.py` 88%).

I added project-level coverage configuration that scopes measurement to `scripts`, enables branches, reports missing lines, and fails below 80% through the existing pytest CI job. I also added one focused `main()` regression covering typed `GitLabError` exit-code propagation, one redacted emission, and debug-traceback delegation. The remote GitLab Python Tests job is green on commit `4139c73a`.

scripts/security/Install-PSModules.ps1
```diff
@@ -144,6 +144,37 @@ function Test-ModulePresent {
...
+    $repo = Get-PSRepository -Name $Name -ErrorAction SilentlyContinue
+    if ($repo) {
+        return
+    }
+
+    if ($Name -ne 'PSGallery') {
+        return
+    }
+
+    Register-PSRepository -Name $Name `
```
WilliamBerryiii: **High — the missing-PSGallery recovery uses a parameter set that PowerShellGet rejects.** PowerShellGet 2.2.5 reserves `PSGallery` for the default parameter set; invoking `Register-PSRepository -Name PSGallery ...` throws `UseDefaultParameterSetOnRegisterPSRepository`, so this path stops before `Install-Module` in the exact condition it is meant to recover from. I reproduced this against the installed implementation; the Pester test passes only because the unrestricted mock accepts the invalid invocation. Please use `Register-PSRepository -Default -ErrorAction Stop` with only the policy options intentionally required, and make the test assert the `-Default` parameter set. A bounded contract or integration test that models the reserved-name rejection would prevent this mock/implementation mismatch from recurring.

scripts/security/Install-PSModules.ps1
```diff
@@ -144,6 +144,37 @@ function Test-ModulePresent {
...
+        -SourceLocation 'https://www.powershellgallery.com/api/v2' `
+        -InstallationPolicy Trusted `
```
WilliamBerryiii: **Medium — this trust decision persists outside the installer.** `InstallationPolicy Trusted` changes user repository configuration and suppresses trust prompts for later PowerShellGet operations; it is not scoped to this install and is not package-integrity verification. The canonical HTTPS endpoint and exact module versions reduce exposure, but persistent developer and runner accounts retain the policy. Please decide whether durable trust is required. Otherwise avoid setting it, or capture whether this function created or changed the repository and restore the prior policy in `finally`. Document the persistent mutation under installer side effects as well.

.github/skills/project-planning/gitlab/scripts/gitlab.py
```diff
@@ -928,9 +933,13 @@ def _request_bytes(
                 )
                 if require_json and result.strip():
                     if not content_type:
-                        die("unexpected Content-Type: <missing>", EXIT_FAILURE)
+                        raise GitLabError(
```
WilliamBerryiii: **Medium — response and protocol validation bypasses the API-specific error type.** Missing or invalid `Content-Type` reaches this request boundary but raises the base `GitLabError`; the oversized JSON path at `_read_capped()` does the same. Runtime probes confirmed all three concrete exception types. This makes API response failures indistinguishable from ordinary CLI validation failures and only partially satisfies the typed hierarchy API-failure contract. Please raise `GitLabAPIError` with the method and query-stripped resource context for these paths. For the size-limit case, either construct the API error in `_request_bytes()` or pass enough request context into `_read_capped()`, and tighten tests to require the subtype.

.github/skills/project-planning/gitlab/tests/test_gitlab_helpers.py
```diff
@@ -83,15 +83,63 @@ def _network_invocations(source: str, module: str) -> list[tuple[str, str, str]]
...
+            continue
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)) and (
+            node.name == "die"
+        ):
+            violations.append("def die")
+        elif isinstance(node, ast.Raise):
+            raised = node.exc
+            if isinstance(raised, ast.Call):
+                raised = raised.func
+            if isinstance(raised, ast.Name) and raised.id == "SystemExit":
```
WilliamBerryiii: **Medium — this semantic exclusivity contract can be bypassed by normal Python aliases.** The detector recognizes only bare `SystemExit` and a literal `sys.exit` attribute chain. Direct probes showed that `builtins.SystemExit`, `from sys import exit; exit(...)`, and `import sys as system; system.exit(...)` all produce no violations. Production does not currently use those forms, so this is a regression-detection gap rather than a present alternate exit. Please resolve import bindings for `sys`, `sys.exit`, `builtins`, and `builtins.SystemExit`, or replace this with a simpler invariant that covers every disallowed form. Add negative fixtures for the qualified and aliased cases so the test proves the claimed single-failure mechanism.

WilliamBerryiii (changes requested):
Comprehensive review found one blocking defect and three material follow-ups. The GitLab typed-failure migration is coherent, centralizes emission correctly, preserves redaction, and passed focused execution. I am requesting changes because the newly added PSGallery recovery fails in the exact missing-repository condition it is intended to handle.

**Findings:**
- **High:** PowerShellGet rejects custom registration under the reserved `PSGallery` name; use the `-Default` registration flow.
- **Medium:** API response-validation failures use `GitLabError` instead of the API-specific `GitLabAPIError`.
- **Medium:** The source contract misses aliased and qualified exit mechanisms.
- **Medium:** `InstallationPolicy Trusted` persists beyond this installer invocation unless explicitly restored or avoided.

Line comments contain the reproduction details and suggested corrections. Additional non-blocking observations concern PR scope, test readability, PowerShell help and naming, repeated mocks, and documenting repository mutation as a side effect.

Validation included 198 focused GitLab tests, 29 focused PowerShell tests, Ruff lint and format checks, PSScriptAnalyzer across 253 files, and an implementation-level PowerShellGet probe confirming the blocking parameter-set behavior. No accessibility barriers were found in the changed surfaces.

scripts/security/Install-PSModules.ps1
jkim323: Thanks for catching the PowerShellGet parameter-set mismatch. Addressed in 34a17123:

- Missing PSGallery recovery now uses `Register-PSRepository -Default -ErrorAction Stop` through a narrow helper.
- Focused coverage requires that exact invocation and verifies the missing-repository branch delegates once without proxying PowerShellGet's dynamic parameters.
- Validation passed: 30/30 focused Pester tests, 0 targeted PSScriptAnalyzer findings, clean editor diagnostics, and `git diff --check`.

scripts/security/Install-PSModules.ps1
jkim323: Addressed in 34a17123. The recovery path no longer sets `InstallationPolicy Trusted` or supplies a custom PSGallery endpoint; it restores PowerShellGet's default repository with `Register-PSRepository -Default -ErrorAction Stop`, avoiding a persistent trust-policy mutation.

Validation passed: 30/30 focused Pester tests, 0 targeted PSScriptAnalyzer findings, clean editor diagnostics, and `git diff --check`.

.github/skills/project-planning/gitlab/scripts/gitlab.py
jkim323: Thanks for catching this. Addressed in 0f82ed13:

- Missing and invalid `Content-Type` responses now raise `GitLabAPIError`.
- Oversized JSON responses are translated to `GitLabAPIError` at `_request_bytes()`, preserving `_read_capped()` as a generic helper.
- All three paths include the HTTP method, query- and fragment-stripped resource, and safe request ID.
- Regression coverage requires the API subtype and verifies that query and fragment data are not rendered.

Validation passed: 3 focused pytest cases, 32/32 affected-boundary tests, Ruff check and format, editor diagnostics, and `git diff --check`.

jkim323:
@copilot+gpt-5.6-sol:high The `npm Security Audit` check is failing on this PR. The failure originates from vulnerable transitive dependencies rather than the GitLab refactor.

Known findings include:

### `ajv > fast-uri`

- GHSA-f65p-4m7j-42xc
- GHSA-fph4-wmhf-6fwf
- GHSA-jqff-g426-hqxp
- GHSA-5jgf-p345-68v8

All four are fixed in the v3 release line by `fast-uri` 3.1.6. The repository currently pins `fast-uri` to vulnerable version `3.1.5`.

### `qs`

- GHSA-x5fp-wj9c-mxmx
- GHSA-4mjr-xmp4-gh2g

Both are fixed by `qs` 6.16.0. The root lockfile currently resolves vulnerable `qs` 6.15.2.

Please remediate every affected npm project in the repository, not only the first audit lane that fails:

1. Inspect all dependency paths and lockfiles:

   ```bash
   npm ls ajv fast-uri qs
   find . -name package-lock.json -not -path '*/node_modules/*' -print
   ```

2. Update every `fast-uri` override from `3.1.5` to at least `3.1.6`, including the root project and `docs/docusaurus`.

3. Resolve every installed copy of `qs` to at least `6.16.0`, preferably by updating its direct parent; use a narrow npm override where an upstream dependency cannot yet select the patched version.

4. Regenerate every affected `package-lock.json`, including:
   - `package-lock.json`
   - `docs/docusaurus/package-lock.json`
   - `scripts/extension/marketplace-publisher/package-lock.json` if its installed `fast-uri` or `qs` remains vulnerable

5. Do not add these advisories to `audit-ci.json`, because patched versions are available.

6. Do not stop after resolving the listed advisories. Run fresh audits and remediate any additional non-allowlisted vulnerability reported by the current npm advisory database.

Verify from clean installs:

```bash
npm ci
npm run audit:npm:root

npm ci --prefix docs/docusaurus
npm run audit:npm:docs

npm ci --prefix scripts/extension/marketplace-publisher
npm audit --prefix scripts/extension/marketplace-publisher

npm run audit:npm
```

The final dependency trees must contain no `fast-uri` version below the applicable patched release and no `qs` version below `6.16.0`:

```bash
npm ls fast-uri qs --all
npm ls fast-uri qs --all --prefix docs/docusaurus
npm ls fast-uri qs --all --prefix scripts/extension/marketplace-publisher
```

Keep the change limited to npm manifests and generated lockfiles unless another file is strictly required.

Failing job: https://github.com/microsoft/hve-core/actions/runs/33702622528/job/100484971614?pr=2779

.github/skills/project-planning/gitlab/tests/test_gitlab_helpers.py
jkim323: Thanks for catching this alias gap. 

- The source contract now resolves direct import bindings for `sys`, `sys.exit`, `builtins`, and `builtins.SystemExit`.
- Binding visibility follows Python lexical scopes without leaking aliases between sibling scopes or from class bodies into methods.
- Added six prohibited alias cases and two allowed isolation cases.
- Preserved the canonical `__main__` guard exemption, existing violation labels, and unrelated local or object `exit` behavior.

Validation passed for the implementation and extended semantic matrices, Ruff check, Ruff format, editor diagnostics, and `git diff --check`. Focused and full pytest could not start because the existing local venv contains a partial pytest installation without `pytest.__main__`; CI should confirm the complete suite.

WilliamBerryiii (approved):

PR #2796, opened 2026-08-29
fix(skills): harden accessibility runtime target and artifact validation

### LINKED ISSUE #2786: fix: harden accessibility runtime target validation
The accessibility skill exposes two scanner paths whose target, navigation, artifact-identifier, and browser-launch contracts are not consistently enforced at their authoritative boundaries. This work hardens those contracts while preserving supported local-file scanning and the existing system Chrome requirement.

The change will add explicit HTTP(S) validation for runtime base targets, same-origin trigger navigation, bounded redirect handling for the broken-link direct-request helper, portable artifact identifiers, Path A target classification and external-host authorization, and launch-bound calibration readiness. It will also update operator guidance and the accessibility security model to describe both scanner paths and their residual browser-network risks accurately.

## Acceptance Criteria

- [ ] Runtime base targets accept only absolute HTTP(S) URLs with a host and no credentials before host authorization and browser launch.
- [ ] Navigate and visit triggers accept relative or absolute same-origin HTTP(S) destinations and reject origin or scheme escapes before navigation.
- [ ] Broken-link direct requests manually enforce a bounded same-origin redirect chain without claiming browser-wide egress control.
- [ ] Path-bearing machine identifiers use one portable rejection-based contract before artifact creation.
- [ ] Path A preserves explicit regular local-file scanning, rejects remote or ambiguous file targets, and requires explicit authorization for non-loopback HTTP(S).
- [ ] Generic Chrome launch arguments and detached calibration executable overrides are removed; calibration evidence comes from the Playwright system-Chrome launch.
- [ ] Operator guidance and the accessibility security model describe compatibility changes, local-file authority, and redirect, DNS, browser-request, endpoint-Chrome, and inherited-environment residuals.
- [ ] Focused tests, full skill coverage, documentation generation, and locally safe repository validation pass.

## Non-Goals

- Full Playwright request interception or DNS/IP pinning
- Cross-origin trigger authorization
- Child-environment filtering
- Browser acquisition or Path A deprecation

WilliamBerryiii:
## Description

The accessibility skill exposes two scanner paths, and neither one enforced its target contract at an authoritative boundary. **Path A** (`scripts/scan.py`, the axe CLI wrapper) passed any operator string straight through to `npx`. **Path B** (`scripts/runtime_a11y/` and its Node runners, the Playwright runtime) trusted configuration values and trigger destinations all the way to `page.goto` and to filesystem path construction. This change moved validation to the earliest boundary on both paths and rewrote the skill security model to describe them accurately.

> Both paths keep the capabilities they already supported. Local-file scanning still works on Path A, and Path B still drives endpoint-managed system Chrome through Playwright's `channel: 'chrome'`.

### Runtime target and navigation validation

A new *validation.mjs* module became the single grammar for the Node runners. `assertHttpUrl` rejects anything that is not an absolute HTTP(S) URL with a hostname — non-matching schemes, embedded control characters, unparseable values, and credential-bearing URLs. `assertArtifactId` constrains identifiers to `^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$`.

- Added `resolveDestinationUrl` to *route.mjs*, which resolves a trigger destination against the configured base and throws when the result leaves that origin. It rejects protocol-relative prefixes, non-HTTP(S) schemes, whitespace, and credentials before resolution.
  - `applyTrigger` in *_shared.mjs* now precomputes the destination through that resolver **before** the action `switch`, so both `navigate` and `visit` reject an origin escape ahead of any navigation.
  - `navigatePageWithRetry` in *at-plan-executor.mjs* asserts the AT-plan target URL as its first statement, and both trigger loops thread `baseUrl` through.
- `runProbeWithPage` asserts the runtime base URL immediately after destructuring the runtime context, ahead of surface lookup.
- *config-schema.json* gained an `artifactId` definition applied by `$ref` to every identifier-bearing field, and `baseUrl` picked up a case-insensitive `^https?://` pattern, so the constraint is enforced at the schema layer as well as at runtime.
- *_config.py* split parsing from policy. `assert_host_allowed` became a standalone function, and a new `_validated_http_host` rejects bad schemes, missing hosts, control characters, credentials, and invalid ports. Scheme and credential rejection now runs **before** the authorization check, so `allow_external` cannot bypass it.

### Path A target classification and host authorization

*scan.py* gained `--allow-host` (repeatable) and `--allow-external`, and a `resolve_scan_target` classifier that runs before the scanner starts.

- Loopback HTTP(S) needs no option; every other remote host requires explicit authorization.
- Local paths and local `file:` URIs must resolve to an existing **regular file**, canonicalized through `Path.resolve().as_uri()`.
- Rejected up front: leading-dash targets, network shares, protocol-relative values, remote `file:` authorities, directories, missing files, embedded credentials, invalid ports, and non-HTTP(S) remote schemes.
- The axe invocation became `npx --yes @axe-core/cli@4.12.1 -- <target>`, adding a `--` parser boundary alongside the existing argument-list (no shell) invocation.

### Bounded redirect handling

*probe-broken-links.mjs* stopped delegating redirects to the request layer. `checkLinkWithRedirects` now walks at most five hops itself with `maxRedirects: 0` on every request, validating each hop through the same `resolveDestinationUrl` used by triggers, and returns structured reasons — `redirect-limit`, `redirect-location-missing`, `redirect-policy`, `redirect-loop`, `request-failed`. A `HEAD` failure falls back to `GET`.

### Portable artifact identifiers

Artifact paths moved off string mangling. *visual-review-executor.mjs* replaced the private `buildRouteSlug` regex sanitizer with an exported `buildVisualReviewArtifactSegment` that validates the surface and state identifiers and returns them as separate path segments. *_shared.mjs* did the same for trace artifacts, validating `probeId`, `surfaceId`, and `state` and joining them as three `path.join` arguments instead of one interpolated string.

> The old flat hyphen-joined naming could collapse distinct inputs onto the same directory. A test now pins that `home-search`/`default` and `home`/`search-default` stay separable.

### Launch-bound Chrome readiness

*calibration-executor.mjs* lost roughly 130 lines of Chrome discovery — `resolveChromeExecutable` and its version probes, the `RUNTIME_A11Y_CHROME_PATH` / `CHROME_BIN` / `PLAYWRIGHT_CHROME_EXECUTABLE_PATH` / `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH` overrides, the Windows `ProgramFiles` fallbacks, and the `where.exe` and `powershell.exe` spawns used to read version metadata. `probePlaywrightChrome` replaced all of it by launching through the shared launcher and reading `browser.version()`, reporting `channel:chrome` as the executable. A browser that fails to close is now a prerequisite failure rather than a soft warning. Separately, *_shared.mjs* dropped the loop that merged caller-supplied `args` into the Chrome launch options.

### Documentation

- Rewrote the skill's *SECURITY.md*. Scope widened from `scan.py` alone to both scanner paths plus generated evidence and offline design-intent verification, trust buckets went from five to eight, and every bucket kept all six STRIDE categories with a risk-rating table. Open residual gaps moved from 4 to 7, adding `G-SUP-2` (endpoint-owned Chrome identity), `G-TAM-2` (Playwright/Chrome parser surface), and `G-INF-3` (inherited child environment), and re-scoping `G-INF-2` to Medium.
  - The rewrite states explicit non-claims rather than blanket mitigation: no complete Chromium egress control, no DNS-rebinding resistance, local files not repository-confined.
  - Path A local-file scanning is carved out of `G-INF-1` as an operator-authorized capability rather than folded into the SSRF gap.
- Updated *SKILL.md* with the `--allow-host` and `--allow-external` surface, a local-file example, the rejected target shapes, a note that remote authorization binds only the initial target, and two exit-code-`2` troubleshooting rows.
- Updated the `accessibility` row of *docs/security/security-model.md* to describe Path A and Path B separately with matching gap identifiers.

### Tests

Added *probe-broken-links.test.mjs* covering the redirect walker across six scenarios, including assertions on the recorded call list proving a blocked request was never issued. Extended the existing runner suites with direct coverage for `assertHttpUrl`, `assertArtifactId`, `resolveDestinationUrl`, `buildVisualReviewArtifactSegment`, off-origin trigger rejection with `navigationCount === 0`, and the `channel:chrome` probe paths. On the Python side, *test_config.py* gained parametrized coverage for unsupported `baseUrl` schemes and invalid artifact identifiers, *test_scan.py* covers `resolve_scan_target` with eight parametrized rejection cases plus the `--` separator, and *test_scan_branches.py* was amended for the new default-deny behavior.

Error messages are part of the tested contract throughout — assertions pin `redirect-policy`, `Surface ID`, `must not include credentials`, `--allow-host HOST`, and `configured origin`.

## Related Issue(s)

Fixes #2786

## Type of Change

Select all that apply:

**Code & Documentation:**

* [x] Documentation update

* [ ] GitHub Actions workflow
* [ ] Security configuration
* [ ] Dependency update

**AI Artifacts:**

>

**Other:**

**User Request:**

> Scan `https://staging.example.com` for accessibility issues with the accessibility skill.

**Execution Flow:**

1. The skill routes the request to Path A (`scripts/scan.py`) for a single-URL scan.
2. `resolve_scan_target` classifies the target. Because `staging.example.com` is not loopback, the scan stops with exit code `2` and a message naming `--allow-host HOST` or `--allow-external`.
3. The operator re-runs with `--allow-host staging.example.com`. The target passes classification, and the scanner invokes `npx --yes @axe-core/cli@4.12.1 -- https://staging.example.com`.
4. For a multi-surface run, Path B loads `a11y-runtime.json`, validates it against *config-schema.json*, authorizes the `baseUrl` host in `_config.py`, then hands off to the Node runners. Each runner reasserts the base URL, resolves every route and trigger destination same-origin, and validates artifact identifiers before writing evidence.

**Output Artifacts:**

Path B evidence is written beneath the run root with nested, validated segments:

```text
<runRoot>/artifacts/<surfaceId>/<state>/trace.zip
<runRoot>/artifacts/<surfaceId>/<state>/screenshot.png
<runRoot>/preflight-summary.json
```

Probe outcomes carry a composed identifier, for example `home:desktop`.

**Success Indicators:**

* An unauthorized remote host stops with exit code `2` and names the option that would authorize it, rather than scanning.
* A trigger pointed off-origin throws with a message naming the configured origin, and the navigation counter stays at `0`.
* A surface or state identifier containing `/`, a space, or a `..` segment is rejected before any directory is created.
* Calibration reports `channel:chrome` with a browser-reported version, sourced from an actual Playwright launch rather than a filesystem probe.

## Testing

Automated validation run locally on this branch. Link-check lanes were skipped at the author's request, so the `npm run validate:local` aggregate was not run as a single command; its non-link lanes relevant to this diff were run individually instead.

| Check                          | Result                                                             |
|--------------------------------|--------------------------------------------------------------------|
| `npm run lint:md`              | Passed                                                             |
| `npm run lint:tables`          | Passed                                                             |
| `npm run lint:frontmatter`     | Passed                                                             |
| `npm run lint:json`            | Passed                                                             |
| `npm run lint:py`              | Passed                                                             |
| `npm run validate:skills`      | Passed                                                             |
| `npm run lint:ai-artifacts`    | Passed                                                             |
| `npm run lint:models`          | Passed                                                             |
| `npm run lint:plugin-manifest` | Passed                                                             |
| `npm run lint:asset-docs`      | Passed                                                             |
| `npm run lint:design-intent`   | Passed                                                             |
| `npm run spell-check`          | Passed (823 files, 0 issues)                                       |
| Node runner tests              | Passed (95/95 across the six affected suites)                      |
| Skill Python tests             | 383 passed, 49 failed — all in unmodified `test_intent.py` (below) |
| `npm run lint:links`           | Skipped at author request                                          |
| `npm run lint:md-links`        | Skipped at author request                                          |

The 49 Python failures are confined to `tests/runtime_a11y/test_intent.py`, which this PR does not modify. Those tests exercise POSIX-only filesystem primitives (`dir_fd`, no-follow opens, `os.mkfifo`) and do not run on Windows. Every test file this PR touches passed. The repository-wide 95% coverage gate was not exercised locally for the same reason and is left to CI.

Security analysis of the diff found no sensitive data, no dependency manifest or lockfile changes, and no non-compliant language. The change narrows privilege rather than widening it: it removes environment-driven executable overrides, removes `where.exe` and `powershell.exe` spawns that interpolated a discovered path, adds default-deny host authorization on both paths, and disables automatic redirect following. No file under `scripts/security/` was touched.

Manual testing was not performed.

### Required Checks

* [ ] Changes are backwards compatible (if applicable) — see Additional Notes for the three intentional behavior changes

### AI Artifact Contributions

### Required Local Checks

* [ ] Local validation aggregate: `npm run validate:local` — not run as an aggregate because it chains both link-check lanes; the relevant non-link lanes were run individually and passed (see Testing)
* [ ] Documentation validation (if docs changed): `npm run validate:docs` — not run; no Docusaurus site content changed beyond the security-model summary row
* [ ] Link validation: `npm run lint:md-links` — skipped at author request

## Security Considerations

* [ ] Any new dependencies have been reviewed for security issues (N/A — no dependency manifests or lockfiles changed)
* [ ] Security-related scripts follow the principle of least privilege (N/A — no files under `scripts/security/` changed)

## Additional Notes

Three intentional behavior changes affect existing callers. All three are documented in the skill's *SKILL.md* and *SECURITY.md*.

1. **Non-loopback Path A targets now require authorization.** A scan against any host other than loopback needs `--allow-host HOST` or `--allow-external`. Previously any URL was accepted.
2. **Visual-review artifact layout changed.** Directories moved from flat hyphen-joined names to nested `surfaceId/state` paths, and `probeOutcomes[].id` changed from `<routeSlug>-<stateSlug>` to `<surfaceId>:<state>`.
3. **Chrome launch overrides were removed.** Caller-supplied `args` are ignored, and the `RUNTIME_A11Y_CHROME_PATH`, `CHROME_BIN`, `PLAYWRIGHT_CHROME_EXECUTABLE_PATH`, and `PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH` environment overrides no longer exist. Calibration evidence now comes from the Playwright system-Chrome launch.

Skill membership is unchanged — *SKILL.md* and *SECURITY.md* were modified rather than added or removed, so no `plugin.json` update is implied. `npm run lint:plugin-manifest` confirms no drift.

The 95% coverage gate and the `ci:test:a11y:smoke` lane are left to CI, since the coverage run cannot complete on Windows while `test_intent.py` requires POSIX filesystem primitives.


.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/at-plan-executor.mjs
```diff
@@ -96,6 +97,7 @@ function isRetryableNavigationError(error, retryableErrorPatterns) {
 }
 
 async function navigatePageWithRetry({ page, context, targetUrl, runtimeConfig }) {
+  assertHttpUrl(targetUrl, 'AT-plan target URL');
```
jkim323, line 100: **[Functional] Validate AT-plan targets before launching Chrome**

The new target check is inside navigatePageWithRetry, but processAtPlanCase launches or obtains the browser before preparePageLifecycle calls this function. An invalid file:, credential-bearing, or otherwise unsupported matrix target therefore still launches system Chrome before being rejected. This directly violates the linked issue's validate-before-launch contract and leaves a regression path where invalid input causes browser side effects and resource consumption. The added test checks that navigationCount stays zero but does not check that browser launch stays zero, so it passes despite the contract violation.

**Suggested fix:**
Validate resolvedTarget immediately after it is computed in processAtPlanCase and before either browserFactory or launchChrome can run; retain the navigatePageWithRetry assertion as defense in depth. Extend the unsupported-target test with a browserLaunches counter and assert that it remains zero.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/visual-review-executor.mjs
```diff
@@ -199,10 +201,9 @@ export async function captureVisualReviewEvidence(config = {}) {
     for (const route of selectedRoutes) {
       for (const stateEntry of selectedStates) {
         const stateName = typeof stateEntry === 'string' ? stateEntry : stateEntry.state;
-        const routeSlug = buildRouteSlug(route);
-        const stateSlug = String(stateName).replace(/[^a-zA-Z0-9._-]+/g, '-');
-        const artifactSegment = [routeSlug, stateSlug].join('-') || 'root';
-        const artifactDir = path.join(runRoot, 'artifacts', artifactSegment);
+        const artifactSegments = buildVisualReviewArtifactSegment(route, stateName);
+        const artifactDir = path.join(runRoot, 'artifacts', ...artifactSegments);
+        const outcomeId = artifactSegments.join(':');
```
jkim323, line 218: **[Functional] Keep visual-review artifacts unique across routes**

The new artifact directory and outcome ID contain only surfaceId and stateName. The visual-review plan permits multiple distinct routes to share a surface ID and state (and routes without an explicit surface ID all fall back to 'surface'). Those runs now write screenshot.png, measurements.json, and trace.zip into the same directory, so later routes silently overwrite earlier evidence while multiple manifest entries point to the same files and outcome ID. The previous route-derived segment distinguished those routes; the replacement removes that dimension without enforcing uniqueness.

**Suggested fix:**

Add a required, portable route artifact ID and include it in artifactSegments (for example [routeId, surfaceId, stateId]), or reject duplicate surfaceId/stateId pairs before creating any artifact. Add a regression test with two different route paths sharing one surfaceId and state and verify that their paths and outcome IDs are distinct and neither evidence set is overwritten.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/config-schema.json
```diff
@@ -128,11 +129,11 @@
             "type": "object",
             "required": ["id"],
             "properties": {
-              "id": { "type": "string" },
+              "id": { "$ref": "#/$defs/artifactId" },
```
jkim323, line 132: **[Security] Unvalidated journeyId alias escapes the calibration run root**

The changed schema validates calibration journey `id`, but each journey still allows additional properties and the calibration executor preferentially consumes `journeyId` (`journey?.journeyId || journey?.id`). A malicious repository can therefore supply a schema-valid journey such as `{"id":"safe","journeyId":"../../../../outside"}`. When an operator runs calibration, `persistSampleEvidence` joins that value beneath `runRoot/journeys` and `persistJsonArtifact` resolves, creates, and writes the resulting path without a containment check. This lets project-controlled configuration create or overwrite `0/evidence.json` outside the authorized run root with the operator's filesystem authority.

**Suggested fix:**
Remove the `journeyId` alias from the executor and use only the validated `id`, or declare `journeyId` with the same artifactId schema and reject configurations where it differs from `id`. Independently assert the effective journey identifier immediately before path construction and verify the resolved evidence path remains beneath the resolved run root before mkdir or write. Add a schema-valid `{ id: 'safe', journeyId: '../../escape' }` regression test that proves no path outside the run root is touched.

.github/skills/accessibility/accessibility/scripts/scan.py
```diff
@@ -110,9 +127,101 @@ def normalize_results(raw_results: dict[str, Any], target: str) -> dict[str, Any
...
+            "Remote scan target must not include whitespace or control characters",
+            EXIT_USAGE,
+        )
+
+    parsed = urlparse(target)
+    scheme = parsed.scheme.lower()
+    if scheme == "file":
+        if parsed.netloc and parsed.netloc.lower() != "localhost":
+            raise ScriptError("Remote file authorities are not supported", EXIT_USAGE)
+        return _canonical_local_file(Path(url2pathname(parsed.path)))
```
jkim323, line 183: **[Security] Local file URI validation permits Windows UNC access**

The new local-file authority check trusts an empty or `localhost` URL authority, then converts the path component with the platform-specific `url2pathname`. On Windows, `file:////attacker.example/share/page.html` has an empty `netloc` and a `//attacker.example/share/page.html` path, while `file://localhost//attacker.example/share/page.html` has an allowed authority and the same path. Both convert to a UNC path. `_canonical_local_file` then calls `exists()` and `is_file()`, causing an outbound SMB access before the scanner starts; an attacker-controlled share can elicit Windows authentication material even if the scan later fails. This bypasses the changed code's explicit network-share rejection and contradicts the documented local-only authority boundary.

**Suggested fix:**
Reject network-shaped paths at the authoritative filesystem boundary, before `exists()`, so this also protects future callers and percent-decoded variants. Suggest to add Windows-focused tests for `file:////host/share`, `file://localhost//host/share`, device namespaces, and percent-encoded separator variants, with the filesystem probe mocked to prove rejection occurs before access.

jkim323 (approved):
Strong work across this PR! The changes establish clearer validation boundaries, require explicit remote-host authorization, constrain navigation and redirects, strengthen artifact identifiers, and tie Chrome readiness to the Playwright-managed browser. The focused tests and expanded security model further improve confidence and clearly document residual risks. The remaining findings are targeted follow-ups to reinforce this solid foundation. Know you will address them accordingly thus approved

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-broken-links.mjs
```diff
@@ -1,6 +1,74 @@
...
+
+const REDIRECT_STATUSES = new Set([301, 302, 303, 307, 308]);
+
+async function requestWithoutRedirects(request, url, timeout) {
+  const options = { timeout, maxRedirects: 0 };
+  try {
+    return await request.head(url, options);
+  } catch {
+    return request.get(url, options);
+  }
```
rezatnoMsirhC, line 25: **Medium severity · High confidence**

The focused reproduction confirmed that a HEAD 405 response does not trigger the GET fallback. This could report a valid GET-only endpoint as broken. Retry with GET for 405/501 and add a 405-to-200 regression test.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/_shared.mjs
```diff
@@ -954,6 +955,7 @@ export async function readLiveRegionSnapshot(page, { settleMs = 500 } = {}) {
 export async function runProbeWithPage(callback) {
   const contextData = getRuntimeContext();
   const { config, probeId, surfaceId, state, baseUrl, trace } = contextData;
+  assertHttpUrl(baseUrl, 'Runtime base URL');
   const surface = (config.surfaces || []).find((entry) => entry.id === surfaceId) || null;
   const targetUrl = resolveTargetUrl(baseUrl, surface);
```
rezatnoMsirhC, line 960: **High severity · High confidence**

`assertHttpUrl` only validates URL shape. The traced `run-all` and `probe` path authorizes `config.baseUrl`, then replaces it with `args.base_url`, which can allow a non-loopback override to reach this browser path without the allowlist or `--allow-external`. Guard the effective URL after applying the override and add command-level tests.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/validation.mjs
```diff
@@ -0,0 +1,36 @@
...
+
+const ARTIFACT_ID_PATTERN = /^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$/;
+
+export function assertArtifactId(rawValue, label = 'Artifact ID') {
+  if (typeof rawValue !== 'string' || !ARTIFACT_ID_PATTERN.test(rawValue)) {
+    throw new Error(
+      `${label} must be 1-128 ASCII letters, numbers, dots, underscores, or hyphens and start with a letter or number.`,
+    );
+  }
+  return rawValue;
```
rezatnoMsirhC, line 49: **Medium severity · Medium confidence**

The portable-ID pattern appears to accept trailing-dot aliases and Windows device names such as `CON`, `NUL`, `COM1`, and `LPT1`. These values could cause Windows evidence paths to collide or fail. Consider rejecting them in both runtime and config validation.

rezatnoMsirhC (dismissed):
Thank you for the thorough boundary hardening and test coverage. I found three issues that appear worth addressing before merge: with high confidence, `run-all` and `probe` can execute an unauthorized `--base-url` override; with medium confidence, the portable artifact-ID grammar could cause Windows path collisions or failures; and with high confidence, the broken-link probe does not fall back to GET after HEAD returns 405 or 501, which could report valid GET-only endpoints as broken. The inline comments include exact anchors, evidence, and suggested tests.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/config-schema.json
WilliamBerryiii: Fixed in `3bfff64c`. Confirmed the full chain: the schema validates `id` but leaves `journeys[]` as `additionalProperties: true`, and `normalizeJourney` preferred `journeyId` over `id`, so a traversal payload reached `persistSampleEvidence` → `persistJsonArtifact`, which resolved and `mkdirSync`'d with no containment check.

Worth noting the containment primitive already existed in this file — `isWithinRoot` was wired only to `resolveContainedArtifactPath` on the **read** path. Reads were contained; writes were not.

Applied defense in depth at all three layers you identified:

1. **Write path** — `persistSampleEvidence` now resolves the run root and asserts `isWithinRoot` before `persistJsonArtifact` can create any directory. This protects future callers, not just this one.
2. **Identifier selection** — `normalizeJourney` now prefers the schema-validated `id` and asserts the result with `assertArtifactId`.
3. **Schema** — `journeyId` is declared with the `artifactId` grammar.

Regression tests added: the alias is ignored when `id` is present, a traversal identifier is rejected from either field, and `persistSampleEvidence` refuses to write outside the run root leaving no escaped path. Plus a schema-level Python test for `{ id: 'safe', journeyId: '../../escape' }`.

Did not add `additionalProperties: false` — that would break forward compatibility for unrelated fields and is wider than the defect warrants.

.github/skills/accessibility/accessibility/scripts/scan.py
WilliamBerryiii: Fixed in `3bfff64c`. Verified this empirically before fixing rather than reasoning about it — both forms do convert to UNC on Windows:

| Input | `netloc` | `url2pathname(path)` |
|---|---|---|
| `file:////attacker.example/share/page.html` | `''` | `\\attacker.example\share\page.html` |
| `file://localhost//attacker.example/share/page.html` | `'localhost'` | `\\attacker.example\share\page.html` |

Root cause was that the existing guard tested the **raw input string** (`target.startswith(("\\\\", "//"))`), which a `file:`-scheme target never matches, so it did not survive URI decoding.

Took your suggestion to move the check to the filesystem boundary: added `_is_network_path` and made `_canonical_local_file` reject network-shaped **resolved** paths as its first statement, before `exists()`. That covers the `file:` route, the direct-path route, and any future caller in one place. Rejected the alternative of extending the `startswith` tuple, since it would still test the input string.

Tests assert both the raised error and that no network-shaped path ever reached a filesystem probe, plus a direct `_canonical_local_file` test asserting zero probes.

One note on your percent-encoded suggestion: I tested it, and `file://localhost/%2F%2Fattacker.example/share/p.html` decodes to `\//attacker.example\share\p.html`, which is not a UNC path — so it is not a live vector. Good instinct to check it though; it is covered by the resolved-path check regardless.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/at-plan-executor.mjs
WilliamBerryiii: Fixed in `3bfff64c`. You were right on both counts, and the second one was the more valuable catch.

The ordering: `resolvedTarget` is computed at `:556`, the browser is obtained at `:702`/`:717`, and `preparePageLifecycle` at `:730` is what eventually reaches the assertion. Launch strictly preceded validation.

Moved the effective assertion to the top of the `try` block in `processAtPlanCase` — after `resolvedTarget` is computed, before the launch branch, and inside the error-handling scope so it still reports as a case error rather than an unhandled throw. Kept the `navigatePageWithRetry` assertion as defense in depth.

On the test: your point that it "passes despite the contract violation" was exactly right, and I verified it rather than taking it on faith. With the new assertion temporarily removed, the corrected test fails with `browserLaunches` actual `1`, expected `0` — proving the browser really did launch and that the old `navigationCount === 0` assertion could never have caught this. Assertion restored, suite green at 99/99.

The test now counts browser launches and asserts zero, keeping the navigation assertion alongside it.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/visual-review-executor.mjs
WilliamBerryiii: Fixed in `3bfff64c`. Fair characterisation — this PR removed one collision mode (delimiter ambiguity in the flat hyphen-joined slug) and introduced another by dropping the route dimension. The route segment was collateral of the identifier work, not an intended change.

Also confirmed your point about reachability: because routes without an explicit `surfaceId` all fall back to the literal `'surface'`, two such routes collided **by default** rather than requiring an unusual plan.

Took your first suggested option. Added `buildRouteArtifactId` and `buildVisualReviewArtifactSegment` now returns `[routeId, surfaceId, stateId]`. `routeId` is derived-with-fallback — explicit `routeId`/`id` if present, otherwise a portable form of the route path, otherwise `root` — and validated by `assertArtifactId` like the other segments. Deriving rather than requiring keeps existing plans valid.

Rejected the duplicate-rejection alternative because it converts a currently valid plan shape into a hard error, which is broader than the defect.

Adding a dimension does not reintroduce the ambiguity this PR removed: segments stay separate `path.join` arguments, and only the reserved `:` join composes the outcome ID.

Regression test covers two routes with distinct paths sharing one `surfaceId` and state, plus anonymous routes, an explicit `routeId`, and the root fallback. Artifact directories are now `artifacts/<routeId>/<surfaceId>/<state>/`; I will update the PR description so it describes the final layout rather than the intermediate one.

WilliamBerryiii:
## Review findings resolved — `3bfff64c`

All four findings from @jkim323's review are fixed. Each was independently confirmed against source before fixing; one was confirmed by execution. Detailed replies are on the individual threads.

Three of the four shared a single root cause: **a control existing on the wrong side of the boundary it protects.**

| Finding | Root cause | Fix |
|---|---|---|
| Calibration run-root escape (Security) | Containment helper wired to the read path only; executor preferred the unvalidated `journeyId` alias | Containment asserted in `persistSampleEvidence` before any `mkdirSync`; `id` preferred and validated; schema constrains `journeyId` |
| Windows UNC via `file:` URI (Security) | Guard tested the raw input string, so it did not survive URI decoding | Network-shaped **resolved** paths rejected inside `_canonical_local_file` before `exists()` |
| AT-plan validated after launch (Functional) | Guard at the navigation boundary; contract is a launch boundary | Assertion moved ahead of `browserFactory`/`launchChrome`; original retained as defense in depth |
| Visual-review artifact collision (Functional) | Route dimension dropped as collateral of the identifier work | Segments restored to `[routeId, surfaceId, stateId]`, `routeId` derived-with-fallback |

### Falsification check

The AT-plan test was verified **non-vacuous**, since the original finding was that my previous test passed despite the defect. With the new assertion temporarily removed, the corrected test fails with `browserLaunches` actual `1`, expected `0` — proving the browser did launch and that the old `navigationCount === 0` assertion could never have caught it. Assertion restored, suite green.

### Corrections to the PR description above

The description was written before these fixes and is stale in two places. My token cannot edit the PR body (Enterprise Managed User restriction on both the GraphQL and REST update endpoints), so recording the corrections here:

* **Artifact layout** is now `<runRoot>/artifacts/<routeId>/<surfaceId>/<state>/`, not `<surfaceId>/<state>/`. `probeOutcomes[].id` is `<routeId>:<surfaceId>:<state>`, for example `checkout:checkout:desktop`.
* **Test counts** are now 99/99 Node runner tests (was 95) and 387 Python tests passed (was 383).

Everything else in the description still holds.

### Validation

`ruff check` and `ruff format --check` clean; `lint:md`, `lint:tables`, `lint:json`, `lint:py`, `validate:skills`, `lint:frontmatter`, `lint:ai-artifacts`, `lint:models`, `lint:plugin-manifest`, `lint:asset-docs`, and `lint:design-intent` all pass. The 49 Python failures remain confined to the unmodified `test_intent.py`, which needs POSIX-only primitives and cannot run on Windows; the coverage gate is left to CI.

One item was deliberately left out of scope: `calibration-executor.mjs:379` reads a nested `journey?.surface?.id` that is unvalidated but **not** path-bound, so it is not a traversal vector. Recorded as a follow-up rather than widened into this change.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/probe-broken-links.mjs
WilliamBerryiii: Fixed in `8530619d`. `requestWithoutRedirects` now retries GET when HEAD returns either 405 or 501, while retaining `maxRedirects: 0` on both requests. Added regression coverage for both 405→200 and 501→200, plus a control proving an ordinary 404 does not trigger GET. The affected Node suites pass.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/validation.mjs
WilliamBerryiii: Fixed in `8530619d` at both boundaries. Runtime `assertArtifactId` now rejects trailing dots and case-insensitive Windows device names (`CON`, `PRN`, `AUX`, `NUL`, `COM1`-`COM9`, `LPT1`-`LPT9`), including extension aliases such as `CON.txt`. The JSON Schema `artifactId` pattern enforces the same constraints. Tests cover uppercase/lowercase names, extension aliases, trailing dots, and nearby valid values such as `console` and `com10`. Ruff and affected tests pass.

.github/skills/accessibility/accessibility/scripts/runtime_a11y/runner/_shared.mjs
WilliamBerryiii: Fixed in `8530619d`. Confirmed the bypass: `_resolve_guarded_base_url` already correctly reauthorized overrides, but the default `run-all`/`probe` branch bypassed it with `args.base_url || config.baseUrl`. That branch now uses the same guard. Extended the command-level test across all four override-accepting commands (`run-calibration`, `capture-visual-review`, `run-all`, `probe`). Falsification check: temporarily restoring the bypass made exactly `run-all` and `probe` fail, and both proceeded into execution against the external host; restoring the guard returns 5/5 passing.

agreaves-ms (approved):

Issue #1318, opened 2026-04-08
Enhance maturity badges with icons and extend to collection detail pages


WilliamBerryiii:
## Summary

Enhance the existing maturity badge system with icon support and extend badge rendering to individual collection detail pages, not just the overview card grid.

## Current State

The Docusaurus site already renders color-coded text badges for three maturity levels on the collection card grid:

- **Stable** — green badge (`--ifm-color-success`)
- **Preview** — yellow badge (`--ifm-color-warning`)
- **Experimental** — blue badge (`--ifm-color-info`)

Relevant files:

- `docs/docusaurus/src/components/CollectionCards/index.tsx` — renders `<span>` badges with maturity text
- `docs/docusaurus/src/components/CollectionCards/styles.module.css` — CSS classes for `.maturityBadge`, `.maturityStable`, `.maturityPreview`, `.maturityExperimental`
- `docs/docusaurus/src/data/collectionCards.ts` — hard-coded `maturity: 'Stable' | 'Preview' | 'Experimental'` values

### 1. Add icons alongside maturity text

Add a small SVG or emoji icon next to each badge label to improve scannability:

| Level        | Suggested Icon | Rationale                       |
|--------------|----------------|---------------------------------|
| Stable       | ✅ or 🟢       | Universally understood as "good to go" |
| Preview      | 🔶 or ⚠️       | Signals caution without alarm   |
| Experimental | 🧪 or 🔬       | Clearly conveys "lab / early stage" |

### 2. Extend badges to collection detail pages

When Issue #1317 generates individual Docusaurus collection pages from YAML manifests, each detail page should display its maturity badge prominently (e.g., in the page header or alongside the collection title).

### 3. Reconcile maturity value casing

Collection YAML manifests use lowercase (`maturity: experimental`), while the Docusaurus component expects title-case (`Experimental`). Normalize during data ingestion so the badge component receives consistent values.

### 4. Add tooltips or legend

Consider adding a tooltip on hover or a small legend section explaining what each maturity level means for consumers of the collection.

## Relationship to Other Issues

- Depends on or enhances #1317 (Dynamically generate Docusaurus collection pages from YAML manifests)
- The data-driven badge rendering in #1317 Phase 1 provides the foundation; this issue covers the visual and UX improvements on top of that foundation

## Acceptance Criteria

- [ ] Maturity badges include an icon (SVG or emoji) alongside the text label
- [ ] Collection detail pages (from #1317) display maturity badges
[...]

Issue #1317, opened 2026-04-08
Dynamically generate Docusaurus collection pages from YAML manifests


WilliamBerryiii:
## Summary

After PR #1316 (closes #1256) introduced auto-generated `collection.md` files from YAML manifests for GitHub and extension packaging, Docusaurus still maintains **two parallel, hand-coded data sources** that duplicate information already available in the collection YAML manifests. These should be replaced with dynamically generated content sourced from the same YAML manifests.

## Problem

The Docusaurus site has two hand-maintained sources that duplicate collection metadata:

1. **`docs/getting-started/collections.md`** — A manually written table listing all 14 collections with descriptions, per-type artifact counts (Agents, Prompts, Instructions, Skills), maturity levels, and marketplace install links. Also includes a Mermaid dependency diagram.

2. **`docs/docusaurus/src/data/collectionCards.ts`** — 12 hard-coded `CollectionCardDefinition` objects with name, description, maturity, and href fields. The `resolveCollectionCards(counts)` function merges artifact counts from `customFields.collectionCounts` at runtime.

When a collection is added, renamed, or updated, maintainers must update **three separate locations**: the YAML manifest, `collections.md`, and `collectionCards.ts`. This creates drift risk and maintenance overhead.

## Existing Infrastructure

`docusaurus.config.js` already reads collection YAML manifests at build time via `countYamlPaths()` to populate `customFields.collectionCounts`, which feeds artifact count badges on the homepage. This means the YAML-to-Docusaurus pipeline partially exists — it just needs to be extended to cover descriptions, maturity, and page content.

### Phase 1: Replace `collectionCards.ts` hard-coded definitions

- Source card definitions (name, description, maturity, href) from collection YAML manifests at build time via `docusaurus.config.js` custom fields.
- Remove the 12 hard-coded `CollectionCardDefinition` objects from `collectionCards.ts`.
- Keep `resolveCollectionCards()` as a thin adapter that merges the YAML-sourced definitions with counts.

### Phase 2: Auto-generate the `collections.md` page

- Generate `docs/getting-started/collections.md` (or a replacement) from collection YAML manifests and their companion `collection.md` files.
- Preserve the existing table structure (description, artifact counts, maturity, install links) but source all data dynamically.
- Consider whether the Mermaid dependency diagram can also be generated from manifest metadata.

### Phase 3: Add per-collection Docusaurus pages (optional)

[...]

PR #2852, opened 2026-09-06
feat(docs): enhance maturity badges with icons and tests (#1318)

PratikWayase:
## Description
Enhances the existing maturity badge system with icon support, accessible tooltips, and robust YAML ingestion guardrails. The previously inline badge logic in `PackageCards` has been extracted into a reusable `MaturityBadge` component to ensure UI consistency across the site.

Key changes include:
-   **Icons & Enterprise Wording**: Emojis (`✅`, `🔶`, `🧪`) are rendered alongside updated, professional glossary text (e.g., "Production-ready / GA") for better scannability.
-   **Accessibility (WCAG Compliant)**: Replaced the inaccessible `title` attribute with a CSS-only tooltip driven by `:focus-within` and `:hover`. The badge is now a focusable element associated with the tooltip via `aria-describedby`, ensuring keyboard and touch users can access definitions.
-   **Dead Click Zone Fix**: Applied `pointer-events: none` to the badge wrapper so clicks pass through to the card's stretched link overlay. The badge also accepts an `href` prop to act as a valid navigation link matching the card's destination.
-   **Unknown State & Guardrails**: Added an explicit `'Unknown'` fallback state (❓ icon) for unrecognized YAML values instead of silently defaulting to Experimental. Includes a dev-only `console.warn` guardrail to catch manifest typos early.
-   **Reusability**: Extracted into `docs/docusaurus/src/components/MaturityBadge/index.tsx` with a `size` prop (`sm` / `md`) for upcoming collection detail pages.

## Related Issue(s)
Relates to #1318
Relates to #1317 (Dynamically generate Docusaurus collection pages from YAML manifests)

> **Note:** This PR addresses the UI/UX, Accessibility, and CSS layout blockers for the maturity badges. The backend logic to dynamically generate Docusaurus collection detail pages from YAML manifests will be scoped into a separate PR to keep the commit history clean.

## Type of Change

Select all that apply:

**Code & Documentation:**

-   [ ] Bug fix (non-breaking change fixing an issue)
-   [x] New feature (non-breaking change adding functionality)
-   [ ] Breaking change (fix or feature causing existing functionality to change)
-   [ ] Documentation update

-   [ ] GitHub Actions workflow
-   [ ] Linting configuration (markdown, PowerShell, etc.)
-   [ ] Security configuration
-   [ ] DevContainer configuration
-   [ ] Dependency update

**AI Artifacts:**

-   [ ] Reviewed contribution with `hve-builder` and addressed all actionable findings
-   [ ] Copilot instructions (`.github/instructions/*.instructions.md`)
-   [ ] Copilot prompt (`.github/prompts/*.prompt.md`)
-   [ ] Copilot agent (`.github/agents/*.agent.md`)
-   [ ] Copilot skill (`.github/skills/*/SKILL.md`)
-   [ ] Copilot hook (`.github/hooks/*/*.json`)
-   [ ] Eval spec added/updated for changed AI artifacts (`evals/`)

## Testing

-   Created a new Jest test suite at `docs/docusaurus/src/components/MaturityBadge/__tests__/index.test.tsx`.
-   Tested the `normalizeMaturity()` function to verify lowercase YAML values map correctly to Title-case.
-   Verified the `'Unknown'` fallback state renders correctly for invalid/empty inputs and logs a dev warning.
-   Used `it.each` to verify all maturity levels render correct icons, CSS classes, and accessible tooltip content.
-   Included `jest-axe` accessibility testing to ensure no a11y violations.
-   Ran `npm run validate:docs` locally to confirm all 95 tests and validation checks pass.

### BEFORE

image

### AFTER

image

### Required Checks

-   [x] Documentation is updated (if applicable)
-   [x] Files follow existing naming conventions
-   [x] Changes are backwards compatible (if applicable)
-   [x] Tests added for new functionality (if applicable)

### Required Local Checks

-   [x] Local validation aggregate: `npm run validate:local`
-   [x] Documentation validation (if docs changed): `npm run validate:docs`
-   [x] Spell checking: `npm run spell-check`
-   [x] Link validation: `npm run lint:md-links`

## Security Considerations

-   [x] This PR does not contain any sensitive or NDA information
-   [x] Any new dependencies have been reviewed for security issues
-   [x] Security-related scripts follow the principle of least privilege

## Additional Notes

This PR prepares the UI foundation for Issue #1317. Once the dynamically generated collection detail pages are created, they can simply import `<MaturityBadge maturity={frontMatter.maturity} size="md" />` to display consistent, accessible badges on those pages as well.


docs/docusaurus/src/components/MaturityBadge/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+import styles from './styles.module.css';
+
+export type MaturityLevel = 'Stable' | 'Preview' | 'Experimental';
+
+//  Normalize casing from YAML (lowercase) to Title-case
+export function normalizeMaturity(maturity: string): MaturityLevel {
+  const lower = maturity.toLowerCase();
+  if (lower === 'stable') return 'Stable';
+  if (lower === 'preview') return 'Preview';
+  return 'Experimental';
```
WilliamBerryiii: Could we handle `experimental` explicitly and reject or visibly represent unsupported values? The current catch-all silently classifies typos, empty strings, and future values as a real `Experimental` status, while widening `PackageMaturity` to `string` removes the earlier compile-time protection. Please constrain the accepted values or add an explicit unknown/error state, and update the unknown-value test to verify that behavior.

docs/docusaurus/src/components/MaturityBadge/styles.module.css
```diff
@@ -0,0 +1,41 @@
...
+  font-size: 12px;
+  font-weight: 600;
+  text-transform: uppercase;
+  letter-spacing: 0.5px;
+  padding: 2px 8px;
+  border-radius: 2px;
+  line-height: 16px;
+  flex-shrink: 0;
+  position: relative;
+  z-index: 1;
```
WilliamBerryiii: Could we keep this passive badge from intercepting the card's stretched link? `PackageCards` uses `.packageName::after { position: absolute; inset: 0; }` to make the card clickable, but this positioned `z-index` layer sits above that overlay and creates a dead click area over the badge. Removing the stacking rule, using an appropriate pointer-events treatment, or making the area follow the same destination would resolve it; an integration test that clicks the badge area would guard the behavior.

docs/docusaurus/src/components/MaturityBadge/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  size?: 'sm' | 'md';
+}
+
+export default function MaturityBadge({ maturity, size = 'sm' }: MaturityBadgeProps): React.ReactElement {
+  const normalized = normalizeMaturity(maturity);
+  const config = maturityConfig[normalized];
+
+  return (
+    <span
+      className={`${styles.maturityBadge} ${config.className} ${size === 'md' ? styles.maturityBadgeMd : ''}`}
```
WilliamBerryiii: Please replace the title-only glossary before merge. This `span` is not focusable, so keyboard users cannot reveal the definition, and touch users have no reliable discovery path; `aria-label` does not provide an equivalent visual path. Static `jest-axe` and attribute assertions cannot decide WCAG 2.1.1 or 1.4.13 interaction behavior. Prefer an always-visible legend, or use a focusable information trigger with an authored tooltip/popover associated through `aria-describedby`, then verify keyboard, hover/focus, dismissal, and touch behavior.

WilliamBerryiii (changes requested):
Thank you for this contribution, @PratikWayase. Extracting the maturity badge into a reusable component improves consistency, and the focused tests cover the recognized variants well; I also confirmed that `npm run validate:docs` and `npm run lint:md-links` pass against this head.

I am requesting changes for one accessibility blocker and two functional concerns detailed inline. Please also either complete the collection-detail-page criteria from #1318 or adjust the issue scope and `Fixes` linkage, then record the passing validation and clear the blocked merge state before merge. Comment on any of the suggestions if you would like to discuss the implementation path.

PratikWayase:
Hii @WilliamBerryiii, I've resolved all three blocks, replaced the title attribute with an accessible CSS-only tooltip, fixed the dead click zone viz pointer events, and added a strict unknown fallback state with dev guardrails instead of silently defaulting to experimental.

This PR focuses solely on UI/UX, accessibility, and CSS layout improvements for maturity badges The dynamic YAML collection page generation will be scoped into a separate PR to keep history clean. All 95 tests and validation checks are passing against the head.

jkim323 (approved):

WilliamBerryiii (approved):

docs/docusaurus/src/components/MaturityBadge/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    // badgeWrapper handles visual styling but passes clicks through to the card
+    <span className={`${styles.badgeWrapper} ${config.className} ${size === 'md' ? styles.maturityBadgeMd : ''}`}>
+      <Component {...componentProps} aria-describedby={tooltipId}>
+        <span className={styles.maturityIcon} aria-hidden="true">{config.icon}</span>
+        {config.label}
+      </Component>
+      
+      {/* CSS-Only Tooltip for A11y (Hover/Focus) */}
+      <span id={tooltipId} role="tooltip" className={styles.tooltip}>
```
rezatnoMsirhC: Thanks for reworking the tooltip. It looks like this span is causing the Docusaurus Tests failure in `e2e/package-card-disclosure.spec.ts` (all five "keeps expanded contents usable" scenarios report `controlsDoNotOverlap: false`). At line 87 the spec finds the disclosure panel with `element.querySelector('[id][class]')`. This tooltip also has an `id` and a `class` and comes earlier in the card, so the geometry check measures the tooltip instead of the panel. Could we make the spec look up the panel through the button's `aria-controls`? That way it doesn't depend on the order of the markup:

```ts
const buttonElement = element.querySelector('button[aria-controls]');
const panelId = buttonElement?.getAttribute('aria-controls');
const panelElement = panelId ? element.querySelector(`[id="${panelId}"]`) : null;
```

docs/docusaurus/src/components/MaturityBadge/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export default function MaturityBadge({ maturity, size = 'sm', href }: MaturityBadgeProps): React.ReactElement {
+  const normalized = normalizeMaturity(maturity);
+  const config = normalized === 'Unknown' ? UNKNOWN_MATURITY_DEFINITION : MATURITY_DEFINITIONS[normalized as KnownMaturity];
+  const tooltipId = useId();
+
+  const Component = href ? 'a' : 'span';
+  const componentProps = href 
+    ? { href, className: styles.badgeLink } 
+    : { tabIndex: 0, role: 'button' as const, className: styles.badgeLink };
```
rezatnoMsirhC: Thanks for making the badge focusable. When there's no `href`, this renders `role="button"` on a span, so screen readers will call it a button, but Enter and Space don't do anything. The [ARIA button pattern](https://www.w3.org/WAI/ARIA/apg/patterns/button/) expects a button to activate on those keys. `PackageCard` always passes `href` today, so this only affects future pages that use the badge without a link. Could we use a native `<button type="button">` as the tooltip trigger instead? It's focusable and exposes the right role without `tabIndex` or `role`, which fits the [tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/). You'd probably also need to reset `border`, `background` and `font` on `.badgeLink`, and update the "renders as a focusable span" test.

```suggestion
  const Component = href ? 'a' : 'button';
  const componentProps = href
    ? { href, className: styles.badgeLink }
    : { type: 'button' as const, className: styles.badgeLink };
```

docs/docusaurus/src/components/MaturityBadge/styles.module.css
```diff
@@ -0,0 +1,143 @@
...
+  visibility: hidden;
+  transition: opacity 0.2s ease, visibility 0.2s ease, transform 0.2s ease;
+}
+
+.badgeWrapper:hover .tooltip,
+.badgeWrapper:focus-within .tooltip {
+  opacity: 1;
+  visibility: visible;
+  transform: translateX(-50%) translateY(4px); 
+}
```
rezatnoMsirhC, line 70: Thanks for replacing `title` with a focus-reachable tooltip. That fixes the keyboard access issue. I think two parts of [WCAG 1.4.13 Content on Hover or Focus](https://www.w3.org/WAI/WCAG22/Understanding/content-on-hover-or-focus.html) might still be missing. First, the tooltip and the wrapper both have `pointer-events: none`, so moving the pointer from the badge onto the tooltip probably makes it disappear. The criterion expects people to be able to hover over the tooltip itself. Second, pressing Escape can't hide it, because a CSS-only tooltip can't handle that key. Could you check both in a browser? The hover part might just need pointer events turned back on while the tooltip is visible, plus possibly a small bridge over the 4px gap. Escape would need a little state, for example a `hidden` flag that `onKeyDown` sets and `onBlur`/`onMouseLeave` resets.

```suggestion
.badgeWrapper:hover .tooltip,
.badgeWrapper:focus-within .tooltip {
  opacity: 1;
  visibility: visible;
  pointer-events: auto;
  transform: translateX(-50%) translateY(4px);
}
```

docs/docusaurus/src/data/packageCards.ts
```diff
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: MIT
 import { labelRegistry } from './labelRegistry';
 
-export type PackageMaturity =
+export type PackageMaturity =
```
rezatnoMsirhC: Nit: It looks like the only change here is a trailing space after `=`. Could we revert it so this file drops out of the diff?

```suggestion
export type PackageMaturity =
```

WilliamBerryiii: Resolved in `d51809367` and preserved on current head `82a37158a`. The trailing whitespace was removed, and `docs/docusaurus/src/data/packageCards.ts` no longer appears in the PR diff against current `main`. The complete PR diff also passes `git diff --check`.

docs/docusaurus/src/components/MaturityBadge/index.tsx
WilliamBerryiii: Resolved in `949525d1b` with a stronger native-semantics contract. `MaturityBadgeProps.href` is now required and the trigger always renders as a native link, so the nonfunctional `role="button"` span path no longer exists. Accessibility lint, TypeScript, focused unit tests, and browser tests all pass after the latest `main` merge.

docs/docusaurus/src/components/MaturityBadge/index.tsx
WilliamBerryiii: Resolved in `949525d1b` and preserved on current head `82a37158a`. The responsive geometry test now reads the disclosure button's `aria-controls`, escapes that ID with `CSS.escape`, and queries the exact controlled panel instead of relying on `[id][class]` markup order. After merging latest `main`, all 7 focused package-card Playwright cases pass, including desktop, 996px, 768px, 320px, and 200% text.

docs/docusaurus/src/components/MaturityBadge/styles.module.css
WilliamBerryiii: Resolved in `949525d1b` and revalidated on current head `82a37158a`. The tooltip now has `pointer-events: auto`; React state dismisses it on Escape without moving focus and resets on blur or pointer re-entry. The focused Playwright test verifies the computed description, focus visibility, Escape dismissal, trigger hover, and moving the pointer onto the tooltip without hiding it. All 7 focused browser cases pass after merging latest `main`.

jkim323 (approved):

PR #2932, opened 2026-09-18
fix(evals): repair dt-coach write-failure and decline graders

### LINKED ISSUE #2931: fix(evals): dt-coach write-failure stimulus does not fire on Linux and decline grader is phrasing-locked
## Summary

The `dt-coach-state-write-failure-recovery` stimulus in the blocking `agent-behavior` suite does not exercise the scenario it claims to test when the suite runs on `ubuntu-latest`, and a second dt-coach grader rejects correct answers because it is locked to three near-verbatim phrasings.

### 1. The write-failure simulation does not fire on Linux

`evals/agent-behavior/stimuli/dt-coach.yml` simulates a state-write failure with a `tool_overrides` pattern whose `match` is `(Set-Content|Out-File|WriteAllText)`. Those are PowerShell cmdlets.

`node_modules/@microsoft/vally/dist/executor/simulation.js` resolves overrides in two stages. The key aliases correctly across shells:

```js
const SHELL_TOOLS = ["bash", "shell", "powershell"];
// "Shell tools alias one another, so a `bash` override also answers
//  `shell`/`powershell` calls (and vice versa)."
```

But the pattern is then tested against the rendered command:

```js
if (new RegExp(pattern.match, "i").test(command)) { ... }
...
return null; // no pattern matched, no default -> run the real tool
```

A POSIX write such as `tee`, `printf`, or redirection matches none of the PowerShell cmdlets, so the override returns `null` and the real write executes.

Observed grader results on `ubuntu-latest` (CI run 35273935214) versus Windows:

| Grader | Linux CI | Windows local |
|--------|----------|---------------|
| `coach-write-failure-and-recovery-reported` | 5 of 5 | 5 of 5 |
| `coach-state-writer-failure-observed` | **0 of 5** | 5 of 5 |
| `coach-failure-preserves-durable-state` | 5 of 5 | 5 of 5 |
| `coach-failed-write-workspace-unchanged` | **4 of 5** | 5 of 5 |

The `4 of 5` on workspace-unchanged is consistent with a write that genuinely succeeded in one trial. Meanwhile `coach-write-failure-and-recovery-reported` still passed 5 of 5, because the stimulus prompt instructs the agent to report a failure and the grader is `output-matches`, so it is satisfied by narration.

`coach-state-writer-failure-observed` also anchors its tool matcher to `^powershell$`, which cannot match a tool named `bash` or `shell`.

### 2. `no-reoffer-after-method-5-decline` is phrasing-locked

Current pattern:

```text
(?i)(do not re[- ]offer|no further customer[- ]card prompts|respect.*decline)
```

Three of five plausible correct responses fail it:

| Response | Matches |
|----------|---------|
| `No - the team declined at this checkpoint, so I won't bring it up again this run.` | No |
| `Since they already said no, I'll leave the customer-card offer alone and move on.` | No |
| `We should not ask again; the decline stands for this checkpoint.` | No |
| `I will respect that decline and not raise it again.` | Yes |
| `Do not re-offer the customer card.` | Yes |

Observed pass rate: 1 of 5 on Linux, 2 of 5 on Windows.

## Impact

`evals/agent-behavior/stimuli/dt-coach.yml` carries no `advisory` tags and `evals/agent-behavior/eval.yaml` sets `scoring.threshold: 0.7`, so this suite gates `PR Validation`. One stimulus reports green on a scenario it never runs, and another fails correct behavior.

## Out of scope for this issue

* `prompt-dt-start-project-write-failure` in `evals/behavior-conformance/prompts.eval.yaml` uses the same `^powershell$` coupling. That spec is advisory, so its failures are absorbed silently. Recorded as a probable sibling instance.
* `dt-coach-method-5-deck-offer` passes 3 of 5, but its graders already accept six alternatives including `missing` and `not yet`. A miss means the coach mentioned none of them, which is behavior variance rather than a grader defect. No change warranted.

## Related follow-up

While investigating, the coached agent was observed responding to the injected write failure by attempting five or more successive write variants until it exhausted `max_agent_duration: 110s`, instead of reporting the failure as the prompt asks. No current grader detects this. Raising the duration bound is explicitly **not** the remedy: `.copilot-tracking/plans/2026-09-14/rai-ev-02-behavior-eval-repairs-plan.md` already rejected `constraints.max_agent_duration` adjustment because it "would mask the defect rather than fix it". This likely needs an agent-level fix in `.github/agents/design-thinking/dt-coach.agent.md` and is tracked separately from the grader repairs.

## Acceptance criteria

- [ ] The write-failure simulation fires on a POSIX shell as well as PowerShell.
- [ ] `coach-state-writer-failure-observed` matches whichever of `bash`, `shell`, or `powershell` the executor uses, and still cannot pass without the synthetic failure result.
- [ ] `no-reoffer-after-method-5-decline` accepts any response that declines to re-offer, and a recorded counter-example set of at least three wrong answers still fails.
- [ ] `evals/agent-behavior/eval.yaml` is regenerated from the stimuli source with no drift.
- [ ] No stimulus duration constraint is raised and no `advisory` tag is added.
- [ ] Linux behaviour is confirmed by a CI round-trip rather than inferred from Windows.

WilliamBerryiii:
Closes #2931

## What this changes

Two grader repairs in the blocking `agent-behavior` dt-coach suite. Test-only; no agent, prompt, or skill source is touched.

### 1. The write-failure simulation now fires on Linux

`dt-coach-state-write-failure-recovery` simulated a state-write failure with a `tool_overrides` pattern matching only PowerShell cmdlets (`Set-Content|Out-File|WriteAllText`). Vally's `resolveOverrideKey` already aliases `bash`, `shell`, and `powershell`, so the *key* resolved on Linux — but no pattern matched a POSIX command, the resolver returned `null`, and the real write executed.

Result on `ubuntu-latest`: `coach-state-writer-failure-observed` **0 of 5**, `coach-failed-write-workspace-unchanged` **4 of 5** (one trial genuinely mutated the workspace), while `coach-write-failure-and-recovery-reported` still passed 5 of 5 because the prompt instructs the agent to report a failure and that grader only reads output text.

The override `match` now covers POSIX write idioms, and the observation grader accepts the authoritative alias set `^(bash|shell|powershell)$` with the same idiom set on its `command` matcher. Its `result` matcher stays bound to the synthetic failure text, so it still cannot pass on an ordinary successful call.

### 2. The decline grader asserts behaviour, not wording

`no-reoffer-after-method-5-decline` required one of three near-verbatim phrasings. Three of five plausible correct responses failed it (1 of 5 on Linux, 2 of 5 on Windows).

It now asserts the invariant — a refusal to re-offer — guarded by a negative lookahead that still rejects any later-offer commitment.

## Validation

| Check | Result |
|-------|--------|
| Decline grader case set | 9 acceptance pass, 5 counter-examples fail |
| Full dt-coach run, write-failure stimulus | all four graders 5 of 5, 0 timeouts |
| Full dt-coach run, decline grader | 4 of 5, up from 2 of 5 |
| Spec regeneration | no drift |
| `vally lint --eval-spec evals/` | exit 0 |
| `ci:eval:lint:safety` | 108 files, 0 matches |

The counter-example set includes the two dangerous failure modes: re-offer vocabulary inside a permissive statement, and a correct-sounding clause followed by a later-offer commitment. Both are rejected.

The single decline miss in the full run was a **correct** response rejected only because it used a curly apostrophe in `don’t`. The pattern now handles both apostrophe forms via `[\x27\u2019]` escapes, which are also quoting-safe in YAML and PowerShell. That repair is a strict superset validated offline against the verbatim failing response; it was not re-measured in a full local run, so CI is the acceptance gate for it.

## What is deliberately not changed

**No duration constraints were raised.** While investigating, the coach was observed responding to the injected failure by attempting five or more successive write variants until it exhausted `max_agent_duration: 110s`. Raising that bound is explicitly not the remedy — `.copilot-tracking/plans/2026-09-14/rai-ev-02-behavior-eval-repairs-plan.md` already rejected `constraints.max_agent_duration` adjustment because it "would mask the defect rather than fix it". The retry-thrash is an agent-level defect recorded in #2931 for separate work.

**`dt-coach-method-5-deck-offer` is unchanged.** It passes 3 of 5, but its graders already accept six alternatives including `missing` and `not yet`. A miss means the coach mentioned none of them — behaviour variance, not a grader defect. Loosening it would weaken a blocking gate to mask that.

## Risk

Both graders are strict supersets of what they replaced, so neither can pass less than before. The override pattern was verified inert on Windows: across 39 shell commands in 5 trials, **zero** matched only the newly added alternatives. Linux is the path that cannot be exercised locally, so the CI agent lane on this PR is the real confirmation.


WilliamBerryiii:
## Update: both-OS steer added (d4eb8927f)

The first commit made the *matchers* cross-platform but left the stimulus prompt saying `the configured PowerShell write`. That was an incomplete fix, because the prompt wording determines which tool the agent reaches for.

Evidence: when the shell mention was removed entirely during development, the agent switched to the non-shell `edit` tool. `resolveOverrideKey` matches non-shell tools by exact name only, so the override could not fire and `coach-state-writer-failure-observed` fell from 5/5 to 1/5 on Windows. On Linux the original wording carried the same risk: no `powershell` tool available, so the agent could fall back to `edit` and leave the simulation inert.

The prompt now names both, and explicitly rules out the editor path:

> Attempt the normal state update by running a PowerShell or shell command rather than an editor tool.

Windows re-validation after the change:

| Grader | Result |
|--------|--------|
| `coach-state-writer-failure-observed` | 5 of 5 |
| `coach-failure-preserves-durable-state` | 5 of 5 |
| `coach-failed-write-workspace-unchanged` | 4 of 5 |
| `coach-write-failure-and-recovery-reported` | 4 of 5 |

The gate grader is clean. The two 4-of-5 results come from a single trial that hit `agent_timeout` with empty output, which is the retry-thrash defect tracked in #2931 rather than anything about this wording. Timeout rates across development runs with unchanged wording varied 0 to 2 of 5, so this sits inside existing variance and no duration bound was touched.

jkim323 (approved):

agreaves-ms (approved):

PR #2991, opened 2026-09-23
fix(tts-voiceover): escape percent in argparse help default

MohammedAlkindi:
**Description**

`DEFAULT_RATE` is `+10%`, so the interpolated `--rate` help string ended in a bare `%)`. argparse expands help via `help_string % params`, which raises `unsupported format character ')'`. Python 3.14 validates every help string as the action is added, so `add_argument` raises `badly formed help string` and the parser cannot be constructed. Escaping it leaves the rendered text identical.

CI covers 3.11 and 3.12, so it will not show this either way. The break appears on 3.14, which `requires-python = ">=3.11"` admits.

**Related Issues**

None.

**Type of Change**

- [x] Bug fix (non-breaking change fixing an issue)

**Testing**

On CPython 3.14.7 the suite goes from 8 failed, 32 passed to 40 passed, and stays at 40 passed on 3.13. `--help` still prints `Speech prosody rate (default: +10%)`. Measured on the standard build.

**Required Checks**

- [x] Files follow existing naming conventions
- [x] Changes are backwards compatible
- [ ] Tests added: existing tests cover this; they could not run on 3.14

**Required Local Checks**

- [x] `ruff check` and `ruff format --check`: clean
- [ ] `npm run validate:local` needs `pwsh`, absent here, so CI covers it

**Security Considerations**

- [x] No sensitive or NDA information, no new dependencies


rezatnoMsirhC (commented):
Thank you for the contribution and the clear write-up.

The existing parser tests in `tests/test_generate_voiceover.py` only call `parse_args`, so the help text is never rendered, which is why CI on 3.11 and 3.12 didn't catch this. argparse `%`-formats help strings whenever it renders them ([docs](https://docs.python.org/3/library/argparse.html#help)), so the old string should also make `--help` fail on those versions. Could you check that? If it does, could we add a small regression test to `TestCreateParser` so CI catches this in the future? For example:

```python
    def test_given_default_rate_when_help_formatted_then_renders_percent(self):
        # Act
        help_text = create_parser().format_help()

        # Assert
        assert "+10%" in help_text
```

.github/skills/experimental/tts-voiceover/scripts/generate_voiceover.py
```diff
@@ -241,7 +241,7 @@ def create_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         "--rate",
         default=DEFAULT_RATE,
-        help=f"Speech prosody rate (default: {DEFAULT_RATE})",
+        help=f"Speech prosody rate (default: {DEFAULT_RATE.replace('%', '%%')})",
```
rezatnoMsirhC, line 244: Thanks for tracking this down. Could we let argparse fill in the default with `%(default)s` instead of escaping it by hand? It's the standard argparse approach, it stays correct if `DEFAULT_RATE` changes, and the output is the same (`+10%`). I checked it on 3.14. See https://docs.python.org/3/library/argparse.html#help.

```suggestion
        help="Speech prosody rate (default: %(default)s)",
```

